PR #24, opened 2021-02-23
Project status

positonic:
Hide deactivated projects, all users to activate and deactivate. Check if users have permissions to change status.


mateodaza (approved):
looking good

PR #25, opened 2021-02-23
adds transaction ID to saveDonation mutation

mateodaza:
As we are now always getting the tx hash ID we can always save it instantly


mateodaza:
This also needs the front-end to be updated

PR #50, opened 2021-04-22
Updated Segment Track calls

divine-comedian:
Updated Made Donaton & Project Created segment events with flattened properties.


resolvers/projectResolver.ts
```diff
@@ -468,6 +468,18 @@ export class ProjectResolver {
       project,
       null
     )
+    const segmentProject = project.map(project => {
```
positonic: Can you indent this properly

positonic (changes requested):
Please fix the indentation formatting

PR #54, opened 2021-04-26
remove password from user response

sschauss:
### What are the changes and their implications?

Before this, the hashed passwords for every user could be requested.

- hide password column from find or query
- explicitly delete password from login/register resolver response


positonic (approved):
Nice one Simon!

PR #56, opened 2021-04-27
'update profile' add properties tracked

divine-comedian:
added segment properties to track 'updated profile' analytics

### DIFF
--- a/resolvers/donationResolver.ts
+++ b/resolvers/donationResolver.ts
@@ -144,8 +144,14 @@ export class DonationResolver {
       const baseTokens =
         Number(priceChainId) === 1 ? ['USDT', 'ETH'] : ['WXDAI', 'WETH']
 
+      let donorEmail = ""
+      let donorFirstName = ""
+      if (user != null) {
+        donorEmail = user.email
+        donorFirstName = user.firstName
+      }
       const segmentDonation = {
-        email: user.email,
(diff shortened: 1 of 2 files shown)


resolvers/donationResolver.ts
```diff
@@ -144,8 +144,14 @@ export class DonationResolver {
       const baseTokens =
         Number(priceChainId) === 1 ? ['USDT', 'ETH'] : ['WXDAI', 'WETH']
 
+      let donorEmail = ""
```
positonic: Would be more elegant like:
`let donorEmail = (user != null) ? user.email : ''`

resolvers/donationResolver.ts
```diff
@@ -144,8 +144,14 @@ export class DonationResolver {
...
+      let donorFirstName = ""
+      if (user != null) {
+        donorEmail = user.email
+        donorFirstName = user.firstName
+      }
       const segmentDonation = {
-        email: user.email,
+        email: donorEmail,
```
positonic: or even better:

```
const segmentDonation = {
   email: (user != null) ? user.email : ''
   ....
}
```

positonic (changes requested):
Check my comments and let me know what you think?

positonic (commented):
Great job!

PR #75, opened 2021-05-19
Copy Email changes live

positonic:


mateodaza:
Just reviewd the code, don't believe it will break anything before moving the nextjs version - hoping for the best haha!

PR #88, opened 2021-06-15
add back segmentProject

divine-comedian:
added back in segmentProject to the analytics.track for project create
segment output locally looks like this - 

image

we need to capture this information in order to send emails correctly in autopilot


mateodaza:
is this the same as #95? If so please close it

divine-comedian:
@mateodaza  these are different, this one deals with  the project created mutation while #95 deals with adding in a new tracker for project deactivated.

PR #95, opened 2021-06-17
add deactivated project tracker

### LINKED ISSUE #91: add tracker to deactivateProjects call
We would like to notify users that their project has been deactivated, by an admin or someone on their account.

we need to create a analytics.track call for when the mutation happens in the impact-graph the information we need is sent to segment that can be passed along into an auotmated email informing the user. 

in project.resolvers.ts this is the specific mutation we're looking at 
```
@Mutation(returns => Boolean)
  async deactivateProject (
    @Arg('projectId') projectId: number,
    @Ctx() ctx: MyContext
  ): Promise<Boolean> {
    try {
      const user = await getLoggedInUser(ctx)
      return await this.updateProjectStatus(projectId, ProjStatus.can, user)
    } catch (error) {
      Logger.captureException(error)
      throw error
      return false
    }
  }
```

and this is an example of the track call that currently is used for when projects are created on giveth.io 
```
const segmentProject = {
      email: project.users[0].email,
      projectOwnerEmail: project.users[0].email,
      title: project.title,
      projectOwnerLastName: project.users[0].lastName,
      projectOwnerFirstName: project.users[0].firstName,
      projectOwnerId: project.admin,
      slug: project.slug,
      projectWalletAddress: project.walletAddress
    }

    analytics.track(
      'Project created',
      `givethId-${ctx.req.user.userId}`,
      segmentProject,
      null
    )
```

divine-comedian:
added in tracker to log when projects are deactivated and capture it in segment

### DIFF
--- a/resolvers/projectResolver.ts
+++ b/resolvers/projectResolver.ts
@@ -387,7 +387,7 @@ export class ProjectResolver {
       heartCount
     )
     const slugBase = slugify(newProjectData.title)
-    
+
     let slug = slugBase
     for (let i = 1; await this.projectRepository.findOne({ slug }); i++) {
       slug = slugBase + '-' + i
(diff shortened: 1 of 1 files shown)


resolvers/projectResolver.ts
```diff
@@ -921,7 +921,30 @@ export class ProjectResolver {
...
+           }
+
+      analytics.track(
+        'Project deactivated',
+        `givethId-${ctx.req.user.userId}`,
+        segmentProject,
+        null
+      )
+    }
+  }
```
cquinterom096, line 946: Looking good, add some spacing to align analytics.track method and the brackets underneath so that it's more readable. 
Align them with the bracket before the catch (error) method.

cquinterom096 (changes requested):

resolvers/projectResolver.ts
```diff
@@ -921,7 +921,30 @@ export class ProjectResolver {
...
+           const segmentProject = {
+             email: user.email,
+             title: project.title,
+             LastName: user.lastName,
+             FirstName: user.firstName,
+             OwnerId: project.admin,
+             slug: project.slug,
+           }
+
+      analytics.track(
```
cquinterom096: Looking good, add some spacing to align analytics.track method and the brackets underneath so that it's more readable.
Align them with the bracket before the catch (error) method.

cquinterom096 (changes requested):

resolvers/projectResolver.ts
divine-comedian: added a commit! does it looks good now?

cquinterom096: LGTM!

PR #106, opened 2021-06-29
fix and simplify project update analytics tracker

### LINKED ISSUE #86: Add back in missing values from 'Project Created' Track call for Segment
## 📋 Anything to add?

current track called calls the properties from an object called formattedProject in projectResolver.ts and returns to segment -
```
analytics.track({
  userId: 'givethId-67',
  event: 'Project created',
  properties: {
    admin: 'atvatvaev',
    categories: [
      'non-profit'
    ],
    description: '<p>atrythrybrah</p>',
    imageStatic: '2',
    impactLocation: '',
    organisationId: 1,
    title: 'arfvadv',
    walletAddress: '0x38b3f21c23424F955EC5D259b04eE198862d6847'
  }
});
```

this is problematic because we're missing some important fields in order for autopilot and segment to work properly
we need to be capturing at least:

- project owner's email
- project owner's name
- project slug
- project title
- project owner user ID
- project wallet address.

### LINKED ISSUE #94: fix track calls for project update - donor
currently there are two analytic.track calls being sent to segment for when a project makes an update on giveth.io  from the giveth impact-graph. The intention is to eventually send emails to donors letting them know a project they donated to has an update. 

one is for the project owner
and another for donors to the project. the problem is that a track call is being sent for every instance of donation made not for every unique address that made a donation. 

this is problematic in that if a user has made several donations to a single project and the project has an update they would receive one email for each donation they had made to the project. meaning for one project update a user could receive several redundant emails

from projectresolver.ts 

```
  const donations = await this.donationRepository.find({
      where: { project: { id: project.id } },
      relations: ['user']
    })

    donations.forEach(donation => {
      analytics.track(
        'Project updated - donor',
        `givethId-${donation.user.id}`,
        {
          project,
          update: title,
          donation
        },
        null
      )
```

divine-comedian:
updated the analytics tracker for project update, related to issue #94 

also trimmed down the amount of info included in the analytics.

### DIFF
--- a/resolvers/projectResolver.ts
+++ b/resolvers/projectResolver.ts
@@ -573,7 +573,8 @@ export class ProjectResolver {
       walletAddress: project.walletAddress
     }
  // -Mitch I'm not sure why formattedProject was added in here, the object is missing a few important pieces of
- // information into the analytics... 
+ // information into the analytics
+ 
     const formattedProject = {
       ...projectInput,
       description: projectInput?.description?.replace(/<img .*?>/g, '')
@@ -646,7 +647,7 @@ export class ProjectResolver {
       return donation.user
     })
     const uniqueDonors = projectDonors.filter((currentDonor, index) => {
-        return projectDonors.findIndex(duplicateDonor => duplicateDonor.id === currentDonor.id) === index
+        return (currentDonor != null) && (projectDonors.findIndex(duplicateDonor => duplicateDonor.id === currentDonor.id) === index)
       })
 
     uniqueDonors.forEach(donor => {
--- a/resolvers/projectResolver.ts
+++ b/resolvers/projectResolver.ts
@@ -573,7 +573,7 @@ export class ProjectResolver {
       walletAddress: project.walletAddress
     }
  // -Mitch I'm not sure why formattedProject was added in here, the object is missing a few important pieces of
- // information into the analytics... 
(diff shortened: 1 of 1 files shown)


resolvers/projectResolver.ts
```diff
@@ -615,14 +615,18 @@ export class ProjectResolver {
       createdAt: new Date(),
       isMain: false
     })
-
+    const projectUpdateInfo = {
+      title: project.title,
+      email: project.users[0].email,
```
cquinterom096: I'll do a small change here @divine-comedian, project.users[0].email will break for old projects.

cquinterom096 (changes requested):

resolvers/projectResolver.ts
divine-comedian: okay, what part needs to be changed?

cquinterom096: The user object is already the owner. So typing user.email should do it.

divine-comedian: so when I log user.email on  projectUpdate it returns undefined
however logging project.users[0].email returns the correct info

cquinterom096 (approved):
looks good!

MoeNick:
@divine-comedian According to our meeting, pls request all backed code reviews from @aminlatifi 
thanks

PR #109, opened 2021-06-30
Analytics add USD and ETH values

divine-comedian:
Added in to analytics the USD and ETH amounts for each donation on 'Donation Received' track


resolvers/donationResolver.ts
```diff
@@ -202,13 +174,46 @@ export class DonationResolver {
 
           donation.valueUsd = Number(amount) * donation.priceUsd
           donation.valueEth = Number(amount) * donation.priceEth
-
+          return [
+             donation.valueUsd,
+             donation.valueEth
+           ]
```
cquinterom096, line 183: this return should be after the await .save() of the donation.

cquinterom096 (changes requested):

PR #125, opened 2021-08-12
pre calculated total donations and hearts for projects

mateodaza:


mateodaza:
Thanks @cquinterom096 for this !!

PR #148, opened 2021-10-01
add listed segment track, verified object on donation track

### LINKED ISSUE #147: create segment call for when a project is listed
Add in a segment call when a project's listed status === `true`

we will link this into an email on autopilot notifying the project.

divine-comedian:


resolvers/donationResolver.ts
```diff
@@ -219,7 +219,8 @@ export class DonationResolver {
             toWalletAddress: toAddress.toString().toLowerCase(),
             fromWalletAddress: fromAddress.toString().toLowerCase(),
             donationValueUsd: donation.valueUsd,
-            donationValueEth: donation.valueEth
+            donationValueEth: donation.valueEth,
+            verified: project.verified != null ? project.verified : ''
```
mohammadranjbarz: @divine-comedian 
I think it's better to set a `boolean` value for `verified` field and you can write this line shorter like this:
`verified: project.verified` (project.verified is boolean)

mohammadranjbarz (changes requested):
All the thing are fine to me, except on thing that put comment for that

resolvers/donationResolver.ts
divine-comedian: my only intention is that if we hit some weird edge cases where the project doesn't have a verified property that we can assign something to it, preventing it from breaking.

Perhaps instead of an empty string `' '` we change it fo `false`? Or if we're absolutely confident we can simply keep it as `verified: project.verified` as you suggested.

mohammadranjbarz: As I know we have it always but because I'm not 100% sure, so you can set something like this (just for having simpler code)
`verified: Boolean(project.verified)`

PS: Sorry for the requested changes because your functionality was ok from the beginning, I think it is worth putting some time to write cleaner code ( but anytime you thought I'm obsessed with small things just tell me :)  )

divine-comedian:
new changes are in @mohammadranjbarz

mohammadranjbarz (approved):
Well done man.

divine-comedian:
resolve issue https://github.com/Giveth/impact-graph/issues/147

PR #154, opened 2021-10-09
Verify donation transaction

mohammadranjbarz:
related to related https://github.com/Giveth/impact-graph/issues/168, https://github.com/Giveth/GIVeconomy/issues/81


mohammadranjbarz:
@aminlatifi 

We need these configs  in .env  ( for local and other environments)

```
BLOCKSCOUT_API_URL=https://blockscout.com/xdai/mainnet/api
ETHERSCAN_MAINNET_API_URL=https://api.etherscan.io/api
ETHERSCAN_ROPSTEN_API_URL=https://api-ropsten.etherscan.io
ETHERSCAN_API_KEY=
XDAI_NODE_HTTP_URL=https://dry-small-sound.xdai.quiknode.pro
MAINNET_NODE_HTTP_URL=https://mainnet.infura.io/v3/infuraApiKey
ROPSTEN_NODE_HTTP_URL=https://ropsten.infura.io/v3/infuraApiKey
NUMBER_OF_VERIFY_DONATION_CONCURRENT_JOB=1
VERIFY_DONATION_CRONJOB_EXPRESSION=0 0 * * * *
```

and if you want to run test you need some keys, please ping me if you want to run tests locally to give you some keys

package.json
```diff
@@ -76,8 +79,12 @@
     ]
   },
   "devDependencies": {
+    "@types/axios": "^0.14.0",
+    "@types/bull": "^3.15.5",
     "@types/chai": "^4.2.15",
     "@types/mocha": "^8.2.1",
+    "@types/node-cron": "^3.0.0",
+    "@types/web3": "^1.2.2",
```
aminlatifi: New web3 has types inside, so we can remove it
image

services/transactionService.ts
```diff
@@ -0,0 +1,507 @@
+import Web3 from 'web3'
+import abiDecoder from 'abi-decoder'
+import { findTokenByNetworkAndSymbol, NETWORK_IDS } from '../utils/tokenUtils';
+import { errorMessages } from '../utils/errorMessages';
+import { NetworkTransactionInfo, TransactionDetailInput } from '../types/TransactionInquiry';
+import axios from 'axios';
+import config from '../config';
+
+const mainnetNodeUrl = config.get('MAINNET_NODE_HTTP_URL') as string
```
aminlatifi, line 13: These can be removed and provider/providers.ts utility function be used

services/transactionService.ts
```diff
@@ -0,0 +1,507 @@
...
+    default:
+      throw new Error(errorMessages.INVALID_NETWORK_ID)
+  }
+
+}
+
+/**
+ * @see @link{https://gist.github.com/veox/8800debbf56e24718f9f483e1e40c35c, https://stackoverflow.com/a/67617052/4650625}
+ */
+const erc20ABI =
```
aminlatifi: Please put it somewhere else, a separate file in asset folder or something like taht.

services/transactionService.ts
```diff
@@ -0,0 +1,554 @@
...
+    return { ...foundTransaction, speedup: true }
+  }
+  console.log('checkIfTransactionHasBeenSpeedup', {
+    recentDonationsLength: userRecentTransactions.length,
+    lastDonation: userRecentTransactions[userRecentTransactions.length - 1],
+    nonce,
+    page
+  })
+  const smallestNonce: number = userRecentTransactions[
+    userRecentTransactions.length - 1
```
aminlatifi, line 203: What if the latest tx is a tx which the intended user is recipient of?

services/transactionService.ts
```diff
@@ -0,0 +1,554 @@
...
+      }
+    )
+    throw new Error(errorMessages.TRANSACTION_NOT_FOUND)
+  }
+  const foundTransaction = userRecentTransactions.find(
+    tx => tx.nonce === input.nonce
+  )
+  if (
+    foundTransaction &&
+    foundTransaction.to.toLowerCase() === input.toAddress.toLowerCase() &&
```
aminlatifi, line 191: What about token transfers?

services/transactionService.ts
```diff
@@ -0,0 +1,554 @@
...
+    recentDonationsLength: userRecentTransactions.length,
+    lastDonation: userRecentTransactions[userRecentTransactions.length - 1],
+    nonce,
+    page
+  })
+  const smallestNonce: number = userRecentTransactions[
+    userRecentTransactions.length - 1
+  ].nonce as number
+
+  if (smallestNonce < nonce) {
```
aminlatifi, line 206: What if user tx has not mined yet?

services/transactionService.ts
```diff
@@ -0,0 +1,554 @@
...
+  ) {
+    console.log('fromAddress of donation is different', {
+      transaction,
+      input
+    })
+    throw new Error(
+      errorMessages.TRANSACTION_FROM_ADDRESS_IS_DIFFERENT_FROM_SENT_FROM_ADDRESS
+    )
+  }
+  if (transaction) {
```
aminlatifi, line 288: I suggest finding the transaction first (hash or nonce matches) and then doing the verification

aminlatifi (commented):
For tx verification first we should be sure the tx has gone through (the nonce is consumed) and then do the searching. Also, for token transfers we can do it simpler by searching transfer events which to and from matches (web3 has that capability of getting past events of token smart contract by filtering to and from, since these two fields are indexed)

aminlatifi (changes requested):
Needs change as mentioned by other comments

package.json
mohammadranjbarz: done

services/transactionService.ts
mohammadranjbarz: done

services/transactionService.ts
mohammadranjbarz: It doesnt matter, because out utility function extract toAddress even for token transfers, so we would have it

services/transactionService.ts
mohammadranjbarz: Then we don't change the transaction status and it would be checked in next cycle

MoeNick:
Do this PR has iny issue?

mohammadranjbarz:
> Do this PR has iny issue?

Yeah it's in the description

mohammadranjbarz:
Just closed this PR and use #166  instead

PR #166, opened 2021-10-18
Verify donation transaction

mohammadranjbarz:
related https://github.com/Giveth/impact-graph/issues/168

Needed configs

```
BLOCKSCOUT_API_URL=https://blockscout.com/xdai/mainnet/api
ETHERSCAN_MAINNET_API_URL=https://api.etherscan.io/api
ETHERSCAN_ROPSTEN_API_URL=https://api-ropsten.etherscan.io/api
ETHERSCAN_API_KEY=
XDAI_NODE_HTTP_URL=https://dry-small-sound.xdai.quiknode.pro
INFURA_API_KEY=
NUMBER_OF_VERIFY_DONATION_CONCURRENT_JOB=1
VERIFY_DONATION_CRONJOB_EXPRESSION=0 0 * * * *
```


package.json
```diff
@@ -111,6 +111,9 @@
       "pre-push": "lint-staged"
     }
   },
+  "standard": {
+    "env": ["mocha"]
```
mohammadranjbarz: Adding this prevents showing invalid tslint warnings in test files
https://stackoverflow.com/a/38667441/4650625

provider.ts
```diff
@@ -1,34 +1,82 @@
...
+
+export function getProvider(networkId: number) {
+  const network = NETWORK_ID_MAP[networkId]
+  if (network === NETWORK_NAMES.XDAI) {
     return new ethers.providers.JsonRpcProvider(
       config.get('XDAI_NODE_HTTP_URL') as string
     )
   }
   // 'https://bsc-dataseed.binance.org/'
-  if (network === 'bsc') {
```
mohammadranjbarz, line 27: Replace these hardcode things with constant variable

services/transactionService.ts
```diff
@@ -1,70 +1,217 @@
-import Web3 from 'web3'
```
mohammadranjbarz: I changed the order of functions to use https://dzone.com/articles/the-stepdown-rule for increase readability of code

aminlatifi (commented):
Still we are checking two times and it's not necessary, also added complexity and reduces code readability
We have txHash and nonce, so find a tx that math to any of these two. At the end we will check other values (to, from, amount)

services/transactionService.ts
```diff
@@ -1,70 +1,217 @@
...
+        page,
+        networkId: input.networkId,
+        contractAddress
+      })
+    : await getListOfNormalTransactionsByAddress({
+        address: input.fromAddress,
+        page,
+        networkId: input.networkId
+      })
+  if (isTransactionListEmpty) {
```
aminlatifi, line 75: What if the first fetched list are all receive transactions (to == userAddress)

aminlatifi (changes requested):
Still we are checking two times and it's not necessary, also added complexity and reduces code readability
We have txHash and nonce, so find a tx that math to any of these two. At the end we will check other values (to, from, amount)

By the way, the code is still complex and hard to follow

services/transactionService.ts
mohammadranjbarz: In here https://github.com/Giveth/impact-graph/blob/check-donations-with-network-revision2/services/transactionService.ts#L244 
I check if the transaction list is empty ( Doesn't matter our address is source or destination) then we return it true, otherwise, we will check another batch of transactions

mohammadranjbarz:
> Still we are checking two times and it's not necessary, also added complexity and reduces code readability We have txHash and nonce, so find a tx that math to any of these two. At the end we will check other values (to, from, amount)

I would try to simplify it, but it's good if we have another call to talk about It to how make the logic simple

MoeNick:
Would u pls connect it to required issues as well @mohammadranjbarz ?

mohammadranjbarz:
> Would u pls connect it to required issues as well @mohammadranjbarz ?

Sure,  it's connected to #154 
and that PR is connected to https://github.com/Giveth/GIVeconomy/issues/62, https://github.com/Giveth/GIVeconomy/issues/81

mohammadranjbarz:
@aminlatifi 
Could you please review this PR again 
I think it's more readable now  ( I changed getting transaction from etherscan logic especially when finding transaction by nonce, as we talked)

services/transactionService.ts
```diff
@@ -0,0 +1,249 @@
...
+export async function getTransactionInfoFromNetwork(
+  input: TransactionDetailInput,
+): Promise<NetworkTransactionInfo> {
+  const { networkId, nonce } = input;
+
+  const web3 = getNetworkWeb3(networkId);
+  const userTransactionsCount = await web3.eth.getTransactionCount(
+    input.fromAddress,
+  );
+  if (nonce && userTransactionsCount <= nonce) {
```
aminlatifi: `nonce` is number and can have 0 value, then this condition will be always false

services/transactionService.ts
```diff
@@ -0,0 +1,249 @@
...
+  );
+
+  if (foundTransaction) {
+    return findTransactionByHash({ ...input, txHash: foundTransaction.hash });
+  }
+
+  // userRecentTransactions just includes the transactions that source is our fromAddress
+  // so if the lowest nonce in this array is smaller than the sent nonce we would know that we should not
+  // check latest transactions
+  const smallestNonce: number = userRecentTransactions[
```
aminlatifi: userRecentTransactions length can be zero

services/transactionService.ts
```diff
@@ -0,0 +1,249 @@
...
+      errorMessages.TRANSACTION_TO_ADDRESS_IS_DIFFERENT_FROM_SENT_TO_ADDRESS,
+    );
+  }
+
+  if (transaction.from.toLowerCase() !== input.fromAddress.toLowerCase()) {
+    throw new Error(
+      errorMessages.TRANSACTION_FROM_ADDRESS_IS_DIFFERENT_FROM_SENT_FROM_ADDRESS,
+    );
+  }
+  if (transaction.amount !== input.amount) {
```
aminlatifi, line 239: It's error-prone but cannot fix it by a simple solution. The `amount` data type is defined as `number` and changing it is laborous.

aminlatifi (changes requested):
Thank you @mohammadranjbarz, a big job
Please fix those few issues and after that, it would be OK.

services/transactionService.ts
mohammadranjbarz: It has been fixed in #185

services/transactionService.ts
mohammadranjbarz: it has been fixed in #185

PR #167, opened 2021-10-18
Feature: Add projects endpoint pagination, search and filters

CarlosQ96:
This PR moves the frontend logic to the backend. Using the activerecord pattern as the model takes care of the queries.
Changes:
1. The query is backward compatible (if no params are sent the same response as before is delivered).
2. Pagination bug fixed, the query was using limit and take which is redundant, also "take" is done in-memory. Replaced them with limit and offset completely.
3. FrontEnd Sorts were added to the backend (verified, amount raised, hearts, donations, recentlyAdded and OldProjects).
4. Category search was also added.

Missing: There is one traceable project (dapp) which appears to be some additional logic in the frontend that impact graph doesn't have. Would like to discuss this point.


mohammadranjbarz:
@CarlosQ96 
It works great, the code seems good to me, but the sot based on hearts doesn't work, maybe because you sort based on project.totalHearts but you should find it in reactions tables, 
If you see  below images you can see the sort doesn't work
you can test it with this query

```
query {
  projects(take: 0, skip: 1, searchTerm: "", category: "",
    orderBy: { field:Hearts , direction: DESC }) {
    reactions{
      reaction
    }
  }
}
```

Screen Shot 1400-07-26 at 17 45 11

Screen Shot 1400-07-26 at 17 45 02

mohammadranjbarz:
@CarlosQ96 
As @RamRamez  mentioned we have problem in in using `take` parameters:
(Adding take parameter return us less items, and I think it's weird

Screen Shot 1400-08-11 at 15 36 01


Screen Shot 1400-08-11 at 15 36 24

CarlosQ96:
I did some investigation before, after testing,  I think it's related to Typeorm limit and offset bug when used with joins.
I would replace offset and limit with the typeorm method Take/Skip @mohammadranjbarz

RamRamez:
@CarlosQ96 It seems the `totalCount` is always a fixed number. It should depend on the query, since I need the total count for the number of pages in the pagination component.

image

mohammadranjbarz:
> @CarlosQ96 It seems the `totalCount` is always a fixed number. It should depend on the query, since I need the total count for the number of pages in the pagination component.
> 
> image

@RamRamez  I think it's ok, it should be the number of all records that match with your query
As you see in below images  it's ok, so I can't get your point ( what result do you expect?)

Screen Shot 1400-08-11 at 20 11 57
Screen Shot 1400-08-11 at 20 12 06

RamRamez:
@mohammadranjbarz The problem is with `take`

image

mohammadranjbarz:
> @mohammadranjbarz The problem is with `take`
> 
> image

Whats the problem with `take`?
you asked for 5 projects, it gave it to you, and also gave you the number of all project , so you can decide to show how many pages,
You say in this case what should be the `totalCoun`t?

RamRamez:
@mohammadranjbarz Yes you're right, I think total count is correct.

CarlosQ96:
Correct the totalCount is always the maximum amount of records, regardless of queries. That's what Typeorm returns.

BTW I am investigating the pagination with take/skip.
Under the hood I was using limit/offset, but typeorm have bugs when using those with joins (unexpected results).
Now I switched to take/skip, but it's not as performant.

CarlosQ96:
https://github.com/typeorm/typeorm/issues/3967 && https://github.com/typeorm/typeorm/issues/4742

mohammadranjbarz (approved):
Thanks @CarlosQ96 
Great work.

PR #173, opened 2021-10-20
F 311 project by address support case insensitive

mohammadranjbarz:
related to Giveth/giveth-next#311

you can test it with this query:

```
query{
  projectByAddress(address:yourWalletAddress){
  	title,
    description,
    slug,
    walletAddress
  }
}
```
and change `yourWalletAddress` to lowercase you should get the result after that


CarlosQ96 (approved):
Tested the queries with varios projects and address and everything looks fine!

PR #175, opened 2021-10-22
F 317 prevent repetitive titles for projects

mohammadranjbarz:
related to Giveth/giveth-next#317

### DIFF
--- a/resolvers/projectResolver.ts
+++ b/resolvers/projectResolver.ts
@@ -366,6 +366,10 @@ export class ProjectResolver {
       !!imageUpload,
       heartCount
     )
+    if(newProjectData.title !== project.title){
+      await validateProjectTitle(newProjectData.title)
+    }
+
     const slugBase = slugify(newProjectData.title)
     const newSlug = await this.getAppropriateSlug(slugBase)
     if (project.slug !== newSlug && !project.slugHistory?.includes(newSlug)){
@@ -533,6 +537,7 @@ export class ProjectResolver {
       throw new Error(errorMessages.CATEGORIES_LENGTH_SHOULD_NOT_BE_MORE_THAN_FIVE)
     }
     await validateProjectWalletAddress(projectInput.walletAddress as string)
+    await validateProjectTitle(projectInput.title)
     const slugBase = slugify(projectInput.title)
     const slug = await this.getAppropriateSlug(slugBase)
     const status = await this.projectStatusRepository.findOne({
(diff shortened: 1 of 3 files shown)


mohammadranjbarz:
For testing this you should use this query

```
{
  isValidTitleForProject(title:"test Upload      pHoto")
}
```

the checking is case insensitive and escapes spaces and checks to makes sure that there is not special characters in title

mohammadranjbarz:
@CarlosQ96 
I have implemented a new query for validating project titles:

```
{
  isValidTitleForProject(title:"test Upload*&      pHoto")
}
```
and also I validate title before creating and updating projects
This is the test scenario, please consider the test flow while reviewing

https://github.com/Giveth/giveth-next/issues/317#issuecomment-950276960

CarlosQ96:
Tested this scenarios and they all passed. Tried a bunch of special characters and all good:
- [x]  Title can not be repetitive
- [x] we would check the case insensitive, we if we have a project with title test we assume Test is invalid
- [x] if user put space between words we drop them when comparing ( For instance if we have a project with title test two we assume test two is invalid
- [x]  We just let the user use letters and numbers and space, no special character is permitted
- [x]  We check all this stuff in creating and updating projects

CarlosQ96:
Just a quick question, should we allow the letter accents like 'á' and letters like Ñ for other languages? Maybe a spanish project for example. We can discuss this tomorrow. (they would be cleaned up in the slug tho)

mohammadranjbarz:
> Just a quick question, should we allow the letter accents like 'á' and letters like Ñ for other languages? Maybe a spanish project for example. We can discuss this tomorrow. (they would be cleaned up in the slug tho)

@MoeNick  @aminlatifi 
I think we should just allow English letters, 
@aminlatifi  @MoeNick  What's your idea about what Carlos said,

aminlatifi:
Good question, @MoeNick I guess it needs to be discussed by users (stakeholders) and be supported in front end too

MoeNick:
> Just a quick question, should we allow the letter accents like 'á' and letters like Ñ for other languages? Maybe a spanish project for example. We can discuss this tomorrow. (they would be cleaned up in the slug tho)

If we support other languages, we have to do it's own features, it's not about just Spanish. If we open it, why we don't do it for others, assume we may allow Chinese alphabet or not? And how we choose? based on our target market? Projected amount of donations? Populations? Demographics on blockchain users worldwide? 

For example some other languages are right-left which may change the whole platform is selected.

So I thinks it's not about just adding some characters, we have to bring language capabilities in our backlog and decide when to have them gradually.

mohammadranjbarz:
Thanks guys
As @MoeNick said, currently we just support the English letter, so I merge this PR.

PR #176, opened 2021-10-22
Check case insensitive uniqueness when valiadting walletAddress

mohammadranjbarz:


utils/validators/projectValidator.ts
```diff
@@ -15,7 +15,11 @@ export const validateProjectWalletAddress = async (walletAddress: string) :Promi
   if (isSmartContractWallet) {
     throw new Error(`Eth address ${walletAddress} is a smart contract. We do not support smart contract wallets at this time because we use multiple blockchains, and there is a risk of your losing donations.`)
   }
-  const projectWithAddress = await Project.findOne({walletAddress})
+  const projectWithAddress = await Project.createQueryBuilder('project')
+    .where(`lower("walletAddress")=lower(:walletAddress)`,{
```
CarlosQ96, line 19: I think it's good for now and we can leave it like this. 
However, for the future, now that I am reading a bit about optimizing queries, we could do an Index on the db with the walletAddresses in lowercase. This would make fast queries when we have thousands of projects. 
I am sure for now it's enough, since we must first fix the migration files that are broken.

CarlosQ96: We can discuss tomorrow.

CarlosQ96 (approved):

PR #178, opened 2021-10-24
add usd value

divine-comedian:


divine-comedian:
seems to be breaking because of missing circles conifg... not sure how to resolve this

resolvers/donationResolver.ts
```diff
@@ -134,18 +134,45 @@ export class DonationResolver {
...
+
+      const baseTokens =
+        Number(priceChainId) === 1 ? ['USDT', 'ETH'] : ['WXDAI', 'WETH'];
+
+      const tokenValues = getTokenPrices(
+        token,
+        baseTokens,
+        Number(priceChainId)
+      )
+        .then(async (prices: number[]) => {
```
mohammadranjbarz: instead of using .then() you can use async-await it makes code more readable

resolvers/donationResolver.ts
```diff
@@ -134,18 +134,45 @@ export class DonationResolver {
...
         anonymous: !!userId
-      })
-      await donation.save()
+      });
+      await donation.save();
+
+      const baseTokens =
+        Number(priceChainId) === 1 ? ['USDT', 'ETH'] : ['WXDAI', 'WETH'];
+
+      const tokenValues = getTokenPrices(
```
mohammadranjbarz: Name of the funciton is not good enough, if I read this I think it's a field not a function, you can rename it `getTokenValues`

resolvers/donationResolver.ts
```diff
@@ -134,18 +134,45 @@ export class DonationResolver {
...
+          throw new Error(e);
+        });
       // Logged in
       if (ctx.req.user && ctx.req.user.userId) {
-        userId = ctx.req.user.userId
-        originUser = await User.findOne({ id: userId })
+        userId = ctx.req.user.userId;
+        originUser = await User.findOne({ id: userId });
+
+        await tokenValues;
```
mohammadranjbarz: I think you forgot to add `()` at the end of the line, this line would not call the tokenValues function

resolvers/donationResolver.ts
```diff
@@ -158,83 +185,60 @@ export class DonationResolver {
...
-
-    const tokenValues = getTokenPrices(token, baseTokens, Number(priceChainId))
-        .then(async (prices: number[]) => {
-          // console.log(`prices : ${JSON.stringify(prices, null, 2)}`)
+      const projectOwner = await User.findOne({ id: Number(project.admin) });
 
-          donation.priceUsd = Number(prices[0])
-          donation.priceEth = Number(prices[1])
+      if (projectOwner) {
+        await tokenValues;
```
mohammadranjbarz: maybe should chanage to `await tokenValues()`

resolvers/donationResolver.ts
```diff
@@ -158,83 +185,60 @@ export class DonationResolver {
           transactionId: transactionId.toString().toLowerCase(),
           transactionNetworkId: Number(transactionNetworkId),
           currency: token,
+          donationValueUsd: donation.valueUsd,
+          donationValueEth: donation.valueEth,
           createdAt: new Date(),
           toWalletAddress: toAddress.toString().toLowerCase(),
           fromWalletAddress: fromAddress.toString().toLowerCase(),
```
mohammadranjbarz: `fromWalletAddress` and `toWalletAddress` are string already so it's not necessary to call `.toString()`, in ther other part of codes that these field have `.toString()`, they are not necessary as well

PR #181, opened 2021-10-26
Refactor transak stuff and make code more funcitonal

mohammadranjbarz:
The feature/transak_integration was great, but it was implemented by Object oriented approach and it is too class based,  so I created this PR to make it functional and simplify the code


resolvers/donationResolver.ts
```diff
@@ -135,8 +135,8 @@ export class DonationResolver {
         createdAt: new Date(),
         toWalletAddress: toAddress.toString().toLowerCase(),
         fromWalletAddress: fromAddress.toString().toLowerCase(),
-        transakId: transakId,
-        transakStatus: transakStatus,
```
mohammadranjbarz, line 139: Is this safe to get transakStatus from user?
user can send `COMPLETED` while creating
I suggest we should  either set the transak status ourself or don't set in saveDonation mutation

CarlosQ96: This is a posible outcome. We should probably skip the initial status and set it ourselves.

mohammadranjbarz: Ok, so I will change it not save in creating a donation
but I don't change the query Arg to our query be backward compatible for frontend

CarlosQ96: I also think we shouldn't notify segment initially if it's a transak Donation as it's never in the completed status the first time.

services/transak/donationService.ts
```diff
@@ -0,0 +1,51 @@
...
+    donation.amount = transakData.webhookData.cryptoAmount
+  }
+
+  if (donation.toWalletAddress.toLowerCase() !== transakData.webhookData.walletAddress.toLowerCase()){
+    donation.toWalletAddress= transakData.webhookData.walletAddress
+    const project = await Project.findOne({
+      walletAddress: transakData.webhookData.walletAddress
+    })
+    // we should check the walletAddress is matched with what is in donation, ir prevents fraud
+    donation.projectId = project?.id || 0
```
CarlosQ96: What happens if they don't match? should we raise an exception?

mohammadranjbarz: Currently I think we can 0 for projectId, in the future we can decide about that

PR #185, opened 2021-10-31
Make difference when nonce is undefined or zero , in verying transactions

mohammadranjbarz:


services/transactionService.ts
```diff
@@ -24,7 +24,7 @@ export async function getTransactionInfoFromNetwork(
   const userTransactionsCount = await web3.eth.getTransactionCount(
     input.fromAddress,
   );
-  if (nonce && userTransactionsCount <= nonce) {
+  if ( nonce !== undefined && userTransactionsCount <= nonce) {
```
aminlatifi: It can be null !! null is different than undefined
```suggestion
  if ( typeof nonce !== 'number' && userTransactionsCount <= nonce) {
```
can be a solution

aminlatifi (changes requested):

services/transactionService.ts
mohammadranjbarz: @aminlatifi  Yeah you are right, it's better, done

PR #187, opened 2021-11-01
Persist transak transactionLink in DB after verifying by transak

mohammadranjbarz:
related to Give/giveth-next#308

@CarlosQ96  please see https://github.com/Giveth/giveth-next/issues/308#issuecomment-956028472


CarlosQ96 (approved):
Great! Just Reviewed and tested the webhook and indeed transak sends the transactionLink ONLY when it's completed, so we can persist it in our db!
Also tested if the undefined / null value is saved without issues while it's not completed and works fine!

Issue #170, opened 2021-10-03
Return all data that front need for project detail page in an endpoint


mohammadranjbarz:
In giveth-next when leading the project detail page the client call multiple requests continuous to get the page needed information you can see the front code here

https://github.com/Giveth/giveth-next/blob/mateo-mig/pages/project/%5Bslug%5D.js#L45-L91

@RamRamez  wants a query from backend then we can return all needed data for him, then he can just call one endpoint and then render the page.
@CarlosQ96  if you have any solution or you can implement it or you think it's not a good solution, please help us

CarlosQ96:
After checking the project details, projects with 900 donations like Giveth are taking over 3.8-4.5 seconds to load all that data. Now imagine if it had 400 project updates and reactions, this would keep getting worse. 

Even if we move all to server side, we would be loading a lot of data in-memory, which isn't the best idea either under heavy traffic.

CarlosQ96:
I have a few thoughts @RamRamez @mohammadranjbarz:

First, I think unifying everything into one endpoint is not the ideal way, as reactions, project updates and donations can grow indefinitely. I remember Mateo back in July removed the donations from the query because it was causing timeout errors on the impact graph.
Maybe if we upgrade to piñata pro plan this won't be an issue, not sure if @geleeroyale did any server upgrade that fixed the issue, or the recent frontend optimizations alleviated the problem.

Another point against unifying them, is that adding pagination in the same endpoint for each entity would be more complex and lead to retrieving redundant data and remapping said info every-time.

CarlosQ96:
@MoeNick @mohammadranjbarz @RamRamez 
My two proposals are:
1. We keep them separated, but unify some, like so:
    - Project by slug will include the admin user. Right now Giveth-Next makes 1 extra request to retrieve 1 user which is redundant and adds latency.
    - Add pagination to the donations endpoint, we shouldn't need to retrieve the 900+ records in one go. This can include the totalCount to calculate the pages and the total ethAmount.
    - We leave the projectUpdates/reactions as is, I'll leave the pagination optional. There is a lot of logic in the view regarding this, potentially lots of breaking changes, so not sure if combining them is a good idea.
    - Add cache to projectBySlug endpoint.

2. **(Don't recommend)** Considering we are close to launch: We unify them into one graphql query, and add cache for subsequent fast results, moving all the loading to the server. 
[...]

PR #191, opened 2021-11-03
Feature: Add pagination to Donations and Admin Query

CarlosQ96:
Optimize donations endpoint by adding pagination (Used Limit and Offset because we do not have joins)
Add AdminUser to projectBySlug query.
Will keep make them separated queries to prevent breaking changes.

Project By Slug with the admin included:
```
query{
  projectWithAdminBySlug(slug:"giveth"){
    project {
      slug
    	id
    	title
    }
    admin {
      id
      firstName
    }
  }
}
```
PaginatedDonations With ethBalance
TotalCount and paginated values.
```
query{
  donationsByDestinationWallets(take: 2, skip: 0, toWalletAddresses: ["134"]){
    donations {
      id
    }
    totalCount
    totalEthBalance
  }
}
```
EthBalance is calculated in the GivethNext frontend, would like to ask if this is correct. Else we have to change it in the frontend too.
In the DonatorView
image


mohammadranjbarz:
@CarlosQ96 
It seems good to me, but is it for what issue?
I want to see if it is ok with that issue

CarlosQ96:
Issue #170, kept the queries separated for no breaking changes. So when Ramin improves the front then we can switch to this endpoint to further optimize a bit.

Issue #193, opened 2021-11-07
Bring back  totalDonations and totalHearts fields and make them be updated


mohammadranjbarz:
We need to add these fields in the project entity, then when we want to sort by hearts or donations we can simply use it ( performance is much higher than joining when sorting by donations), and it's good when frontend wants to show the number of hearts of projects, and amount of donations

MoeNick:
@mohammadranjbarz appreciate if you can update us on this.

mohammadranjbarz:
> @mohammadranjbarz appreciate if you can update us on this.

Yes we have it now, you can close this issue

PR #194, opened 2021-11-07
F 193 add total reactions and total donations fields to project

mohammadranjbarz:
related #193 

@CarlosQ96  I added `totalDonations` and `totalHearts` as we talked, so because now we have these fields I changed the `sortBy` queries and make them simpler, 
These changes are deployed on https://dev.serve.gievth.io 
so you can run `giveth-next` locally and connect to dev server and after like/unlike, or donating to a project the `totalHearts` and `totalDonations` should be updated


entities/project.ts
```diff
@@ -155,6 +155,14 @@ class Project extends BaseEntity {
   @RelationId((project: Project) => project.status)
   statusId: number;
 
+  @Field(type => Float)
+  @Column({ type: 'real', default: 0 })
+  totalDonations: number = 0;
+
+  @Field(type => Float)
+  @Column({ type: 'real', default: 0})
+  totalHearts: number = 0;
```
CarlosQ96: I think we shouldn't name it totalHearts, more like "totalReactions" as we might have more types of reactions in the future and that way it would be more consistent with the use of terms

services/donationService.ts
```diff
@@ -69,3 +72,19 @@ const notifyTransakUpdate = async donation => {
     }
   }
 };
+
+export const updateTotalDonationsOfProject = async (projectId: number) => {
+  try {
+    const donationsAmount = await Donation.query(
+      `select COALESCE(SUM("valueUsd"),0) as total from donation where "projectId" = ${projectId}`,
```
CarlosQ96: This is just conventions in sql, writing the commands capitalized and separated per line.
```javascript
SELECT COALESCE(SUM("valueUsd"),0) AS total
FROM donation
WHERE "projectId" = ${projectId}
```

CarlosQ96 (commented):
Just some small suggestions let me know what you think.

entities/project.ts
```diff
@@ -254,9 +234,9 @@ class Project extends BaseEntity {
 
     // Sorts
     if (sortBy === OrderField.Reactions) {
-      this.addReactionsCountQuery(query, direction);
+      query.orderBy(`project."totalHearts"`, direction);
     } else if (sortBy === OrderField.Donations) {
-      this.addTotalDonationsQuery(query, direction);
+      query.orderBy(`project."totalDonations"`, direction);
```
CarlosQ96: I think since this are columns in the same entity instead of a relation. We can use the "else" query that orders by the attribute instead of another if.

CarlosQ96: Changing the Enum to
```
export enum OrderField {
  CreationDate = 'creationDate',
  Balance = 'balance',
  QualityScore = 'qualityScore',
  Verified = 'verified',
  Reactions = 'totalReactions',
  Donations = 'totalDonations'
}
```

migration/1636314323228-UpdateTotalHearts.ts
```diff
@@ -0,0 +1,20 @@
...
+        `select count(*) from reaction where "projectId"=${project.id}`,
+      );
+      await queryRunner.query(
+        `update project set "totalHearts"=${totalHearts[0].count} where id=${project.id}`,
+      );
+    }
+  }
+
+  async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`update project set "totalHearts"=0`);
```
CarlosQ96: We would rename the migration also with totalReactions

entities/project.ts
CarlosQ96: We would change this also in the migration.

services/donationService.ts
mohammadranjbarz: Thanks Carlos, , I wasn't aware about conventions, now I changed it

entities/project.ts
mohammadranjbarz: Yeay it makes sense, 
done

migration/1636314323228-UpdateTotalHearts.ts
mohammadranjbarz: Done.

entities/project.ts
mohammadranjbarz: Very good point, done.

CarlosQ96 (approved):
LGTM! 
Thanks Mohammad!

PR #195, opened 2021-11-08
Change query of donations, from walletAddresss to projectId

mohammadranjbarz:
I've changed query of getting donations, you can test it like this :

```
query{
  donationsByProjectId(take: 2, skip: 0, projectId:2){
    donations {
      id
      valueUsd
    }
    totalCount
    totalUsdBalance
  }
}
```
@CarlosQ96  If you think it's ok. please merge it to your branch

### DIFF
--- a/resolvers/donationResolver.ts
+++ b/resolvers/donationResolver.ts
@@ -83,22 +83,16 @@ export class DonationResolver {
     });
     return donations;
   }
-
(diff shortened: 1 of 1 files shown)


CarlosQ96:
Thanks! I agree by ID is less prone to error and faster. I'll Merge it.

PR #196, opened 2021-11-09
Updates monoswap for new tokens prices

mateodaza:


mohammadranjbarz (approved):
Because the changes are mainly in monoswap library and here just we update this package, I would merge it ( I think the changes were adding some token so we don't have any breaking changes)

PR #192, opened 2021-11-07
Feature: Add cronjob to update project listing

CarlosQ96:
This PR will setup a cronjob when the server runs that executes every day a check on projects creationDate to see which haven't been reviewed for more three weeks, and proceeds lists them.

This includes the last day of the three weeks as the day of listing.

Also configurable through a env vars:
MAXIMUM_DAYS_FOR_LISTING_PROJECTS: 21 by default
VERIFY_DONATION_CRONJOB_EXPRESSION: * * * * * (format) 

Use this website to know how to configure the cronjob format date for running the task: https://crontab.guru/

PR #198, opened 2021-11-10
Feature: Add cronjob to update project listing revision2

mohammadranjbarz:
Make some changes on  https://github.com/Giveth/impact-graph/pull/192
(The functionality of #192 was ok)

### DIFF
--- a/services/segment/projectTracker.ts
+++ b/services/segment/projectTracker.ts
@@ -1,19 +1,8 @@
 import { User } from '../../entities/user';
 import { Project } from '../../entities/project';
-import { getAnalytics } from '../../analytics';
+import { getAnalytics, SegmentEvents } from '../../analytics';
 
 const analytics = getAnalytics();
-
-export enum SegmentEvents {
-  PROJECT_LISTED = 'Project listed',
-  PROJECT_UNLISTED = 'Project unlisted',
-  PROJECT_EDITED = 'Project edited',
-  PROJECT_VERIFIED = 'Project verified',
-  PROJECT_UNVERIFIED = 'Project unverified',
-  PROJECT_ACTIVATED = 'Project activated',
-  PROJECT_DEACTIVATED = 'Project deactivated',
-  PROJECT_CANCELLED = 'Project cancelled',
-}
 
 /**
  * Notifies Segment any event concerning the project
--- a/services/segment/projectTracker.ts
+++ b/services/segment/projectTracker.ts
@@ -4,15 +4,26 @@ import { getAnalytics } from '../../analytics';
 
 const analytics = getAnalytics();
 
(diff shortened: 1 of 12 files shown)


services/syncDonationsWithNetwork.ts
```diff
@@ -16,7 +16,7 @@ const numberOfVerifyDonationConcurrentJob =
   Number(config.get('NUMBER_OF_VERIFY_DONATION_CONCURRENT_JOB')) || 1;
 
 const cronJobTime =
-  String(config.get('VERIFY_DONATION_CRONJOB_EXPRESSION')) || '0 0 * * * *';
```
mohammadranjbarz, line 19: @CarlosQ96 
I understood it's tricky because `String(undefined)` return `"undefined"`  so I changed it to `config.get('VERIFY_DONATION_CRONJOB_EXPRESSION') as string`

services/syncProjectsRequiredForListing.ts
```diff
@@ -2,31 +2,33 @@ import { Project } from '../entities/project';
 import { schedule } from 'node-cron';
 import { getRepository } from 'typeorm';
 
-// @ts-ignore
 import config from '../config';
 
 const cronJobTime =
-  String(config.get('VERIFY_DONATION_CRONJOB_EXPRESSION')) || '0 0 * * *';
+  (config.get('MAKE_UNREVIEWED_PROJECT_LISTED_CRONJOB_EXPRESSION') as string) ||
```
mohammadranjbarz, line 9: I think it's better to have different configs for different cron jobs

mohammadranjbarz:
@divine-comedian 
I added a segment event for listing projects automatically, the name should be `Project listed` or you want something else?

services/segment/projectTracker.ts
```diff
@@ -4,15 +4,26 @@ import { getAnalytics } from '../../analytics';
 
 const analytics = getAnalytics();
 
+export enum SegmentEvents {
```
mohammadranjbarz: @CarlosQ96  It's not related to this issue, but I thought maybe it's good to use enum instead of string :)))

services/syncDonationsWithNetwork.ts
CarlosQ96: Perfect I didn't know that. Thanks!

services/syncProjectsRequiredForListing.ts
CarlosQ96: Whoops I thought I changed it. Thanks

services/segment/projectTracker.ts
CarlosQ96: Good idea.

CarlosQ96 (approved):

PR #202, opened 2021-11-17
Dispatch update event when edit project in view page in admin bro

mohammadranjbarz:
it's related to 
https://github.com/Giveth/giveth-next/issues/364#issuecomment-969044875


CarlosQ96 (approved):
Good idea to use the after hook on adminbro to enqueue the trace request.
I think this is fine.

PR #209, opened 2021-11-24
Feature/add project update count

CarlosQ96:
Add the total project updates to the project entity so it's queryable for the new design.

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -168,6 +168,10 @@ class Project extends BaseEntity {
   @Column({ type: 'integer', nullable: true })
   totalReactions: number;
 
+  @Field(type => Int, { nullable: true })
+  @Column({ type: 'integer', nullable: true })
+  totalProjectUpdates: number;
+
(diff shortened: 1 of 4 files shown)


CarlosQ96:
@mohammadranjbarz Tested locally and creating a project update so the count changes. However the dev server seems stuck running the migration while doing the deploy. 

Checking what could it be, or if you see there is something in the migration missing, let me know .

mohammadranjbarz (approved):
Thanks @CarlosQ96 
Seems good to me, 
I just changed migration file a little, to create column and delete it in migration down.

mohammadranjbarz:
> @mohammadranjbarz Tested locally and creating a project update so the count changes. However the dev server seems stuck running the migration while doing the deploy.
> 
> Checking what could it be, or if you see there is something in the migration missing, let me know .

You know in `totalDonations` we had the column from before so we didn't need to add column, but in this case we needed that

Issue #204, opened 2021-11-22
Add UpdatedAt / Last Updated to projects for new design


CarlosQ96:

CarlosQ96:
@RamRamez What actions modify the updatedAt? 
This is my list:
1. Editing the project
2. removing or adding a project update

optional in my opinion:
3. Adding a donation.

RamRamez:
> @RamRamez What actions modify the updatedAt? This is my list:
> 
> 1. Editing the project
> 2. removing or adding a project update
> 
> optional in my opinion: 3. Adding a donation.

@MoeNick @mohammadranjbarz What do you think?

PR #210, opened 2021-11-24
For new Design: Add project updatedAt

CarlosQ96:
This creates a new column UpdatedAt for keeping track of project updates:
1. It will be updated if the project is edited.
2. It will be updated if a projectUpdate record is added or removed.
3. Default value of the updatedAt will be the creationDate setup with a migration.

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -1,7 +1,9 @@
 import { Field, Float, ID, ObjectType } from 'type-graphql';
 import {
+  AfterInsert,
   AfterUpdate,
   BaseEntity,
+  BeforeRemove,
   Brackets,
   Column,
   Entity,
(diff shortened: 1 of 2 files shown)


CarlosQ96:
Per defined in the meeting: UpdatedAt will be the same as creationDate to begin.
Donating won't affect this date.

mohammadranjbarz:
> Per defined in the meeting: UpdatedAt will be the same as creationDate to begin. Donating won't affect this date.

We should not change `UpdatedAt` when user edits the project info?
or when the project info changes by admin bro?

CarlosQ96:
Yes it should update. The updatedAt has a onUpdate clause that handles it.

mohammadranjbarz:
> Yes it should update. The updatedAt has a onUpdate clause that handles it.

Oh great, I saw that now

mohammadranjbarz (approved):
Thanks @CarlosQ96 

It seems good to me

mohammadranjbarz:
related to #204

PR #211, opened 2021-11-24
Fix pagination with in-memory slice

CarlosQ96:
This would be a temporary measure because of all the typeorm bugs regarding using:
Joins + Take/skip pagination + OrderBy
There are multiple issues returning unexpected results.

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -249,10 +249,7 @@ class Project extends BaseEntity {
 
     query.orderBy(`project.${sortBy}`, direction);
 
-    const projects = query
-      .take(limit)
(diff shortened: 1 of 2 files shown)


mohammadranjbarz:
@CarlosQ96 
After our memory leak on `projects` query  I afraid of doing this, it may cause memory leak and decrease our performance,
So I think it's better to fix this with using `Peisma` ORM as you said before ( Maybe it's better to close this PR and create new one with using Prisma)
What do you think?

CarlosQ96:
Yes I agree, lets close this PR. @mohammadranjbarz

Issue #152, opened 2021-10-05
Add Segment events to bulk actions on adminbro


mateodaza:

mohammadranjbarz:
@divine-comedian 
it's ready for test

divine-comedian:
can you provide more information on how to test this - I should list/unlist multiple projects on adminBro and check they appear on segment?

mohammadranjbarz:
> can you provide more information on how to test this - I should list/unlist multiple projects on adminBro and check they appear on segment?

Yeah exactly, 
@CarlosQ96  Am I right?
could you add something if Mitch missed?

divine-comedian:
I have tested all of the actions (list/unlist/verify/unverify/activate/deactivate/cancel) in bulk and they do successfully appear in segment, however their seems to be a serious issue of not each instance arriving in segment. 

For examples when I bulk list 10 projects someitmes only 9 events show up in segment, other times only 8, even as few as 6 events show up. 

I tried lowering the number to bulk list/cancel/verify/etc..  5 events.. sometimes 5 show up... sometimes 4 events show up

so something weird is happening, perhaps its overloaded with having multiple requests at the exact same moment. 

Would it be helpful to add a short delay between each analytcis send if we do bulk actions like this?

CarlosQ96:
Hey @divine-comedian Ill double check if maybe we need to add a delay or an await so that it sends them all. Will let you know after I finish testing.

CarlosQ96:
I have a question @mohammadranjbarz if analytics.track is a call to an external api, shouldn't this be a async function? we have it as a normal function in the analytics.ts

Maybe it works in most calls, because it's not an expensive call and before the request cycle ends, it answers. But in this case multiple calls do not manage to finish before the server moves on to the next nonblocking task. 

https://github.com/segmentio/analytics-node/issues/303 this is not exactly the same issue but he does experience something similar.

CarlosQ96:
Actually after doing some testing, it seems analytics-node batches the requests in memory to send the track payload sometime in the future. (we allow batching in prod and staging). 

Subsequent requests from the batches.... seem to take a up to a few seconds max to be sent, however at the same time after 10 seconds the batch is flushed, so any message not sent is lost. _***(this is most likely the issue)***_
Documentation: https://segment.com/docs/connections/sources/catalog/libraries/server/node/#batching
Gonna keep testing and let you know guys know what solution I come up with. Probably we will have to add config options to analytics constructor.

[...]

PR #212, opened 2021-11-24
Add bulk events to segment when editing projects in admin bro

mohammadranjbarz:
related to #152

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -249,17 +249,28 @@ class Project extends BaseEntity {
 
     query.orderBy(`project.${sortBy}`, direction);
 
-    const projects = query
-      .take(limit)
-      .skip(offset)
-      .getMany();
+    const projects = query.take(limit).skip(offset).getMany();
     const totalCount = query.getCount();
 
     return Promise.all([projects, totalCount]);
   }
 
-  static notifySegment(project: any, eventName: SegmentEvents) {
+  static notifySegment(project: Project, eventName: SegmentEvents) {
     new ProjectTracker(project, eventName).track();
+  }
+
+  static sendBulkEventsToSegment(
+    projects: [Project],
+    eventName: SegmentEvents,
+  ) {
+    // TODO when sending bulk events, if number of events are more than 4, the segment misses some events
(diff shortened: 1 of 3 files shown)


CarlosQ96:
This works locally, because we have batching disabled on the analytics.ts file.
```
 options = {
      flushAt: 1,
    };
```
I tried with the staging configuration (removing the flush at). So the default is every 20 messages and every 10 seconds it flushes the batch. https://segment.com/docs/connections/sources/catalog/libraries/server/node/#batching
And the bug persisted from 10 messages it only sent 8.
So the bug is related to the http batching of the library.

@mohammadranjbarz I think we should disable batching on staging and merge this PR for the bulk and the issue should be fixed.

entities/project.ts
```diff
@@ -249,19 +249,30 @@ class Project extends BaseEntity {
...
+    eventName: SegmentEvents,
+  ) {
+    // TODO when sending bulk events, if number of events are more than 4, the segment misses some events
+    // So we have to put a delay between events, it works now, but it's not best practice for sure
+    const ONE_SECOND = 1000;
+    for (let i = 0; i < projects.length; i++) {
+      setTimeout(() => {
+        this.notifySegment(projects[i], eventName);
+      }, ONE_SECOND * i);
+    }
```
CarlosQ96, line 273: In the analytics.ts we can ADD this line to other environments like staging/dev/prod to disable batching.
```javascript
    options = {
        flushAt: 1
    }
```

CarlosQ96 (changes requested):
I think if we disable batching and use this delay for bulk it should work fine.
The bug of analytics library is:
1. Batching/sending to many messages too fast.
2. Batching more than 4 at the same time. (I'll be honest it even failed me once with 4, so a bit unsure after retesting).

entities/project.ts
CarlosQ96: Upon further testing even with the delay, in RARE cases it only sends 9 of the 10. But seems more reliable.

CarlosQ96 (approved):
LGTM!

PR #215, opened 2021-11-28
Send segment events in queue

mohammadranjbarz:
related to #152

### DIFF
--- a/analytics/analytics.ts
+++ b/analytics/analytics.ts
@@ -1,6 +1,6 @@
 import Analytics from 'analytics-node';
-import config from './config';
-import { User } from './entities/user';
+import config from '../config';
+import { User } from '../entities/user';
 
 export enum SegmentEvents {
   PROJECT_LISTED = 'Project listed',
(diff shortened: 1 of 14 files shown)


CarlosQ96 (approved):
Everything looks good to me.  

However I could reproduce the same issue where only 9/10 events were sent.
Seems they are working on the solution https://github.com/segmentio/analytics-node/issues/309, maybe it's this issue they will solve soon. (Some form of race condition when the library flushes messages)

Another thing we can try last is the async track implementation. to kind of force the promise to be solve, I tried this and got the same issue tho: https://github.com/segmentio/analytics-node/issues/303

I think we should proceed and wait for them to fix it. Unless you want to try something else.

mohammadranjbarz:
> Everything looks good to me.
> 
> However I could reproduce the same issue where only 9/10 events were sent. Seems they are working on the solution [segmentio/analytics-node#309](https://github.com/segmentio/analytics-node/issues/309), maybe it's this issue they will solve soon. (Some form of race condition when the library flushes messages)
> 
> Another thing we can try last is the async track implementation. to kind of force the promise to be solve, I tried this and got the same issue tho: [segmentio/analytics-node#303](https://github.com/segmentio/analytics-node/issues/303)
> 
> I think we should proceed and wait for them to fix it. Unless you want to try something else.

Thanks Carlos, I agree, we can merge it as temporary solution, till Segment library will be fixed

CarlosQ96:
@aminlatifi This is the analytics-node library that presents issues. We did a big workaround with queues and delays for it to be as stable as posible.

PR #219, opened 2021-11-30
Fix projects query

mohammadranjbarz:
The backend kept eating memory and restarting in production, after monitoring DB logs I found the `donations` join is very expensive so we have to delete that

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -241,6 +241,7 @@ class Project extends BaseEntity {
   ) {
     const query = this.createQueryBuilder('project')
       .leftJoinAndSelect('project.status', 'status')
+      // TODO It was very expensive query and made our backend down in production, maybe we should remove the reactions as well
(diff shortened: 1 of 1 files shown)


CarlosQ96 (approved):
I agree, I think @RamRamez should confirm with us first if he doesn't use that included data.

RamRamez:
@mohammadranjbarz @CarlosQ96 What query is removed?

The `donations` object in `projectBySlug`?

mohammadranjbarz:
> @mohammadranjbarz @CarlosQ96 What query is removed?
> 
> The `donations` object in `projectBySlug`?

No in `projects`
I think it's not needed in frontend because staging is up and running

CarlosQ96:
In the all projects query (with all the filters/searching etc) we removed the donations because it was causing a memory leak.

Issue #206, opened 2021-11-23
add number of updates in projectBySlug


RamRamez:
@mohammadranjbarz @CarlosQ96 Is it possible to add No. of updates in projectBySlug?

I need it for new design

![image](https://user-images.githubusercontent.com/73441857/143064483-03e2e62e-6e62-4fe5-8c20-bccb5bf6a381.png)

RamRamez:
@mohammadranjbarz @CarlosQ96 When I use the new query in PR #209 for [this](https://next.giveth.io/project/ram-second-metamask-2) project, it returns 3 for totalProjectUpdates.
But using getProjectUpdates, it returns 2 updates.

MoeNick:
@CarlosQ96 @MohammadPCh shall we bring it back?

MoeNick:
@CarlosQ96 Ramin said it shows +1 donation, I make it back to in-progress.

CarlosQ96:
@MoeNick @RamRamez Found the issue. The count is "correct" but we have in all projects a Hidden projectUpdate that is empty for some reason..... It's created as default ProjectUpdate when the project is created.
I'll update the code to ignore this.

mohammadranjbarz:
@RamRamez 
We have deployed it on staging, could you check it, please?

RamRamez:
> @RamRamez We have deployed it on staging, could you check it, please?

It seems good.

divine-comedian:
does this issue need further testing or can it be moved to done?

CarlosQ96:
I think we can move it to Done, this is all done. We keep the total project updates # updated at all times every time a new projectUpdate is added or deleted.

PR #227, opened 2021-12-01
Hotfix: Fix projectUpdateCount by ignoring projectUpdate with isMain as true

CarlosQ96:
The default projectUpdate with isMain = true is ignored. Added the code to ignore this and a migration. 

Issue: https://github.com/Giveth/impact-graph/issues/206


mohammadranjbarz (approved):
Thanks @CarlosQ96 
It seems good to me

PR #232, opened 2021-12-01
Hotfix: Fix inverted anonymous boolean

CarlosQ96:
We currently have this boolean inverted, Ramin seems to use this on the new design. So fixing it once and for all 👍

### DIFF
--- a/migration/1638386837934-FixAnonymousDonationBoolean.ts
+++ b/migration/1638386837934-FixAnonymousDonationBoolean.ts
@@ -5,20 +5,20 @@ export class FixAnonymousDonationBoolean1638386837934
 {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      'UPDATE donation SET anonymous = true WHERE "userId" IS NULL;',
+      'UPDATE donation SET anonymous = true WHERE "userId" IS NULL',
     );
     await queryRunner.query(
-      'UPDATE donation SET anonymous = false WHERE "userId" IS NOT NULL;',
+      'UPDATE donation SET anonymous = false WHERE "userId" IS NOT NULL',
     );
   }
 
   // Revert in case something happens, seems only the new design will use this
   public async down(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      'UPDATE "donation" SET anonymous = false WHERE "userId" IS NULL;',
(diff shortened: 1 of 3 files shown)


CarlosQ96:
@mohammadranjbarz Validated with Mateo that they are not using this boolean in the front.
Seems that Ramin will do in the new design, so fixing it now.

Checked the rest of the segment events and it seems they are using it correctly so nothing to change with segment.

migration/1638386837934-FixAnonymousDonationBoolean.ts
```diff
@@ -0,0 +1,24 @@
...
+    );
+    await queryRunner.query(
+      'UPDATE donation SET anonymous = false WHERE "userId" IS NOT NULL',
+    );
+  }
+
+  // Revert in case something happens, seems only the new design will use this
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      'UPDATE "donation" SET anonymous = false WHERE "userId" IS NULL',
```
mohammadranjbarz, line 18: @CarlosQ96  I think you should set anonymous `true` in this case

migration/1638386837934-FixAnonymousDonationBoolean.ts
```diff
@@ -0,0 +1,24 @@
...
+    );
+    await queryRunner.query(
+      'UPDATE "donation" SET anonymous = true WHERE "userId" IS NOT NULL',
```
mohammadranjbarz, line 21: @CarlosQ96  I think you should set anonymous `false` in this case

mohammadranjbarz (changes requested):

migration/1638386837934-FixAnonymousDonationBoolean.ts
CarlosQ96: I was thinking what would be the down method in this case? I was like reverting it to how it was  before. The up method is like you are refering.

mohammadranjbarz: Oh man sorry , my mistake
You are totally right

mohammadranjbarz (approved):
Seems good to me

mohammadranjbarz (approved):
Thanks @CarlosQ96 
LGTM

Issue #238, opened 2021-12-09
Call projectById query cause crash in backend


mohammadranjbarz:
```
query {
  projectById(id:1){
  	title
    slug
    totalProjectUpdates
  }
}
```

divine-comedian:
testing this out on graphql staging - 
![image](https://user-images.githubusercontent.com/67759413/147155713-96737e3c-88b0-4519-aced-c71098eccfdf.png)

I can successfully run this query and don't see any error - output looks good 

was there anything else to test on this issue @mohammadranjbarz ?

mohammadranjbarz:
> testing this out on graphql staging - ![image](https://user-images.githubusercontent.com/67759413/147155713-96737e3c-88b0-4519-aced-c71098eccfdf.png)
> 
> I can successfully run this query and don't see any error - output looks good
> 
> was there anything else to test on this issue @mohammadranjbarz ?

No that's it , the playground is disabled but backend is up
I think this testing is enough

PR #239, opened 2021-12-09
Remove donations relation from projectById query

mohammadranjbarz:
related to #238

joining donations to project table cause disaster in production ENV
so I have removed that relation

### DIFF
--- a/resolvers/projectResolver.ts
+++ b/resolvers/projectResolver.ts
@@ -320,7 +320,7 @@ export class ProjectResolver {
   async projectById(@Arg('id') id: number) {
     return await this.projectRepository.findOne({
       where: { id },
-      relations: ['donations', 'reactions'],
+      relations: ['reactions'],
     });
   }
 


CarlosQ96 (approved):
Agree! We ended up removing donations relation from all posible endpoints. 😆

mohammadranjbarz:
@RamRamez 
If you use donations for `projectById` query, be careful and change it please

mohammadranjbarz:
@RamRamez 
If you use donations for `projectById` query, be careful and change it please

mohammadranjbarz:
> @RamRamez If you use donations for `projectById` query, be careful and change it please

@RamRamez  I assume you don't use it, so I will merge this PR

RamRamez:
> > @RamRamez If you use donations for `projectById` query, be careful and change it please
> 
> @RamRamez I assume you don't use it, so I will merge this PR

I think we are not using `projectById` at all

CarlosQ96:
Mateo is using it in the last release. But donations is not being used, so it's alright.

PR #247, opened 2021-12-13
Feature/allow users to set anonymous donations

CarlosQ96:


mohammadranjbarz (approved):
@CarlosQ96 
Seems good to me

mohammadranjbarz:
related to #248

PR #254, opened 2021-12-16
Feature: Add acceptGiv filter to projects

CarlosQ96:
This is related to issue https://github.com/Giveth/giveth-next/issues/516
Only non giving blocks projects accept giv.


mohammadranjbarz (approved):
@CarlosQ96 Thanks, LGTM

Issue #257, opened 2021-12-20
Security Analysis


aminlatifi:
- [x] is there a reason we've got access-control-allow-origin: set to * in some places?
- [x] I would recommend setting X-Frame-Options: DENY or X-Frame-Options: SAMEORIGIN to prevent clickjacking (unless you expect to be hosted in a frame on another site, in which case I have a different suggestion)
- [x] It would be nice to suppress the "X-Powered-By: Next.js" header
- [ ] You might wanna consider stripping comments before launch
- [x] Would you be ok with adding "Strict-Transport-Security: max-age=10000; includeSubDomains; preload" in order to force all assets to load over HTTPS

geleeroyale:
ok - I solved these:
- It would be nice to suppress the "X-Powered-By: Next.js" header (by setting a value in next.config)
- Would you be ok with adding "Strict-Transport-Security: max-age=10000; includeSubDomains; preload" in order to force all assets to load over HTTPS (by setting it in impact-graph nginx)

this needs clarification:
- I would recommend setting X-Frame-Options: DENY or X-Frame-Options: SAMEORIGIN to prevent clickjacking (unless you expect to be hosted in a frame on another site, in which case I have a different suggestion)

I did set x-frame-options: deny on the backend, but of course the frontend can still be loaded via iframe ... can we somehow deny on the frontend as well, or is it sufficient to do it on the backend?

aminlatifi:
Created #259 PR to fix the first one

PR #260, opened 2021-12-21
hostname whitelist on cors revision2

mohammadranjbarz:
related #257, #258

@aminlatifi There are three points that we need to consider
1. When we write CORS middleware manually it would not answer the non-browser requests by default, so I made some changes
2. We should support subdomains otherwise if we put `vercel.app` we would not support preview links of vercel
3. We should format origin, remove protocol (http, https) and port from origin

### DIFF
--- a/server/bootstrap.ts
+++ b/server/bootstrap.ts
@@ -140,14 +140,31 @@ export async function bootstrap() {
     // Express Server
     const app = express();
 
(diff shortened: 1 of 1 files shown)


aminlatifi:
Thanks @mohammadranjbarz 
Very good points

Issue #261, opened 2021-12-21
Disable Apollo playground in production


mohammadranjbarz:

mohammadranjbarz:
@MoeNick 
The backend is up in staging but if you go https://serve.giveth.io/graphql you can't see apollo playground

divine-comedian:
navigating to `https://mainnet.serve.giveth.io/graphql`

this is currently what appears 
![image](https://user-images.githubusercontent.com/67759413/152166571-851c32c7-d79b-4c00-9175-65f254798996.png)


would we consider this state effectively disabled @mohammadranjbarz ?

mohammadranjbarz:
> navigating to `https://mainnet.serve.giveth.io/graphql`
> 
> this is currently what appears ![image](https://user-images.githubusercontent.com/67759413/152166571-851c32c7-d79b-4c00-9175-65f254798996.png)
> 
> would we consider this state effectively disabled @mohammadranjbarz ?

Yeah exactly, it mean playground had been disabled successfully

PR #262, opened 2021-12-21
Can disable Apollo playground by changing value in config

mohammadranjbarz:
related to #261

If we want to disable the playground, we just need to add
`DISABLE_APOLLO_PLAYGROUND=true` in config


CarlosQ96 (approved):
Tested and works fine! I also thought this was the right idea to fix the issue. 
LGTM!!!

PR #263, opened 2021-12-21
Feature: Set projects as pending review when edited

CarlosQ96:
When the project is edited the value of listed is set to Null, meaning it's back to review.
From this PR on we will use UpdatedAt to set the listing of the project, because we want to list it again based on the last time it was updated not on creation or it will be listed the next day instantly.

Issue: #264

### DIFF
--- a/entities/project.ts
+++ b/entities/project.ts
@@ -195,8 +195,8 @@ class Project extends BaseEntity {
   totalProjectUpdates: number;
 
   @Field(type => Boolean, { nullable: true })
-  @Column({ default: null, nullable: true })
-  listed: boolean;
+  @Column({ type: 'boolean', default: null, nullable: true })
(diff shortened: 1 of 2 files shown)


CarlosQ96:
@mohammadranjbarz I validated this changed on the @column decorator doesn't delete data (everything seem fine), but  please double check. Always gotta be careful when touching the columns.

CarlosQ96:
@MoeNick Just to confirm, when the owner of the project edits it's own project, we set it back listed = null and it's pending on review based on that. Correct?
If so, the PR is correct.

MoeNick:
Related #263 
Thanks, Seems to be as expected @CarlosQ96

mohammadranjbarz (approved):
Thanks @CarlosQ96 
I have tested it, and it was ok

PR #268, opened 2021-12-23
Add migration file to fill donation.userId

mohammadranjbarz:
Please check it to see if is ok or not


CarlosQ96 (approved):
Just checked users are saving their addresses in lowercase and also donations save it in lowercase. So LGTM!

Issue #273, opened 2021-12-25
Having problem in verifying Gitcoin donation on mainnet


mohammadranjbarz:
Donation sample

```
{
    "id": 3384,
    "transactionId": "0x860bd7499393a02e0e732bffec4b2cc52ac04220989f053770aa05c63dbf9725",
    "toWalletAddress": "0x236daa98f115caa9991a3894ae387cdc13eaad1b",
    "fromWalletAddress": "0xed8db37778804a913670d9367aaf4f043aad938b",
    "currency": "GTC",
    "anonymous": false,
    "amount": 2,
    "valueUsd": 24.757584,
    "createdAt": "2021-12-08 20:07:00.616",
    "projectId": 251,
    "userId": 10,
    "transactionNetworkId": 1,
    "valueEth": 0.00564214,
    "priceEth": 0.00282107,
    "priceUsd": 12.378792,
    "donationType": "NULL",
    "nonce": "NULL",
    "status": "pending",
    "verifyErrorMessage": "NULL",
    "speedup": "False",
    "isFiat": "False",
    "transakStatus": "NULL",
    "transakTransactionLink": "NULL",
    "tokenAddress": "NULL"
  }
```

Maybe we should add test case for this in our automated tests to see what's the reason

PR #277, opened 2021-12-26
F 273 verify gitcoin and giv donations

mohammadranjbarz:
related to #273


mohammadranjbarz:
@mateodaza 
I just set you as reviewer to sync you,
when we want to verify donations we check token addresses in tokenUtils file, so when adding a new token in monoswap, we should add it here, to net getting error when verifying fonations

mateodaza (approved):

Issue #77, opened 2021-06-08
Fix migration files


mateodaza:
this happens trying to run migration fails using npm run typeorm:cli -- migration:run

cquinterom096:
Currently the project seems to be setup to sync with the entities completely. This leads to migrations not being run and some being redundant. I am somewhat unsure how to proceed here.

Would like to ask for some insight on how the project is managing database changes.

divine-comedian:
deep in the matrix today with @cquinterom096  I made some observations on the migration files - 
I was able to successfully run all migration files by making a couple modifications. 

I commented out the first query send in each file  - EXAMPLE: 
in `migration/1612547269219-UpdateUserEmailUnique.ts`
```
import {MigrationInterface, QueryRunner, TableUnique} from "typeorm";

export class UpdateUserEmailUnique1612547269219 implements MigrationInterface {
    name = 'UpdateUserEmailUnique1612547269219'

    public async up(queryRunner: QueryRunner): Promise<void> {
        // await queryRunner.query(`COMMENT ON COLUMN "user"."email" IS NULL`);
        // //await queryRunner.query(`ALTER TABLE "user" DROP CONSTRAINT "UQ_e12875dfb3b1d92d7d7c5377e22"`);
        // await queryRunner.dropUniqueConstraint('user', new TableUnique({ columnNames: ['email'] }))
        // // staging await queryRunner.dropUniqueConstraint('user', 'unique_email_address')
    }

    public async down(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.createUniqueConstraint('user', new TableUnique({ columnNames: ['email'] }))
        await queryRunner.query(`COMMENT ON COLUMN "user"."email" IS NULL`);
    }

}
``` 
I did this for all migrations. Specifically however in seedProjectStatus it gets a bit spicy 

```
import { MigrationInterface, QueryRunner, getRepository } from 'typeorm'
import { ProjectStatus } from '../entities/projectStatus'

export class SeedProjectStatus1614082100757 implements MigrationInterface {
  public async up (queryRunner: QueryRunner): Promise<void> {
    // const projectStatusRepository = getRepository(ProjectStatus)
    // const projectStatuses = projectStatusRepository.create([{
    //     symbol: 'rjt',
    //     name: `rejected`,
    //     description: 'This project has been rejected by Giveth or platform owner'
    //   },
    //   {
    //     symbol: 'pen',
    //     name: 'pending',
[...]

PR #279, opened 2021-12-29
F 77 fix migrations for test

mohammadranjbarz:
related to #77


mohammadranjbarz:
@CarlosQ96 
I have deleted useless migrations( Because they made needed changes, and we don't need the anymore), for sratup we just need two migrations , and from now on our migrations should be executed during CI/CD, so we can't write a migration file that users with clean DB have problem with running that

CarlosQ96 (approved):
I tested locally migrating my db with fresh dbs with some data, and it works fine.
However, we should proceed to test on the dev server, just to see there aren't conflicts running migrations. After that let's merge. 
(btw, I think the dev server is not deploying correctly tho, gets stuck).

mohammadranjbarz:
> I tested locally migrating my db with fresh dbs with some data, and it works fine. However, we should proceed to test on the dev server, just to see there aren't conflicts running migrations. After that let's merge. (btw, I think the dev server is not deploying correctly tho, gets stuck).

Thanks Carlos, you are right, Develop has problem

@geleeroyale Can you create a db connection for develop environment, and import DB data of develop env to that?
dev server: dev.serve.giveth.io

geleeroyale:
Dev.serve has been migrated successfully to a hosted database.

mohammadranjbarz:
@CarlosQ96 

I have merged this branch to develop and it went right, I merge this just now

PR #282, opened 2022-01-03
Add logger

mohammadranjbarz:
related to #302


src/middleware/logger.ts
```diff
@@ -1,9 +1,9 @@
...
-};
+// import { MiddlewareFn } from 'type-graphql';
+//
+// import { MyContext } from '../../types/MyContext';
+//
+// export const logger: MiddlewareFn<MyContext> = async ({ args }, next) => {
+//   logger.debug('args: ', args);
+//
+//   return next();
+// };
```
CarlosQ96: If this file is not gonna be used, let's delete it.

src/delete-uniswap/sdk.ts
```diff
@@ -52,8 +53,8 @@ export default class Sdk {
     const route = new Route([pair], token);
     const price = route.midPrice.toSignificant(6);
 
-    // console.log('inv', route.midPrice.invert().toSignificant(6)) // 0.00496756
-    // console.log('price', price) // 201.306
+    // logger.debug('inv', route.midPrice.invert().toSignificant(6)) // 0.00496756
+    // logger.debug('price', price) // 201.306
```
CarlosQ96, line 58: I assume this comments serve as documentation? but they will break if we try to use them. I'll add the dependency.

src/config.ts
```diff
@@ -91,7 +92,7 @@ class Config {
     envVars.forEach(envVar => {
       if (envFile[envVar]) {
         this[envVar] = envFile[envVar];
-        // console.log(`envVar ---> : ${this[envVar]}`)
+        // logger.debug(`envVar ---> : ${this[envVar]}`)
```
CarlosQ96, line 96: I'll add the logger dependency.

src/services/syncGivPrices.ts
```diff
@@ -8,7 +8,7 @@ const cronJobTime =
   process.env.REVIEW_OLD_GIV_PRICES_CRONJOB_EXPRESSION || '0 0 * * *';
 
 export const runUpdateHistoricGivPrices = () => {
-  console.log('runUpdateHistoricGivPrices() has been called');
+  logger.debug('runUpdateHistoricGivPrices() has been called');
```
CarlosQ96, line 12: This line broke we missed the import. I'll add it.

src/server/components/FilterListedComponent.jsx
```diff
@@ -5,7 +5,7 @@ import { FormGroup, filterStyles, Label } from '@admin-bro/design-system'
 
 const FilterListedComponent = props => {
   const { onChange, property, theme } = props;
-  console.log(props);
+  logger.debug(props);
```
CarlosQ96: Removing this, I added it for debugging purposes but it's not required. This is a adminBro custom component in jsx.

CarlosQ96 (changes requested):

CarlosQ96 (approved):
Checked every single file didn't miss the import.
I added the missing dependencies on some files so they don't break and removed the old logger. Ran the server and a few processes and they seem to work fine.
Great that we set a rotation for the logger I think it's fine. Good job!
We can merge tomorrow.
LGTM!!

src/delete-uniswap/sdk.ts
mohammadranjbarz: Thanks man

mohammadranjbarz:
> Checked every single file didn't miss the import. I added the missing dependencies on some files so they don't break and removed the old logger. Ran the server and a few processes and they seem to work fine. Great that we set a rotation for the logger I think it's fine. Good job! We can merge tomorrow. LGTM!!

Thanks for your review and fixing errors

PR #286, opened 2022-01-05
Hotfix: Add cronjob for updating old prices of GIV tokens

CarlosQ96:
Uses Cronjobs to run a update. I believe this is a one time only thing. 
To test, create a donation on the ranges 24-28 of december which was the time monoswap wasn't updated


mohammadranjbarz (approved):
Thanks @CarlosQ96 
It seems ok, I ran it on my machine connecting to our production DB
And after that I changed the query to always check GIV donations that doesn't have price

PR #298, opened 2022-01-11
Find donations conflicts for givbacks

mohammadranjbarz:
@CarlosQ96 
I have deleted src/delete-uniswap/tokenLists.ts  list to just have one list for tokens list
( the findConflict script doesn't have any side effect on our application, you just should check the token list is ok)


src/scripts/findConflicts.ts
```diff
@@ -0,0 +1,303 @@
...
+
+const EXCLUDED_FROM_ADDRESSES = ['0x0000000000000000000000000000000000000000'];
+
+const smartContractAddresses: string[] = [];
+
+// 2021/12/24 16:00:00 GMT, beginning of round
+const beginTimestamp = 1640361600;
+
+// 2021/01/07 16:00:00 GMT, end of round
+const endTimestamp = 1641571200;
```
CarlosQ96, line 34: Is the script reusable? Shouldn't we move this variables to env vars, so we only have to change them and rerun the script when required?

CarlosQ96: Maybe for future rounds.

CarlosQ96 (approved):
Aside from that I think the token list is fine and I am sure you tested with Griff and the team the script after looking at the chats.

src/scripts/findConflicts.ts
mohammadranjbarz: You know when changing it for future round it's not needed to commit anything, I think we can merge it now, in the future we can change it
Thanks for your suggestion

Issue #285, opened 2022-01-05
Sort by raised amount with trace donations counted. 


MoeNick:
We need to implement a query to count trace donation plus io donations we may need it for TP version.

mohammadranjbarz:
@mateodaza 

You can use this query in staging 

```

query {
  projects(

  orderBy: {
      field: TraceDonations,
      direction: DESC
  }
) {
    totalCount
    projects {
      traceCampaignId
      id
      title
      slug
      qualityScore
      walletAddress
      creationDate
      adminUser{
        name
      }
      users {
        name
      }
      updatedAt
      totalReactions
      totalDonations
      totalTraceDonations
    }
    categories {
      name
    }
  }
}

```

MoeNick:
@divine-comedian can you find a band with and help us on this with service testing:
1- total amount donated for a non-traceable project should be correct and the same with DB
2- total amount donated for a traceable project should be correct: all donations with giveth.io + its campaign donation amount in trace. 
All of them in staging.

divine-comedian:
testing out donations on trace vs. donations on IO 
develop.giveth.io
next.giveth.io

right now I can see zero consistency between the two dapps. I tested making a few donations on both via Ropsten
https://develop.giveth.io/campaign/testing-trace-endpoint
https://next.giveth.io/project/testing-trace-endpoint
![image](https://user-images.githubusercontent.com/67759413/152071470-5ee5ce16-85e2-4374-ac01-aede38bda6b3.png)

I made two separate transactions each for 0.003 ETH to each project (trace and IO) so it should show both donations but only shows the one I made on the platform (sorry I used the same amount so it makes it more confusing)
![image](https://user-images.githubusercontent.com/67759413/152072134-e980e312-7c3b-4b6b-82f6-c39b4e28814c.png)
https://develop.giveth.io/campaign/ram-4th-project
https://next.giveth.io/project/ram-4th-project

mohammadranjbarz:
@divine-comedian @MoeNick Sorry I had missed the last comment of Mitch, I have created a PR for fixing query in featehrs after merging that it would fix this issue

mohammadranjbarz:
@divine-comedian And the frontend side is not implemented yet, so we need to test it with calling webservice directly

divine-comedian:
@mohammadranjbarz  okay - how can I test this issue? should we wait for the frontend implementation?

mohammadranjbarz:
> testing-trace-endpoint

[...]

PR #305, opened 2022-01-20
F 285 count trace donations

mohammadranjbarz:
related to #285


src/resolvers/donationResolver.ts
```diff
@@ -94,30 +96,43 @@ export class DonationResolver {
...
+        .where(`donation.projectId = ${projectId}`);
+
+      const [donations, donationsCount] = await query
+        .take(take)
+        .skip(skip)
+        .getManyAndCount();
+      return {
+        donations,
+        totalCount: donationsCount,
+        totalUsdBalance: project.totalDonations,
```
CarlosQ96, line 133: Mateo told me to add the totalEth balance for the frontend / new design that's why I had it. We should add it to the database and update our usdvalue updater script to include totalEth value. 
But I am not sure how this will interact with the trace, does trace have ethvalue too?

CarlosQ96 (changes requested):
Would be my only concern regarding new design / current frontend.

src/resolvers/donationResolver.ts
mohammadranjbarz: @mateodaza  @CarlosQ96 
We dont have `ethValue` for trace donations, is it ok to send it just for givethio donations?

CarlosQ96: As per the meeting we can ignore the totalEthValue and merge.

CarlosQ96 (approved):

PR #344, opened 2022-01-29
F 285 add sort by trace total donation

mohammadranjbarz:
related #285  #309 
@CarlosQ96 
Now we can get sorted list by `traceTotalDonations`, the num field is `TraceDonations`

BTW I have added test cases for get projects query

### DIFF
--- a/test/testUtils.ts
+++ b/test/testUtils.ts
@@ -152,7 +152,16 @@ export const SEED_DATA = {
     id: 2,
     admin: '2',
   },
-  CATEGORIES: ['food1', 'food2', 'food3', 'food4', 'food5', 'food6', 'food7'],
+  CATEGORIES: [
+    'food1',
+    'food2',
+    'food3',
+    'food4',
+    'food5',
+    'food6',
+    'food7',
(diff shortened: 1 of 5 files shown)


CarlosQ96 (approved):
Ran all tests and they worked fine! I shall fix the remaining tests in another PR. Retaking on prisma now that I am full time. So many changes so I'll adapt what I did with all this.

test/testUtils.ts
```diff
@@ -51,6 +54,74 @@ export const generateTestAccessToken = async (id: number): Promise<string> => {
   );
 };
 
+export interface CreateProjectData {
```
CarlosQ96, line 57: I was thinking we have a seed file called helpers.ts Maybe we can reuse that? or adapt it for test env?

CarlosQ96: Not required for this PR but there are a bunch of models there

mohammadranjbarz: great idea, I had forgotten about that, I think in the future we can use that, thanks for mentioning

Issue #341, opened 2022-01-26
After the Cancelling of project  (By admin): change verification status to false & listing status to false. 


MoeNick:

MoeNick:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/impact-graph/341?planning-poker) with ZenHub @mohammadranjbarz

MoeNick:
shall I move it to UAT?

PR #345, opened 2022-01-30
F 341 delist and unverified projects after cancellation by admin

mohammadranjbarz:
related to #341 

This should be merge after #344


CarlosQ96 (approved):
Tested admin bro manually and ran tests. LGTM! 

Learning a lot about the way you write test!

Issue #342, opened 2022-01-27
Backend: Rainbowroll Donation distribution scripts


CarlosQ96:
After discussing with Mohammad this issue will be done with two scripts:
1. Update our Giv Script to also catch mainnet donations and create in our db the donation records.
2. Add a SegmentNotified column to donations to keep track which have been notified.
3. Add a script that notifies those donations that haven't been notified to customers.

MoeNick:
@mohammadranjbarz would u pls put your comments here? as we talked earlier on this.

mohammadranjbarz:
@MoeNick 
As I talked with @divine-comedian 
We have a use case, sometimes happens that we move some funds(ETH/XDAI/or other tokens) to multiple addresses(recipient address of projects)  by **Disperse** app, and we want to create donations for those transfers, 
My suggestion is add a page in admin bro, you enter the txHash of disperse transaction, something like this: https://blockscout.com/xdai/mainnet/tx/0xd10f17ce5671f6d1084a902f0d7df3edc5c877dd83705ef55b7dacbab72922db
and get price of donated token at time of donation from admin, then I get the transaction data from network, and create donation per every transfer of that transaction and besides, I can give this option to admin to enter a txHash of a normal transfer to create one donation for that transaction,
please correct me if I'm mistaken or my suggestion needs some changes

divine-comedian:
we'll use CSV airdrop since it is built right into the gnosis-safe apps

![image](https://user-images.githubusercontent.com/67759413/152146007-67a60f7c-f6fd-4de3-88cd-4832171486b4.png)

mohammadranjbarz:
@CarlosQ96  I moved it to in-progress because my side is not finishedd yet

mohammadranjbarz:
<img width="1006" alt="Screen Shot 1400-11-24 at 12 19 40" src="https://user-images.githubusercontent.com/9850545/153745969-ed39783b-f8c2-4f1f-89e3-d939362bbe29.png">
@MoeNick  @divine-comedian 

I have implemented a new page in donation page, so after you donated to projects by csv airdrop, you need to have `txHash`, price USD of that token at that time, then I will get token transfers of that tx and add them to DB, 

Now I support **erc20** token transfers, because we didn't have any transaction of csv airdrop for **native** transfers ( And I and @aminlatifi  guessed it would have a little complexity),
[...]

Issue #332, opened 2022-01-26
Write test cases for donationsByDonor() in donationResolver.test.ts


mohammadranjbarz:

CarlosQ96:
This card was actually helpful to setup the testing of donations for other issues: #342

mohammadranjbarz:
@MoeNick As it's just technical I think we can skip UAT and just move it to DONE

CarlosQ96:
I'll move it to Done. This was mostly technical and a great start to setup future tests related to donations.

PR #348, opened 2022-02-02
Add donations by donor query tests

CarlosQ96:
Issue: https://github.com/Giveth/impact-graph/issues/332

I did this issue first as it will be helpful for the issue #342 I am also working on for setting the donations for testing.


src/resolvers/projectResolver.test.ts
```diff
@@ -1050,7 +1047,3 @@ function editProjectTestCases() {
     );
   });
 }
-
-before(async () => {
-  apolloServer = await createServerWithDummyUser();
```
mohammadranjbarz, line 1055: JFYI
I noticed that createServerWithDummyUser() is useless in our tests so I deleted them.

CarlosQ96: Ohhhhh. I really thought it was for axios to work. Got it!

mohammadranjbarz (approved):
Thanks @CarlosQ96 
It was a good jump in automatic tests

PR #349, opened 2022-02-03
Add cronjob to call segment of unnotified donations

CarlosQ96:
Related to issue: #342 

Changed the config of segment in test mode to not make requests to the api
image
and 
Added Sinon mock library to ignore sleeps that might slow down test suite execution.

Tested both manually and with the automatic test.


src/entities/donation.ts
```diff
@@ -126,6 +126,10 @@ export class Donation extends BaseEntity {
   @Column({ nullable: true })
   transakTransactionLink?: string;
 
+  @Field(type => Boolean, { nullable: true })
+  @Column({ nullable: true, default: false })
+  segmentNotified: boolean;
+
```
CarlosQ96, line 132: @mohammadranjbarz what do you think? I set the default as false. But we must avoid notifying everyone once we deploy, so We would need to manually set all to true for old donations.
Then afterwards new donations catched by your script will go through my cronjob.

mohammadranjbarz: I think ut's good to set it 'false` for default value, but please write a migration file to set `true` for older donations otherwise your cronjob will send notification for all older donations and it's not good at all

src/services/cronJobs/notifyDonationsWithSegment.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { notifyMissingDonationsWithSegment } from './notifyDonationsWithSegment';
+import * as utils from '../../utils/utils';
+
+describe(
+  'notifyMissingDonationsWithSegment() test cases',
+  notifyMissingDonationsWithSegmentTestCases,
+);
+
+function notifyMissingDonationsWithSegmentTestCases() {
+  it('should call segment per donation not notified', async () => {
```
mohammadranjbarz, line 17: Bueno, great test case

migration/1643891610952-setSegmentNotifiedForOlderDonations.ts
```diff
@@ -0,0 +1,32 @@
...
+        SELECT FROM pg_tables
+        WHERE  schemaname = 'public'
+        AND    tablename  = 'donation'
+      );
+    `);
+    if (donationTableExists[0].exists) {
+      await queryRunner.query(`UPDATE donation SET "segmentNotified" = false`);
+    }
+  }
+}
```
mohammadranjbarz, line 13: I think you can make migration file much easier with this:

```

import { MigrationInterface, QueryRunner } from 'typeorm';

export class setSegmentNotifiedForOlderDonations1643891610952 implements MigrationInterface {
  async up(queryRunner: QueryRunner): Promise<void> {
    await queryRunner.query(
      `ALTER TABLE IF EXISTS  "donation" ADD  "segmentNotified" boolean DEFAULT true`,
    );
  }

  async down(queryRunner: QueryRunner): Promise<void> {
    await queryRunner.query(
      `ALTER TABLE IF EXISTS  "donation" DROP "segmentNotified"`,
    );
  }
}
```

Other parts of the PR seems good to me

mohammadranjbarz (changes requested):

migration/1643891610952-setSegmentNotifiedForOlderDonations.ts
CarlosQ96: The thing is, I don't want to change the default value constraint to **true**, or else my crontask will never run.
Nor drop the column, as the sync mode of typeorm is automatically handling it the creation/dropping of columns.

I only want to set the **value** of the current donations segmentNotified on the db to true.

CarlosQ96: Let me see if I an try something else

mohammadranjbarz: Ok, so if you delegate dropping that to typeorm, you can leave the down() function empty, you can change your migration file like this

```
import { MigrationInterface, QueryRunner } from 'typeorm';

export class setSegmentNotifiedForOlderDonations1643891610952 implements MigrationInterface {
  async up(queryRunner: QueryRunner): Promise<void> {
    await queryRunner.query(
      `ALTER TABLE IF EXISTS  "donation" ADD  "segmentNotified" boolean DEFAULT true`,
    );
  }

  async down(queryRunner: QueryRunner): Promise<void> {

  }
}
```
Isn't it enough?

CarlosQ96: Actually you are right, I got confused, we run migrations before the app syncs the database. So if we set it to true in the migration it will work. 
Then the sync shall set it default to false for future donations. 
Also yeah, we can delegate to typeorm the down function.

mohammadranjbarz (approved):

mohammadranjbarz:
@CarlosQ96  Great work, thanks

PR #350, opened 2022-02-04
Make segmentNotified when creating donation by saveDonation() mutation

mohammadranjbarz:
I created a branch from your PR and made some changes

### DIFF
--- a/src/resolvers/donationResolver.ts
+++ b/src/resolvers/donationResolver.ts
@@ -279,6 +279,7 @@ export class DonationResolver {
         transactionNetworkId: Number(transactionNetworkId),
         currency: token,
         projectWalletAddress: project.walletAddress,
+        segmentNotified: true,
         createdAt: new Date(),
       };
 
--- a/src/services/cronJobs/notifyDonationsWithSegment.ts
+++ b/src/services/cronJobs/notifyDonationsWithSegment.ts
@@ -35,6 +35,9 @@ export const notifyMissingDonationsWithSegment = async () => {
     );
     const project = await Project.findOne({ id: donation.projectId });
     const owner = await User.findOne({ id: Number(project?.admin) });
+    // Notify owner donation was received ( if user and owner and project exist)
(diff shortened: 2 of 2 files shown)


src/resolvers/donationResolver.ts
```diff
@@ -279,6 +279,7 @@ export class DonationResolver {
         transactionNetworkId: Number(transactionNetworkId),
         currency: token,
         projectWalletAddress: project.walletAddress,
+        segmentNotified: true,
```
mohammadranjbarz, line 282: @CarlosQ96 
When creating donations we should set this to true, otherwise if user create donation, we will send him an email but because we didn't set this field to `true`, the cronjob will send another one

src/services/cronJobs/notifyDonationsWithSegment.ts
```diff
@@ -35,6 +35,9 @@ export const notifyMissingDonationsWithSegment = async () => {
     );
     const project = await Project.findOne({ id: donation.projectId });
     const owner = await User.findOne({ id: Number(project?.admin) });
+    // Notify owner donation was received ( if user and owner and project exist)
+    donation.segmentNotified = true;
```
mohammadranjbarz, line 39: @CarlosQ96 
I think we move this update out of `if` scope, because if either project or user doesn't exist, it's not our problem, we set `segmentNotified‍‍‍‍` anyway, to don't check it again

src/resolvers/donationResolver.ts
CarlosQ96: Agreed, thanks !

src/services/cronJobs/notifyDonationsWithSegment.ts
CarlosQ96: That's right, we tried our best to notify but if its not posible its not posible. Best effort service.

CarlosQ96 (approved):

PR #352, opened 2022-02-05
Public Profile: Add donationByUserId query and LikedProjects Query

CarlosQ96:
Related: https://github.com/Giveth/giveth-io-typescript/issues/41
I made a new query for this because we can re use it in two views that do exactly the same.
User profile and public profile of users like in the design.

This PR also contains query for Liked Projects of a user.


CarlosQ96:
To test the liked projects query. You can create a reaction record on the default ProjectUpdate of a project. Then run the following query:
```
query {
  likedProjectsByUserId(take: 1, skip: 0, userId: 1) {
    projects {
      id
      title
    }
    totalCount
  }
}
```

CarlosQ96:
To tests the donations pagination and filtering. Run the following Query: 

The field enum has: TokenAmount, UsdAmount, CreationDate as options.
```
query {
  donationsByUserId(
    take: 10
    skip: 0
    orderBy: { field: TokenAmount, direction: ASC }
    userId: 1
  ) {
    donations {
      id
      transactionId
      transactionNetworkId
      toWalletAddress
      fromWalletAddress
      currency
      anonymous
      valueUsd
      amount
      user {
        id
      }
      project {
        id
      }
      createdAt
    }
    totalCount
  }
}

```

mohammadranjbarz (approved):
@CarlosQ96  Great work, I'm really happy that we have you full time.

PR #353, opened 2022-02-07
F 84 provide reasons to deactivating project

mohammadranjbarz:
related to Giveth/giveth-io-typescript#84

What Id did on this PR:
1. Create a table named `project_status_reason` for adding template reasons there https://github.com/Giveth/giveth-io-typescript/issues/84#issuecomment-1016941722
2. Add this entity to adminBro so admins can add new reasons
3. Add a new table for adding history changes there 
4. Add new record when someone activate/deactivate from our dapp ( ideally in future we should add record even if admin change status from adminBro)
5. Add projectStatusHistory to admin bro, so admins can see the changes and history of a project there
6. Refactor deactivate/activate mutations
7. Writing some test cases

PS: @CarlosQ96  I wrote down above items to make the review easy for you

### DIFF
--- a/migration/1643957749780-addProjectStatusReasonTable.ts
+++ b/migration/1643957749780-addProjectStatusReasonTable.ts
@@ -1,0 +1,28 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class addProjectStatusReasonTable1643957749780
+  implements MigrationInterface
+{
+  async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE IF NOT EXISTS public.project_status_reason
+              (
+                  id serial NOT NULL,
+                  description character varying,
+                  "statusId" integer,
+                  CONSTRAINT "PK_453e771767c65a08619aff4ba53" PRIMARY KEY (id),
+                  CONSTRAINT "FK_2edf0fca3c1aca99d1c4924e6a2" FOREIGN KEY ("statusId")
+                      REFERENCES public.project_status (id) MATCH SIMPLE
+              )
+`,
+    );
+  }
+
+  async down(queryRunner: QueryRunner): Promise<void> {
+    try {
+      await queryRunner.query(`DROP TABLE "project_status_reason"`);
+    } catch (e) {
+      console.log('addProjectStatusReasonTable1643957749780 error', e);
+    }
+  }
+}
--- a/migration/1643957749780-addProjectStatusReasonTable.ts
+++ b/migration/1643957749780-addProjectStatusReasonTable.ts
@@ -18,11 +18,5 @@ export class addProjectStatusReasonTable1643957749780
     );
   }
 
-  async down(queryRunner: QueryRunner): Promise<void> {
-    try {
-      await queryRunner.query(`DROP TABLE "project_status_reason"`);
-    } catch (e) {
-      console.log('addProjectStatusReasonTable1643957749780 error', e);
-    }
-  }
+  async down(queryRunner: QueryRunner): Promise<void> {}
 }
--- a/migration/1643962364050-seedProjectStatusreasons.ts
+++ b/migration/1643962364050-seedProjectStatusreasons.ts
@@ -16,6 +16,5 @@ export class seedProjectStatusreasons1643962364050
   }
 
   async down(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.query(`DELETE FROM project_status_reason`);
   }
 }
--- a/migration/1643962364050-seedProjectStatusreasons.ts
+++ b/migration/1643962364050-seedProjectStatusreasons.ts
@@ -5,6 +5,7 @@ export class seedProjectStatusreasons1643962364050
   implements MigrationInterface
 {
   async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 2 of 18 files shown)


migration/1643957749780-addProjectStatusReasonTable.ts
```diff
@@ -0,0 +1,28 @@
...
+                  CONSTRAINT "FK_2edf0fca3c1aca99d1c4924e6a2" FOREIGN KEY ("statusId")
+                      REFERENCES public.project_status (id) MATCH SIMPLE
+              )
+`,
+    );
+  }
+
+  async down(queryRunner: QueryRunner): Promise<void> {
+    try {
+      await queryRunner.query(`DROP TABLE "project_status_reason"`);
```
CarlosQ96: Remember if you rollback and run the server, synchronization will create the entity schema again. 

Second scenario would be if the person deletes the entity, run server and then rollbacks. It will break the migrations as sync mode already deleted the table.

I think we can delegate this to the sync mode and leave a comment.
What do you think?
Maybe we should document how to rollback correctly. But you can decide if it's fine for now or not.

test/testUtils.ts
```diff
@@ -208,6 +210,24 @@ export const SEED_DATA = {
     'food7',
     'food8',
   ],
+  STATUS_REASONS: [
+    {
+      description: 'The project has completed its goals!',
+      statusId: ProjStatus.deactive,
+    },
+    {
+      description: 'The project is not longer active.',
```
CarlosQ96: Just a typo. It should should be "no" instead of "not".
"The project is no longer active."

migration/1643962364050-seedProjectStatusreasons.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class seedProjectStatusreasons1643962364050
+  implements MigrationInterface
+{
+  async up(queryRunner: QueryRunner): Promise<void> {
+    // https://github.com/Giveth/giveth-io-typescript/issues/84#issuecomment-1016941722
+    await queryRunner.query(`INSERT INTO public.project_status_reason ("statusId", description ) VALUES 
+                    (${ProjStatus.deactive},'The project has completed its goals!'),
+                    (${ProjStatus.deactive},'The project is no longer in need of funding.'),
+                    (${ProjStatus.deactive},'The project is not longer active.'),
```
CarlosQ96: Just a small typo. It should be "no" instead of "not".
"The project is no longer active."

src/resolvers/projectResolver.ts
```diff
@@ -1141,7 +1144,26 @@ export class ProjectResolver {
...
+      };
+      analytics.track(
+        SegmentEvents.PROJECT_ACTIVATED,
+        `givethId-${ctx.req.user.userId}`,
+        segmentProject,
+        null,
+      );
+      return true;
     } catch (error) {
       SentryLogger.captureException(error);
```
CarlosQ96, line 1169: we should add the error in our logs like you did in the deactivate mutation, to keep consistency.
      logger.error('projectResolver.activateProject() error', error);

src/entities/project.ts
```diff
@@ -341,13 +351,38 @@ class Project extends BaseEntity {
...
   // Status 7 is deleted status
   mayUpdateStatus(user: User) {
-    if (this.statusId === ProjStatus.cancel) return false;
+    if (this.statusId === ProjStatus.cancel) {
+      throw new Error(
+        errorMessages.THIS_PROJECT_IS_CANCELLED_OR_DEACTIVATED_ALREADY,
+      );
+    }
 
     if (this.users.filter(o => o.id === user.id).length > 0) {
```
CarlosQ96: I have a faint memory this works fine now, but on really  REALLY REALLLY Oold projects won't. It seems they don't have the user relation set up correctly or non existant, but they have admin string. This happened months ago.

I think we should use the filter as is but also add an OR  for checking the admin string which is also important.
```
if (this.users.filter(o => o.id === user.id).length > 0 || user.id === Number(this.admin))
```

CarlosQ96 (approved):
I ran the tests, queries and checked all the code. It's seems working good! Great job Mohammad!
I left some comment for some minor things and one edge case we encountered like easily 6 months ago.

src/entities/project.ts
CarlosQ96: Or just the admin string... which honestly has the source of truth, without filtering a list of users.

migration/1643962364050-seedProjectStatusreasons.ts
mohammadranjbarz: Oh thanks it has been fixed

src/resolvers/projectResolver.ts
mohammadranjbarz: You are exactly write, I added the logger

src/entities/project.ts
mohammadranjbarz: Nice catch, I have changed the condition

test/testUtils.ts
mohammadranjbarz: Thanks, changed

migration/1643957749780-addProjectStatusReasonTable.ts
mohammadranjbarz: You are right, I deleted them

Issue #343, opened 2022-01-27
Writing Auto Test for changing Statuses Scenarios


MoeNick:
- [x] If Listing=Null stayed more than N days, automatically set it to "Listed"
- [x] Edit Projects: won't affect Verification status - changes listing to "not reviewed" means listing=null
- [x] Deactivating (Archiving): won't affect Verification status - won't affect listing status. 
- [x] Activating (Unarchiving): won't affect Verification status - changes listing to "not reviewed"
- [x] Cancelling (By admin): make it unverified & delisted. 
- [x] Verification approved: won't affect the listing status
- [x] Verification removed: won't affect listing status. 
- [x] Delisting: won't affect the verification status,
- [x] Listing: won't affect the verification status,

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/impact-graph/343?planning-poker) with ZenHub @mohammadranjbarz @CarlosQ96

CarlosQ96:
I'll put 8, as I have to get familiar with certain testing tools first.

mohammadranjbarz:
@WhyldWanderer  @MoeNick  I saw the code and I noticed the user can not active a deactivated/cancelled project, so this endpoint is useless, and I can't write test for this
**Activating (Unarchiving): won't affect Verification status - changes listing to "not reviewed"**
If it's a bug please tell me to fix that otherwise we can remove that from checklist

mohammadranjbarz:
> @WhyldWanderer @MoeNick I saw the code and I noticed the user can not active a deactivated/cancelled project, so this endpoint is useless, and I can't write test for this **Activating (Unarchiving): won't affect Verification status - changes listing to "not reviewed"** If it's a bug please tell me to fix that otherwise we can remove that from checklist

Sorry it was my mistake, the description of statuses was misleading, so I misunderstood,

just another question if **User** activate the deactivated project we should change `listed` to null, what about when Admin activates a project?
@MoeNick

mohammadranjbarz:
@MoeNick 
I wrote a migration file to change description/name of statuses in this PR, and I wrote a table for that, you can see that here
https://github.com/Giveth/impact-graph/tree/f_343_write_tests_for_changing_statuses_scenarios#statuses

CarlosQ96:
PR approved! All tests correlate the list.

MoeNick:
[...]

PR #354, opened 2022-02-08
F 343 write tests for changing statuses scenarios

mohammadranjbarz:
related to #343

### DIFF
--- a/src/resolvers/projectResolver.test.ts
+++ b/src/resolvers/projectResolver.test.ts
@@ -1030,4 +1030,104 @@ function editProjectTestCases() {
       createProjectResult.data.data.addProject.walletAddress,
     );
   });
+  it('Should update successfully and verified(true) field would not change', async () => {
+    const accessToken = await generateTestAccessToken(SEED_DATA.FIRST_USER.id);
+    const project = await saveProjectDirectlyToDb({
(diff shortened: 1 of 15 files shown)


src/resolvers/projectResolver.test.ts
```diff
@@ -1065,6 +1049,106 @@ function editProjectTestCases() {
...
+      },
+      {
+        headers: {
+          Authorization: `Bearer ${accessToken}`,
+        },
+      },
+    );
+    assert.isTrue(editProjectResult.data.data.editProject.verified);
+  });
+  it('Should update successfully and verified(true) field would not change', async () => {
```
CarlosQ96: Just a typo, the test indicated verified(true) in the description but its actually false if you look at the test.

CarlosQ96 (approved):
All tests passed. Looking at all code and where ProjStatus is being used after renaming and everything is in order. 

Checked the list one by one and the tests correlate the list in issue #343

Just saw a typo on one of the test descriptions.
Everything else LGTM!

src/resolvers/projectResolver.test.ts
mohammadranjbarz: Oh you are right, It has been fixed
Thank you for being so punctilious

PR #356, opened 2022-02-09
Feat/project reactions

aminlatifi:


src/resolvers/projectResolver.ts
```diff
@@ -256,63 +320,95 @@ export class ProjectResolver {
...
-        .where('project.statusId = 5 AND project.listed = true')
-        .orderBy(`project.${field}`, direction)
-        .limit(skip)
-        .take(take)
-        .innerJoinAndSelect('project.categories', 'c')
-        .getManyAndCount();
+    let query = this.projectRepository.createQueryBuilder('project');
+    // .innerJoin('project.reactions', 'reaction')
+    if (category) {
+      query = query.innerJoin(
```
CarlosQ96: To avoid duplicated code we can reuse the query chaining you moved and declared on the projectResolver.ts.

addCategoryQuery

src/resolvers/projectResolver.ts
```diff
@@ -806,136 +907,37 @@ export class ProjectResolver {
...
-      await Reaction.save(newReaction);
-      response.reactionCount = response.reactionCount + 1;
-      response.reaction = true;
+    const viewerUserId = connectedWalletUserId || user?.userId;
+    if (viewerUserId) {
+      query = query.leftJoinAndMapOne(
+        'projectUpdate.reaction',
+        Reaction,
+        'reaction',
+        'reaction.projectUpdateId = CAST(projectUpdate.id AS INTEGER) AND reaction.userId = :viewerUserId',
```
CarlosQ96, line 936: Is it required to CAST as INTEGER?
Both columns are type ID and js wise as Number, maybe the casting isn't required.

CarlosQ96: Is it required to CAST as INTEGER?
Both columns are type ID and TS wise as Number, I don't think the casting is necessary, the tests/query pass without it.

src/resolvers/projectResolver.ts
```diff
@@ -996,31 +998,45 @@ export class ProjectResolver {
...
       .leftJoinAndSelect('project.status', 'status')
       .leftJoinAndMapOne(
         'project.adminUser',
         User,
         'user',
         'user.id = CAST(project.admin AS INTEGER)',
-      )
-      .where('CAST(project.admin AS INTEGER) = :userId', { userId })
+      );
+    // .loadRelationCountAndMap('project.liked', 'project.reactions')
```
CarlosQ96: We can remove the comment. The migrations took care of the count. 👍

src/resolvers/reactionResolver.ts
```diff
@@ -0,0 +1,260 @@
...
+      { select: ['projectId'] },
+    );
+
+    if (!projectUpdate) throw Error(errorMessages.PROJECT_UPDATE_NOT_FOUND);
+
+    try {
+      const reaction = await queryRunner.manager.create(Reaction, {
+        userId: user?.userId,
+        projectUpdateId,
+        reaction: 'heart',
```
CarlosQ96, line 68: Should we move the 'heart' to a constant? I know right now we only have hearts, but maybe will add more in the future. So we don't have heart string hardcoded.
And reuse it for likeProjectUpdate and likeProject.

CarlosQ96 (approved):
LGTM in general!
I did: 
1. All tests pass, I did a local merge with a staging copy to see if tests don't conflict and all good.
2. Checked all code and just left some minor comments. 
3. Didn't see any data deleted with the entity changes locally, so WE ARE GOOD!! and package.json db/test scripts run fine!

Two important details:
1. There is one merge conflict introduced after the merge we did today, nothing dangerous.
2. Need to modify the totalReactions Migration adding this new columns, my local breaks if I don't run the server sync first. Currently prod/staging run migrations before typeorm sync, so it will break deploying.

Side note:
Great Refactors, learned a lot about Typeorm transactions!!

migration/1643716017830-ProjectTotalReactions.ts
```diff
@@ -0,0 +1,25 @@
...
+  async up(queryRunner: QueryRunner): Promise<void> {
+    const [projectTableExists, projectUpdateTableExists, reactionTableExists] =
+      await Promise.all([
+        queryRunner.hasTable('project'),
+        queryRunner.hasTable('project_update'),
+        queryRunner.hasTable('reaction'),
+      ]);
+
+    if (projectTableExists && reactionTableExists)
+      await queryRunner.query(
```
CarlosQ96, line 16: I tried running my migrations locally. But they fail. I think we should add an Alter table query, I have to run the server first so it doesn't break, but this leads to a problem:

In staging/prod migrations are run first, so they will break here as the tables do exist but the column doesn't, so a similar scenario to my local will occur, breaking the deploy.

It doesn't occur on CI/CD because as Mohammad explained me, it runs on a brand new db. So the tables don't exist in this exact instance, hence doesn't break, the if's ignore the totalReaction count.

I think we should add additionally something like: 
```
      `ALTER TABLE IF EXISTS  "project" ADD COLUMN IF NOT EXISTS "totalReactions" integer DEFAULT 0`
      `ALTER TABLE IF EXISTS  "project_update" ADD COLUMN IF NOT EXISTS "totalReactions" integer DEFAULT 0`,
```
and the respective rollback.
@mohammadranjbarz can confirm.

CarlosQ96 (changes requested):

src/resolvers/projectResolver.ts
aminlatifi: I didn't get the point!

aminlatifi: Done, thanks for meticulous review

aminlatifi:
Thank @CarlosQ96 I resolved all your comments

CarlosQ96 (approved):
Great work thanks!

PR #363, opened 2022-02-16
Hotfix: allow development hostnames only

CarlosQ96:
Hotfix to unblock Pedro and Mateo regarding deploys and auth.

### DIFF
--- a/src/user/LoginResolver.ts
+++ b/src/user/LoginResolver.ts
@@ -75,9 +75,7 @@ export class LoginResolver {
   allowHostnameForDevelopment(hostname): boolean {
     if ((config.get('ENVIRONMENT') as string) === 'production') return false;
 
-    const regex = new RegExp(
-      config.get('DEVELOPMENT_HOSTNAME_REGEX') as string,
-    );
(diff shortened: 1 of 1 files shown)


src/user/LoginResolver.ts
```diff
@@ -53,7 +53,11 @@ export class LoginResolver {
     const cache = this.hostnameSignedMessageHashCache;
     if (cache[hostname]) return cache[hostname];
 
-    if (!this.hostnameWhitelist.has(hostname)) return null;
+    if (
+      !this.hostnameWhitelist.has(hostname) ||
```
aminlatifi:
```suggestion
      !this.hostnameWhitelist.has(hostname) &&
```
Current implementation returns null if hostname is not in the whitelist

aminlatifi (changes requested):

src/user/LoginResolver.ts
CarlosQ96: Thanks! No more midnight coding

aminlatifi:
LGTM
Thanks @CarlosQ96

aminlatifi (approved):

Issue #361, opened 2022-02-14
Having Draft Status for Projects and all of it's side-effects


MoeNick:
As the project owner I need to have an option to save the project as draft, (or having a preview of the project) so I can make edits and publish it late. 

**Details:** 
0- we need to have a draft status in DB

1- 
saveProject():
Add an optional field, isDraft otherwise it would be active  by default

2-
getProject(I dont know exactly the name)
For draft status , user should be logged in and be admin of project

3-
activateProject
draft status can be activated as well

PS: The cronjob that make projects listed after a time , should check the project status and not make listed the draft projects
All endpoints should be backward compatible 

4-
Get project list by owner ( we dont know wht's the query name )
My account --> My projects
Should check user authentication and return all projects including draft, cancelled, deactive, active, ...

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/impact-graph/361?planning-poker) with ZenHub @mohammadranjbarz @CarlosQ96

CarlosQ96:
Taking up this issue today and tomorrow.

MoeNick:
> Taking up this issue today and tomorrow.

Appreciate it, shall we have some estimates before we start?

CarlosQ96:
Did points 0, 1, 3, 4 and the cronjob modified.
Regarding 2, I think logic regarding the status is done on the frontend. I'll modified it so it returns null if you are not logged in. Drafts shouldn't be visible. I assume it's the project by slug query and projectbyId

mohammadranjbarz:
@MoeNick 

When getting list of projects of user should we return just `listed` and `active` projects?
I mean when visiting someone else projects not my projects tab

MoeNick:
> @MoeNick
> 
> When getting a list of projects of users should we return just `listed` and `active` projects? I mean when visiting someone else projects not my projects tab

Yes, yes :)

MoeNick:
We have owner view and public view,  listed and active is for getting in all public view (search, home page, projects, and project list of public profile ) 

But for the owner view, all statuses should be visible on my account -> list if projects,

We have a little bit different scenario for Project-Single page which is not the case, but we can discuss it in DM, thanks

mohammadranjbarz:
> We have owner view and public view, listed and active is for getting in all public view (search, home page, projects, and project list of public profile )
> 
[...]

PR #364, opened 2022-02-19
F_361: Add draft status and tests

CarlosQ96:
Related to #361 
Details of what I did per point.
0- we need to have a draft status in DB (added migration)
1- saveProject(): Added isDraft otherwise it would be active by default. (added to projectInput)
2- getProject:
Assumed projectbyslug and projectById. (If not logged in and draft project Ill return null)
3- activateProject: draft status can be activated (this is already allowed).
4- Modified cronjob that list projects to ignore drafted ones.
5- Get project list by owner (Mateo told me is the MeResolve -> MyProjects )
Should check user authentication and return all projects including draft, cancelled, deactive, active, ... (Already does this)

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,15 +264,16 @@ You will need to add the above command to your build process so that all databas
 You can generate table with this site
 https://www.tablesgenerator.com/markdown_tables
 
-| id | symbol        | name          | description                                                                   | who can change to |
-|----|---------------|---------------|-------------------------------------------------------------------------------|-------------------|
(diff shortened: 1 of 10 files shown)


CarlosQ96:
Thanks for the corrections, I think we can merge. I'll merge tomorrow or you can do it too.

mohammadranjbarz:
> Thanks for the corrections, I think we can merge. I'll merge tomorrow or you can do it too.

Thanks @CarlosQ96 , amazing work

Issue #326, opened 2022-01-26
Write test cases for projectsByUserId() in projectResolver.test.ts


mohammadranjbarz:

mohammadranjbarz:
@MoeNick  As this issue is for writing automatic tests, I think you can move it to done

PR #365, opened 2022-02-20
Write automatic tests for projectsByUserId query, made some changes on draft status story

mohammadranjbarz:
related to #326

@CarlosQ96  I made some changes on your PR, and wrote test cases for projectByUserId

### DIFF
--- a/src/resolvers/projectResolver.test.ts
+++ b/src/resolvers/projectResolver.test.ts
@@ -24,6 +24,7 @@ import {
   fetchProjectsBySlugQuery,
   projectByIdQuery,
   walletAddressIsValid,
+  projectsByUserIdQuery,
 } from '../../test/graphqlQueries';
 import { ProjectInput } from './types/project-input';
 import { errorMessages } from '../utils/errorMessages';
@@ -43,6 +44,8 @@ describe('addProject test cases --->', addProjectTestCases);
 describe('editProject test cases --->', editProjectTestCases);
 
 describe('projects test cases --->', projectsTestCases);
+describe('projectsByUserId test cases --->', projectsByUserIdTestCases);
+
 describe('deactivateProject test cases --->', deactivateProjectTestCases);
 describe('activateProject test cases --->', activateProjectTestCases);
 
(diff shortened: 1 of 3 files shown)


src/resolvers/projectResolver.test.ts
```diff
@@ -1695,12 +1809,10 @@ function activateProjectTestCases() {
     const firstUserAccessToken = await generateTestAccessToken(
       SEED_DATA.FIRST_USER.id,
     );
-    const project = await saveProjectDirectlyToDb(createProjectData());
-    const deactiveStatus = await ProjectStatus.findOne({
-      id: ProjStatus.deactive,
```
mohammadranjbarz, line 1700: It was good idea that you get `statusId` when creating test project, so I changed the codes in tests to use that instead of updating status after creating project

src/resolvers/projectResolver.ts
```diff
@@ -538,11 +538,11 @@ export class ProjectResolver {
       });
     query = ProjectResolver.addUserReaction(query, connectedWalletUserId, user);
     const project = await query.getOne();
-    const userId = connectedWalletUserId || user?.userId;
 
     if (
       project?.statusId === ProjStatus.drafted &&
-      (!userId || project?.admin !== String(userId))
```
mohammadranjbarz, line 545: @CarlosQ96 
You should have checked logged in userId not `connectedWalletUserId` because use can send `connectedWalletUserId` and view draft projects of other users

src/resolvers/projectResolver.ts
```diff
@@ -816,16 +816,9 @@ export class ProjectResolver {
...
-        id: ProjStatus.drafted,
-      });
-    } else {
-      status = await this.projectStatusRepository.findOne({
-        id: ProjStatus.active,
-      });
-    }
+    const status = await this.projectStatusRepository.findOne({
+      id: projectInput.isDraft ? ProjStatus.drafted : ProjStatus.active,
+    });
```
mohammadranjbarz, line 821: @CarlosQ96, In this case, conditional (ternary) operator make the code simpler

CarlosQ96 (approved):
Thanks for the corrections !

Issue #372, opened 2022-02-22
Admin Dashboard - Admin roles/permissions and save history of changes


WhyldWanderer:
Eventually we will need more people who can review and verify projects...

As of right now I am the only one on the verification team who can access the panel. I am hesitant to give anyone else access to the Admin Panel until we have roles/permissions and a way to track actions that were taken in the Admin Panel and by who.

Those who have access to this dashboard hold a lot of power and I would hate for someone on the team to drastically change something or delete things without knowing etc. and then not being able to undo any of this.

- [x] Implement roles for Admins

1. Admin: Can only 'list project' 'unlist project' 'verify project' 'reject project' 'revoke badge' 
2. Super Admin: full access the same as it is now.

- [x] Implement a log history of changes made and by who

mohammadranjbarz:
@WhyldWanderer Role and permission is implemented in admin bro, you can create a user with `operator` role and test it in staging env

![Screen Shot 1400-12-09 at 19.31.52.png](https://images.zenhubusercontent.com/5fa7ed7fb0994323d1a66034/fd4070ca-d96b-4a8c-81b7-74eba5f4fe31)

mohammadranjbarz:
@WhyldWanderer 

You can test the second item as well, as you see I am logged in with mate@admin.com and after some changes, the history item appeared, and there si there that who and when made changes

<img width="1920" alt="Screen Shot 1400-12-11 at 09 48 13" src="https://user-images.githubusercontent.com/9850545/156306399-5696c800-5d4f-414a-919e-39d93a7273f9.png">

WhyldWanderer:
Tested and confirmed working on staging!

PR #375, opened 2022-02-22
Implement roles and permissions in admin bro

mohammadranjbarz:
related to #372

PR #378, opened 2022-02-24
F 372 save log history for admin actions

mohammadranjbarz:
related to #372


mohammadranjbarz:
@CarlosQ96  It should be merge after merging #375

CarlosQ96 (changes requested):
I think we should add the createdAt to projectStatus history so that we know when the change was done. This way Ashley can track each change. 

Also, when editing the project we should all add the project status history if the status/listed/verified is changed using the after hook or before hook.
Screen Shot 2022-02-28 at 10 30 00 AM

CarlosQ96 (approved):
LGTM with the added changes!

Issue #367, opened 2022-02-21
Admin Dashboard - Need an export to .csv option


WhyldWanderer:
As an admin, I want to be able to export a set of data into a .csv file.

CarlosQ96:
Writing a middle ware for downloading csv. Might take a bit longer.

divine-comedian:
teting on `serve.admin`

I'm not sure this is working..

on the project page in the admin dashboard I see the option to export filter to CSV I've tried a few things all with the same result:

- selecting a few or all projects on the page with the checkbox and hitting export
- applying a filter and hitting export
- applying a filter then selecting a few projects then hitting export

all attempts to export anything are met with this:
![image](https://user-images.githubusercontent.com/67759413/175644368-6be90857-0afd-4499-bc6c-e138cf902c4f.png)

also, this is only available on the project section, is it intended to be available on other sections as well?

moving this to backlog!

jeevanplus007:
@CarlosQ96 Can you look into this ?

PR #386, opened 2022-03-01
Feature export csv

CarlosQ96:
This PR is related to https://github.com/Giveth/impact-graph/issues/367
What was done:
1. Improve security of adminbro session store moving from in-memoryStore to redis session.
2. Add query cache for csv exporting middleware. (workaround recommended by maintainers of adminbro: https://github.com/SoftwareBrothers/adminjs/issues/947)
3. Added resource action for exporting filters to csv. (if no filter, it exports all).
4. Added google sheets service file. 

Result:
image


CarlosQ96:
Every export will add a new sheet to the selected sheet. (can be private)
To get the env vars and for the exports to work you need to:
1. Create a new project in google cloud services, enable sheets API and generate new credentials json and email.
2. Create a new google spreadsheet, and invite the email (of the previous step) as EDITOR.
3. Get the id in the url of the spreadsheet.
4. set the private key, email and spreadsheet id as env vars.

CarlosQ96:
The name of each new sheet is the current date + epoch of the dateTime. 
Google APi has a bug that doesn't allow formats with ":" so that's why I had to use epoch, at least it works like an unique id

mohammadranjbarz (approved):
LGTM
Thanks Carlos, your idea for exporting to google sheet was amazing, and adding redis session was good as well, I just added some changes on #415  , after merging that and getting gmail account info from Ashley feel free to merge it

CarlosQ96:
Thanks Mohammad! I'll Demo Ashley the feature and then decide to merge. But no rush to include this in the new versión (we can do it, right after).

PR #387, opened 2022-03-01
Added orderBy option to getProjectUpdate endpoint

aminlatifi:


src/resolvers/projectResolver.ts
```diff
@@ -1036,9 +1036,17 @@ export class ProjectResolver {
...
     @Arg('take', type => Int, { defaultValue: 10 }) take: number,
     @Arg('connectedWalletUserId', type => Int, { nullable: true })
+    @Arg('orderBy', type => OrderBy, {
+      defaultValue: {
+        field: OrderField.CreationAt,
+        direction: OrderDirection.DESC,
+      },
+    })
+    orderBy: OrderBy,
     connectedWalletUserId: number,
```
CarlosQ96: Move this variable to it's corresponding Argument.

CarlosQ96:
```javascript
@Arg('connectedWalletUserId', type => Int, { nullable: true })
connectedWalletUserId: number,
```

src/resolvers/projectResolver.ts
```diff
@@ -1047,6 +1055,7 @@ export class ProjectResolver {
           projectId,
         },
       )
+      .orderBy(field, direction)
```
CarlosQ96: For some reason, typeorm not recognizing the column. 
Usually we interpolate the entity like this.
```javascript
.orderBy(`projectUpdate.${field}`, direction)
```

CarlosQ96 (changes requested):

src/resolvers/projectResolver.ts
aminlatifi: Nice, good catch

CarlosQ96 (approved):
LGTM thanks man!

PR #392, opened 2022-03-02
F 372 save log history for admin actions revision

mohammadranjbarz:


CarlosQ96 (approved):
Thanks man, I think this is good!

PR #405, opened 2022-03-07
Feat/sort by work as filter

aminlatifi:
@MoeNick It makes sort by traceable and accept giv as filter

### DIFF
--- a/src/resolvers/projectResolver.ts
+++ b/src/resolvers/projectResolver.ts
@@ -469,7 +469,10 @@ export class ProjectResolver {
             orderBy.direction === OrderDirection.ASC ? '' : ' NOT'
           } NULL`,
         );
-        query.orderBy(`project.${OrderField.CreationAt}`, OrderDirection.DESC);
+        query.orderBy(
+          `project.${OrderField.CreationDate}`,
+          OrderDirection.DESC,
+        );
         break;
       case OrderField.AcceptGiv:
         // const acceptGivDirection: {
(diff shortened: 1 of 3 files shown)


CarlosQ96:
It works, and test pass. 
However, my only concern is, that if it works like a filter, and is ordered by ID, what if later on in the designs they add sort by desc/asc? we should add some form of orderby -> creationDate -> desc/asc to each of those.
So the filter can be reordered.

This won't break the nextjs version, so it's mergeable, as they use the filterBy option I added.

src/resolvers/projectResolver.ts
```diff
@@ -449,20 +449,49 @@ export class ProjectResolver {
...
-        `project.${orderBy.field}`,
-        orderBy.direction,
-        // @ts-ignore
-        traceableDirection[orderBy.direction],
-      );
-    } else {
-      query.orderBy(`project.${orderBy.field}`, orderBy.direction);
+    switch (orderBy.field) {
+      case OrderField.Traceable: // TODO: PRISMA will fix this, temporary fix inverting nulls.
+        // const traceableDirection: {
```
CarlosQ96, line 454: I guess we can save the snippets somewhere and remove the comments. Everything else looks ok.

CarlosQ96 (approved):
This won't break the nextjs version and is mergeable. Let me know if you add the additional orderBy for each filter. 
But I think looking at the design this will fulfill the requirement.

CarlosQ96 (approved):
LGTM, thanks Amin.

PR #408, opened 2022-03-08
Feature: Search by project owner name

CarlosQ96:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/126

InnerJoin makes more sense, there is no project that shouldn't have an Admin


CarlosQ96:
image
So searching by james name, returns both a project with james on the title and a project with james as the owner

mohammadranjbarz (approved):
Nice job @CarlosQ96 
Thank you

LGTM

PR #413, opened 2022-03-15
Refactor PR 395

aminlatifi:
Refactored seed token add migration


src/seedToken-ormconfig.ts
```diff
@@ -0,0 +1,24 @@
...
+  password: process.env.TYPEORM_DATABASE_PASSWORD,
+  database: process.env.TYPEORM_DATABASE_NAME,
+  entities,
+  migrations: ['migration/1646303882607-seedTokes.ts'],
+  cli: {
+    migrationsDir: 'migration',
+  },
+};
+
+export = ormConfig;
```
mohammadranjbarz, line 24: @aminlatifi Do we need this file?

aminlatifi: I made it to test token seed migration individually. We can remove it or keep it as an example for next migration test

mohammadranjbarz: Yeah you are right, it's not bad to have it

mohammadranjbarz (approved):
@aminlatifi  Thanks Amin, the improvements are huge, I learned from your changes, 
I ran your migrations locally and tests passed, and codes seems ok, I will merge it now

PR #414, opened 2022-03-16
Support purple list in projects

mohammadranjbarz:
related Giveth/giveth-next#671, Giveth/giveth-next#642


mohammadranjbarz:
@CarlosQ96 JFYI

Screen Shot 1401-01-14 at 17 29 50

Screen Shot 1401-01-14 at 17 29 30

src/resolvers/projectResolver.ts
```diff
@@ -1049,6 +1052,27 @@ export class ProjectResolver {
...
+            `,
+    );
+    const recipientsAddresses = recipients.map(({ walletAddress }) =>
+      walletAddress.toLowerCase(),
+    );
+    const purpleAddresses = await PurpleAddress.query(
+      `
+            SELECT address FROM purple_address
+            `,
+    );
```
CarlosQ96, line 1070: I think we can offload the LowerCase function to the database and get some extra performance. But I didn't find a way for typeorm to map directly to an array sadly, other ORMS provide this functionality.

CarlosQ96: Like so:
```typescript
  @Query(returns => [String])
  async getPurpleList(): Promise<String[]> {
    const recipients = await Project.query(
      `
          SELECT LOWER("walletAddress") as recipient
          FROM project
          WHERE verified=true
      `,
    );
    const recipientsAddresses = recipients.map(({ recipient }) => recipient);
    const purpleAddresses = await PurpleAddress.query(
      `
          SELECT LOWER(address) as "purpleAddress"
          FROM purple_address
      `,
    );
    return purpleAddresses
      .map(({ purpleAddress }) => purpleAddress)
      .concat(recipientsAddresses);
  }
```   
```

File: src/entities/purpleAddress.ts [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+  @Field()
+  @Column({ unique: true })
+  address: string;
+
+  @Index()
+  @Field(type => Project)
+  @ManyToOne(type => Project, { eager: true })
+  project: Project;
+  @RelationId((purpleAddress: PurpleAddress) => purpleAddress.project)
+  projectId: number;
```
CarlosQ96, line 36: I have a question, we have the projectId here but we are not using it on the queries below. 
This is only to track from what project is the address from, right? Because a purple list address is banned from getting givBacks from all projects, it's what I read on the issues you linked.

CarlosQ96 (approved):
Hey Mohammad! Good work, I think the code is good, I left some comments suggesting minor improvements. Test pass with those.
Also tested the admin and all good!

src/entities/purpleAddress.ts
CarlosQ96: Or should we use the projectId for filtering? I don't so tho.

src/resolvers/projectResolver.ts
CarlosQ96: Left a PR for this 👍

mohammadranjbarz: Thank you man, I merged your PR

src/entities/purpleAddress.ts
mohammadranjbarz: @CarlosQ96 We dont use projectId in our query, but it's can be viewed in admin-bro, so admins can know this purple address is related to what project, or they can see the purple addresses of a project in admin-bro, so I think it's useful to have it, we can manage purple addresses easier in this way

CarlosQ96: Great, Thanks! I think you can merge!

mohammadranjbarz: And all purple addresses are related to some projects, so it's better to reflect that relationship in our DB

PR #415, opened 2022-03-16
Feature export csv revision

mohammadranjbarz:
@CarlosQ96  I thought we need some refactor, but because I didn't want to make review long I made this PR, these are what I did:

1. Refactor some function and specify type of input parameters
2. Add one test case for failed scenario of export CSV


CarlosQ96 (approved):
I agree with the changes! thanks man for the error handling.

PR #425, opened 2022-03-17
hotfix: Editing project would not change slug (if title didnt change)

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#265

@aminlatifi the `addProject` and `editProject` is not used anymore we just kept it for being backward compatible, because painting test cases for those mutations took time I removed them completely

**PS** If you think it's better to merge it to staging first, we can change the target branch


aminlatifi (approved):

aminlatifi:
Thanks @mohammadranjbarz

PR #429, opened 2022-03-19
F 113 search by token address amount

CarlosQ96:
Related to improvements to https://github.com/Giveth/giveth-dapps-v2/issues/113

### DIFF
--- a/src/resolvers/donationResolver.test.ts
+++ b/src/resolvers/donationResolver.test.ts
@@ -170,6 +170,63 @@ function donationsByProjectIdTestCases() {
       DONATION_SEED_DATA.FIFTH_DONATION.id,
     );
   });
+  it('should search by donation amount', async () => {
+    const result = await axios.post(
+      graphqlUrl,
+      {
+        query: fetchDonationsByProjectIdQuery,
+        variables: {
+          projectId: SEED_DATA.FIRST_PROJECT.id,
+          searchTerm: '100',
+        },
+      },
+      {},
+    );
+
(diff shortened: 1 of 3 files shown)


src/resolvers/donationResolver.test.ts
```diff
@@ -170,6 +170,63 @@ function donationsByProjectIdTestCases() {
...
+          projectId: SEED_DATA.FIRST_PROJECT.id,
+          searchTerm: '100',
+        },
+      },
+      {},
+    );
+
+    const donations = result.data.data.donationsByProjectId.donations;
+    assert.equal(
+      Number(donations[0]?.id),
```
mohammadranjbarz: @CarlosQ96  I think it should be better instead of checking id of returned donation you can check the amount of returned donation, or other test case you can check equality of donation.currency, because when you are checking the id the test would fail if create donations with same amount or same currency in other part of our test codes, so it can't be reliable

mohammadranjbarz (approved):
@CarlosQ96 Awesome
I think the test cases can be improved as I mentioned, other part of codes seems fine

mohammadranjbarz (changes requested):
The codes seems good to me

CarlosQ96:
Thanks for the feedback @mohammadranjbarz ! Improved the tests.

mohammadranjbarz:
> Thanks for the feedback @mohammadranjbarz ! Improved the tests.

@CarlosQ96 
Awesome, great work as always

Issue #430, opened 2022-03-20
Mask internal server errors before sending response to client


mohammadranjbarz:
When there is a probelm in connecting to db our apollo server return host of DB in response, I simulated it in my local system by stopping postgress container and Got this

![Screen Shot 1400-12-29 at 07.42.23.png](https://images.zenhubusercontent.com/5fa7ed7fb0994323d1a66034/8d714d1a-9f31-40dd-b583-10f9afbb0c6e)

I think it's security vulnurability

PR #431, opened 2022-03-20
Return general error if we had db error

mohammadranjbarz:
related to #430

@aminlatifi I think we should deploy a hotfix to production after you merged this

### DIFF
--- a/src/server/bootstrap.ts
+++ b/src/server/bootstrap.ts
@@ -144,7 +144,7 @@ export async function bootstrap() {
           err?.message?.includes(process.env.TYPEORM_DATABASE_HOST as string)
         ) {
           logger.error('DB connection error', err);
-          return new Error(errorMessages.INTERNAL_SERVER_ERROR);
+          return new Error(errorMessages.ERROR_CONNECTING_DB);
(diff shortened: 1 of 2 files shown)


mohammadranjbarz:
* **Before**
Screen Shot 1400-12-29 at 07 42 23


* **After**
Screen Shot 1400-12-29 at 07 56 03

If you think it's not needed to say in response that we have DB problem, we can return **Internal server error** there as well

src/server/bootstrap.ts
```diff
@@ -132,6 +134,27 @@ export async function bootstrap() {
...
+        /**
+         * @see {@link https://www.apollographql.com/docs/apollo-server/data/errors/#for-client-responses}
+         */
+        // Don't give the specific errors to the client.
+
+        if (
+          err?.message?.includes(process.env.TYPEORM_DATABASE_HOST as string)
+        ) {
+          logger.error('DB connection error', err);
+          return new Error(errorMessages.ERROR_CONNECTING_DB);
```
aminlatifi, line 147: Is this error returned to the end-user? If yes it would be better to return internal server error and not make the user involved

Issue #336, opened 2022-01-26
Write test cases for userByAddress() in userResolver.test.ts


mohammadranjbarz:

mohammadranjbarz:
It has been implemented on #357

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/impact-graph/336?planning-poker) with ZenHub @CarlosQ96 @mohammadranjbarz

CarlosQ96:
@mohammadranjbarz I saw you added tests for this already, added one for an empty case and thats it. But this is practically done.

mohammadranjbarz:
> @mohammadranjbarz I saw you added tests for this already, added one for an empty case and thats it. But this is practically done.

Thank you man

PR #435, opened 2022-03-22
add empty test case for userByAddress

CarlosQ96:
Main test already existed. Added empty case.
Should we add any more tests? I think it's enough.
Related to https://github.com/Giveth/impact-graph/issues/336


mohammadranjbarz (approved):
Cool, I think it's enough, thanks @CarlosQ96

PR #436, opened 2022-03-22
ISImported column

CarlosQ96:
IsImported column migration only, for future integrations.
Pre Step to the change api integration.

### DIFF
--- a/src/server/adminBro.ts
+++ b/src/server/adminBro.ts
@@ -481,6 +481,9 @@ const getAdminBroInstance = () => {
                 edit: false,
               },
             },
+            isImported: {
+              isVisible: { list: false, filter: false, show: true, edit: true },
+            },
             totalReactions: {
               isVisible: false,
             },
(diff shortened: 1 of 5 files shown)


CarlosQ96:
Extracted From the Revoke badge pr

src/server/adminBro.ts
```diff
@@ -481,6 +481,9 @@ const getAdminBroInstance = () => {
                 edit: false,
               },
             },
+            isImported: {
+              isVisible: { list: false, filter: false, show: true, edit: true },
+            },
```
mohammadranjbarz: As `isImported` is a variable that depends on `traceCampaignId` and `givingBlockId`,.. I think we should not add it in edit page , because we should not give this option to admins to can change it because it might cause some inconsistency in our system

mohammadranjbarz (changes requested):
Thanks @CarlosQ96  LGTM

PR #437, opened 2022-03-22
Add mainnetAddress to token entity and write migration to fill it

mohammadranjbarz:
@aminlatifi  @CarlosQ96  Ramin wanted this, so I have set both of you as reviewer then any of you that had time sooner review it

### DIFF
--- a/migration/1646295724658-createTokensTable.ts
+++ b/migration/1646295724658-createTokensTable.ts
@@ -12,6 +12,7 @@ export class createTokensTable1646295724658 implements MigrationInterface {
           address text COLLATE pg_catalog."default" NOT NULL,
           "networkId" integer NOT NULL,
           decimals integer NOT NULL,
+          "mainnetAddress" text,
           CONSTRAINT "PK_82fae97f905930df5d62a702fc9" PRIMARY KEY (id)
            )`,
     );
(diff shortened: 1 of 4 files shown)


aminlatifi (changes requested):
@mohammadranjbarz The migration needs some fix
```
  driverError: error: column "mainnetAddress" of relation "token" does not exist
```

migration/1646295724658-createTokensTable.ts
```diff
@@ -12,6 +12,7 @@ export class createTokensTable1646295724658 implements MigrationInterface {
           address text COLLATE pg_catalog."default" NOT NULL,
           "networkId" integer NOT NULL,
           decimals integer NOT NULL,
+          "mainnetAddress" text,
```
aminlatifi, line 15: I doubt this will fix the issue on machines has run this migration before. I mean we need to alter table in `migration/1647927483869-fillMainnetAddressForSomeTokens.ts`

aminlatifi (changes requested):

migration/1646295724658-createTokensTable.ts
mohammadranjbarz: I have altered table in `fillMainnet...` before, 
It fixed our migrations, the `Run migrations` passed in Github Action job

mohammadranjbarz:
> @mohammadranjbarz The migration needs some fix
> 
> ```
>   driverError: error: column "mainnetAddress" of relation "token" does not exist
> ```

Thanks Amin, it has been fixed

aminlatifi (approved):
LGTM!
Thanks @mohammadranjbarz

Issue #390, opened 2022-03-01
Importing projects from CHANGE API in our staging


MoeNick:

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/impact-graph/390?planning-poker) with ZenHub @CarlosQ96 @mohammadranjbarz

CarlosQ96:
To test this, go into the adminbro:
Enter **Third Party Project Imports** link to the left. It will take you to this page where you will be able to create a new import from the changeAPI.
<img width="1480" alt="image" src="https://user-images.githubusercontent.com/92376054/160735694-61a1840d-77b9-4a79-99d1-c57a33875b3f.png">

This project should be visible in the site. To import the project write the exact title or IEN from the change website. Example: 
https://getchange.io/nonprofit/?address=D1XBJG85gs7DT43TraBBpUiLpDBQ8zrTLyjneP5onaWw

Title: Electronic Frontier Foundation
The EIN is on the bottom of the page: 
<img width="120" alt="image" src="https://user-images.githubusercontent.com/92376054/160735869-52467255-3f0f-4576-9f4f-eeae976c9bd0.png">


Form:
<img width="1464" alt="image" src="https://user-images.githubusercontent.com/92376054/160735964-af4eb207-6e0f-4818-8909-68bf81b08676.png">

CarlosQ96:
Project duplication is not allowed. Honestly this is pointing to prod projects because there is no staging on the changeAPI.
I wrote some tests for testing donations.

CarlosQ96:
@MoeNick you can search more projects in this link https://getchange.io/nonprofit/?address=D1XBJG85gs7DT43TraBBpUiLpDBQ8zrTLyjneP5onaWw in the search bar.
And import them.
I think this is ready to be merged after testing in staging.

MoeNick:
Good job @CarlosQ96 Thanks
The funny thing is we say it's a GivingBlock project on our UI. Like this: 
https://giveth-dapps-v2.vercel.app/project/alzheimers-association
Do we set proper Org config for them?

CarlosQ96:
Hello @MoeNick I think the frontend has to do a change, thanks for finding it.
On the db it has the proper org
[...]

PR #440, opened 2022-03-22
F 390 change api admin integration

CarlosQ96:
Related to https://github.com/Giveth/impact-graph/issues/390
Did:
1. Fixed organization migration for givingblocks and added the additional attributes.
2. Refactored some logic for quality score and slug
3. Added migrations to related ropsten tokens and mainnet eth to the projects of  CHANGE. Include isImported flag.
4. Created an import record for history page on adminbro -> linking to the respective user and project imported
image
5. Connect to the ChangeAPI and imports projects using the exact title or EIN ID. 
Example: https://getchange.io/nonprofit/?address=8EhZ7waiWj8SW1KXkAieCd5N6cRjXNJKZfES2WYz9Uva
image
or
image


CarlosQ96:
@mohammadranjbarz let me know your feedback! DM you my public key of changeApi

migration/1648085509369-addGivingBlockOrganizationToGivingBlockProjects.ts
```diff
@@ -0,0 +1,26 @@
...
+              WHERE label='givingBlock'`)
+    )[0];
+
+    if (!givingBlockOrganization) return;
+    await queryRunner.query(`
+            UPDATE project
+            SET "organizationId" = ${givingBlockOrganization.id}
+            WHERE "givingBlocksId" IS NOT NULL
+        `);
+  }
```
mohammadranjbarz, line 23: I think this is redundant because we related givingBlocks projects to organization before in this migration file

https://github.com/Giveth/impact-graph/blob/staging/migration/1646306281286-relateExistingProjectsToOrganizations.ts

migration/1648066794387-addChangeAcceptedtokens.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    // for both production and staging
+    await queryRunner.query(
+      `INSERT INTO organization_tokens_token ("tokenId","organizationId") VALUES
+        (${mainnetNativeToken.id}, ${changeOrganization.id})
+      ;`,
+    );
+
+    // add test token for any other env
+    if (config.get('NODE_ENV') !== 'production') {
```
mohammadranjbarz: We have problem and that is we have set `NODE_ENV=produciton` in all of our environments (develop, staging, production)
so this part will never run on our servers ( we might add new variable in config for specifying the environment, because I don't know how much side effects would have changing NODE_ENV ins taging

src/entities/thirdPartyProjectImport.ts
```diff
@@ -0,0 +1,53 @@
...
+  @Column({ nullable: true })
+  projectName: string;
+
+  // History of who exported
+  @Field(type => User)
+  @ManyToOne(type => User)
+  user?: User;
+
+  @RelationId(
+    (thirdPartyProjectExport: ThirdPartyProjectImport) =>
```
mohammadranjbarz: I think you named it `thirdPartyProjectExport` by mistake

src/entities/thirdPartyProjectImport.ts
```diff
@@ -0,0 +1,53 @@
...
+  userId: number;
+
+  // Link to project
+  @Field(type => Project)
+  @ManyToOne(type => Project)
+  project?: Project;
+
+  @RelationId(
+    (thirdPartyProjectExport: ThirdPartyProjectImport) =>
+      thirdPartyProjectExport.project,
```
mohammadranjbarz: I think you named it thirdPartyProjectExport by mistake

src/services/changeAPI/nonProfits.ts
```diff
@@ -0,0 +1,157 @@
...
+  zip_code?: string;
+}
+
+// exact title returns 1 element
+export const getChangeNonProfitByNameOrIEN = async (
+  nonProfit: String,
+): Promise<ChangeNonProfit> => {
+  try {
+    const result = await Axios.get(
+      'https://api.getchange.io/api/v1/nonprofits',
```
mohammadranjbarz: It's better to put this url in config

src/services/changeAPI/nonProfits.ts
```diff
@@ -0,0 +1,157 @@
...
+    twitter?: string;
+    youtube?: string;
+  };
+  state?: string;
+  website?: string;
+  zip_code?: string;
+}
+
+// exact title returns 1 element
+export const getChangeNonProfitByNameOrIEN = async (
```
mohammadranjbarz, line 53: It might be good if you can write test for change APIs, if you want to do that, you have to pass CHANGE API token in github actions like this (https://github.com/Giveth/impact-graph/blob/staging/.github/workflows/CI-CD.yml#L46-L53)
But I don't know if you use a specific project's name, when the project name changes our test will fail, so please do everything you think is fine, I just suggested that

mohammadranjbarz (changes requested):
Thanks  @CarlosQ96 , great implementation , the main codes of PR was great I just had few tiny feedbacks

migration/1648085509369-addGivingBlockOrganizationToGivingBlockProjects.ts
CarlosQ96: That's true, but remember our cronjob kept adding projects, so those are not related. (at least no in my local which I keep synchronizing with givingblocks). We did not update the givingBlocks cronjob, so with this migration I fix those new projects, and also I fixed the givingblocks cronjob to add the organization.

mohammadranjbarz: Oh got it, you are absolutely right

migration/1648066794387-addChangeAcceptedtokens.ts
CarlosQ96: Thanks, we had another variable called Environment.

src/services/changeAPI/nonProfits.ts
CarlosQ96: Good Suggestion. Ill actually write two tests, one for simulating the API for the function createProjectFromChangeNonProfit and another for importThirdPartyProject, simulating the error messages.

But we shouldn't add api calls to our test suite in my opinion, as it will slow down our test suite and if their api is down, so is our tests. Calling apis is somewhat slow from what I tested.

Even if the API is down, we log and throw the errors.

CarlosQ96:
Thanks for the feedback Mohammad. Will be merging.

Issue #441, opened 2022-03-23
Support custom tokens for Giveth (not third party) projects


aminlatifi:

mohammadranjbarz:
- [x] Set a field name `supportCustomToken` to `organization` table ( with `false` as default value)
- [x] Write a migration file to set this field `true`  `giveth` and `trace` organization
- [x] In `donation` table we should add a field `isTokenEligibleForGivback`
- [x] in `saveDonation` mutation we should check if sent token is not in project(organzation) token list , we check if organization support custom token we allow user to donate but we set `isTokenEligibleForGivback:false`  for this donations ( other donations should have `isTokenEligibleForGivback:true` )
- [x] Return organization (join table) in `projects` query response
- [x] Return organization (join table) in `projectById` query response
- [x] Return organization (join table) in `projectBySlug` query response
- [x] Return organization (join table) in `projectByAddress` query response

CarlosQ96:
The isTokenEligibleForGivbacks should go in the token table also too no?
Maybe other tokens might be in the list but not elegible for givbacks. Correct me if I am wrong with the business logic.

Everything else seems good.
Do we still use ```projectByAddress``` ? if so we should also return the organization there.

mohammadranjbarz:
> The isTokenEligibleForGivbacks should go in the token table also too no? Maybe other tokens might be in the list but not elegible for givbacks. Correct me if I am wrong with the business logic.
> 
> Everything else seems good. Do we still use `projectByAddress` ? if so we should also return the organization there.

Thanks, nice suggestion about `projectByAddress`

`isTokenEligibleForGivbacks` can't be in token, because a token can be eligible for givback for an organization but for another organization it can be not-eligible, so when creating donation we should consider that and fill this field in donation ( and givback-calculation would check that field

mohammadranjbarz:
@RamRamez 
After merging PR of this issue you can get organization when getting project/projects
example:

```
query{
  projectBySlug(slug:"how-many-photos-is-too-many-photos"){
    slug
    slugHistory
    updatedAt
    listed
    id
    title
    traceCampaignId
    totalReactions
    totalDonations
    image
    totalTraceDonations
    status{
        id
        name
    }
    organization {
    name
    label
    supportCustomTokens
    }
  }
}
```

mohammadranjbarz:
> projectByAddress

I asked @RamRamez  and he said they dont use `projectByAddress` in frontend

mohammadranjbarz:
[...]

PR #442, opened 2022-03-24
Support custom tokens for giveth and trace projects

mohammadranjbarz:
related to #441

### DIFF
--- a/migration/1648103938557-addSupportCustomTokensToOrganizations.ts
+++ b/migration/1648103938557-addSupportCustomTokensToOrganizations.ts
@@ -1,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { ORGANIZATION_LABELS } from '../src/entities/organization';
+
+export class addSupportCustomTokensToOrganizations1648103938557
+  implements MigrationInterface
+{
+  async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        ALTER TABLE organization
+        ADD COLUMN  IF NOT EXISTS "supportCustomTokens" boolean DEFAULT false
+     `);
+
+    await queryRunner.query(`
+         UPDATE organization
+         SET "supportCustomTokens" = true
+         WHERE label='${ORGANIZATION_LABELS.GIVETH}'
+      `);
+
+    await queryRunner.query(`
+         UPDATE organization
+         SET "supportCustomTokens"=true
+         WHERE label='${ORGANIZATION_LABELS.TRACE}'
+      `);
+  }
+
+  async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      ALTER TABLE organization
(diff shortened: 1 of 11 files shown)


migration/1648103938557-addSupportCustomTokensToOrganizations.ts
```diff
@@ -0,0 +1,32 @@
...
+        ADD COLUMN  IF NOT EXISTS "supportCustomTokens" boolean DEFAULT false
+     `);
+
+    await queryRunner.query(`
+         UPDATE organization
+         SET "supportCustomTokens" = true
+         WHERE label='${ORGANIZATION_LABELS.GIVETH}'
+      `);
+
+    await queryRunner.query(`
```
CarlosQ96: This can be done in one query.

src/resolvers/projectResolver.ts
```diff
@@ -564,6 +562,7 @@ export class ProjectResolver {
       })
       .leftJoinAndSelect('project.status', 'status')
       .leftJoinAndSelect('project.categories', 'categories')
+      .leftJoinAndSelect('project.organization', 'organization')
```
CarlosQ96, line 565: Ill add this to projectbyAddress

src/resolvers/donationResolver.ts
```diff
@@ -366,14 +370,20 @@ export class DonationResolver {
...
+      let isTokenEligibleForGivback = false;
+      if (!tokenInDb && !project.organization.supportCustomTokens) {
+        throw new Error(errorMessages.TOKEN_NOT_FOUND);
+      } else if (tokenInDb) {
+        const acceptsToken = await isTokenAcceptableForProject({
+          projectId,
+          tokenId: tokenInDb.id,
+        });
+        if (!acceptsToken && !project.organization.supportCustomTokens) {
+          throw new Error(errorMessages.PROJECT_DOES_NOT_SUPPORT_THIS_TOKEN);
```
CarlosQ96, line 382: I think we shouldn't stop the process from saving the donation, because even if the project don't support givbacks for this donation with that token, the donation was still done, so we should keep a record of that.

So we should log the error and set isTokenElegibleForGivback = false.
Because what I am seeing from the tests and code, is there is no case in which it is false.

CarlosQ96 (changes requested):
Thanks Mohammad great work. I just had some small comments and a doubt.
Will fix the comments with another Pr before merging.

src/resolvers/donationResolver.ts
CarlosQ96: Actually I think the case we are missing is. If it accepts custom tokens but its not in db. We should let this case pass and not throw error.

CarlosQ96: Does this case make isTokenEligibleForGivback = false? Will Confirm tomorrow.

CarlosQ96 (approved):
Just confirmed all tokens in list are elegible for givbacks so The code LGTM now!

The test confirms the case when a project accepts custom tokens and the token is not in the db!

Merging!

PR #448, opened 2022-03-29
Hotfix prevent draft status from notifying creation

CarlosQ96:
It didn't publish the project, as only active projects are the ones that the 2 week cronjob lists. 

The email was sent on creation, so preventing that.
For now I disabled the email, but maybe changing the segment event would be better

Related to https://github.com/Giveth/impact-graph/issues/449


MoeNick:
Unfortunately, I can't test it on UAT, due to it won't send emails on this environment. ANy ideas?

PR #455, opened 2022-04-02
Write test case for donations to wallets

FatemeGhasemi:
Add test for donationsToWallets
Related to #329


FatemeGhasemi:
@mohammadranjbarz can you review this?

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1572,96 @@ function donationsByDonorTestCases() {
...
+  const walletAddress1 = generateRandomEtheriumAddress();
+  const walletAddress2 = generateRandomEtheriumAddress();
+  const walletAddress3 = generateRandomEtheriumAddress();
+  const walletAddress4 = generateRandomEtheriumAddress();
+  it('should find donations with special destination successfully', async () => {
+    const project = await saveProjectDirectlyToDb(createProjectData());
+    const user1 = await User.create({
+      walletAddress: walletAddress1,
+      loginType: 'wallet',
+      firstName: 'bacheTest',
```
mohammadranjbarz: Please change `bacheTest` to a meaningful name

src/resolvers/donationResolver.test.ts
```diff
@@ -38,6 +39,7 @@ describe('donationByUserId() test cases', donationsByUserIdTestCases);
 // describe('tokens() test cases', tokensTestCases);
 describe('donationsByDonor() test cases', donationsByDonorTestCases);
 describe('saveDonation() test cases', saveDonationTestCases);
+describe('donationsToWallets() test cases', donationsToWalletsTestCases);
```
mohammadranjbarz, line 40: @FatemeGhasemi  You could uncomment 36th line instead of adding this line

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1572,96 @@ function donationsByDonorTestCases() {
...
+      },
+      {},
+    );
+    let test = true;
+    result.data.data.donationsToWallets.forEach(item => {
+      if (item.toWalletAddress !== walletAddress1) {
+        test = false;
+      }
+    });
+    assert.equal(test, true);
```
mohammadranjbarz: Instead of what you have written, you can use this

```
  result.data.data.donationsToWallets.forEach(item => {
      assert.equal(item.toWalletAddress, walletAddress1)
    });
```

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1572,96 @@ function donationsByDonorTestCases() {
     );
   });
 }
+
+function donationsToWalletsTestCases() {
+  const walletAddress1 = generateRandomEtheriumAddress();
+  const walletAddress2 = generateRandomEtheriumAddress();
+  const walletAddress3 = generateRandomEtheriumAddress();
+  const walletAddress4 = generateRandomEtheriumAddress();
```
mohammadranjbarz: It's better to define all needed variables inside the test case, not in  `donationsToWalletsTestCases()` function, you can move it to arrow function of test case, so in there you can just say `walletAddress` and you don't need to add a  number suffix to them

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1572,96 @@ function donationsByDonorTestCases() {
...
+      },
+      {},
+    );
+    let test = false;
+    result.data.data.donationsToWallets.forEach(item => {
+      if (item.toWalletAddress === walletAddress2) {
+        test = true;
+      }
+    });
+    assert.equal(test, false);
```
mohammadranjbarz: Please change as I said above

mohammadranjbarz (changes requested):
@FatemeGhasemi 
Thanks so much, I left some comments for you, and please write a test case when you send a wallet address ( random etherum address) that you know there is no donation to that address, and you will get a list with length `0` in response

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1570,167 @@ function donationsByDonorTestCases() {
...
+      {
+        query: donationsToWallets,
+        variables: {
+          toWalletAddresses: [walletAddress1],
+        },
+      },
+      {},
+    );
+    result.data.data.donationsToWallets.forEach(item => {
+      assert.notEqual(item.toWalletAddress, walletAddress1);
```
mohammadranjbarz: please create a donation with that have `walletAddress1` as `toWalletAddress` as well

src/resolvers/donationResolver.test.ts
```diff
@@ -1570,3 +1570,167 @@ function donationsByDonorTestCases() {
...
+    do {
+      walletAddress = generateRandomEtheriumAddress();
+    } while (
+      (
+        await Donation.find({
+          where: { toWalletAddress: walletAddress },
+        })
+      ).length !== 0
+    );
+
```
mohammadranjbarz: Why dont you use just
```
await Donation.find({
          where: { toWalletAddress: walletAddress },
        })
```     
What's the point of this `while` loop?

mohammadranjbarz (changes requested):
Thanks for fixing last request changes, I left two other comments

mohammadranjbarz (approved):
@FatemeGhasemi  Great work, LGTM

Issue #315, opened 2022-01-26
Write test cases for addProjectUpdate() in projectResolver.test.ts


mohammadranjbarz:

CarlosQ96:
Moving to done. Tests were added.

CarlosQ96:
@divine-comedian

PR #460, opened 2022-04-04
Write test cases for addProjectUpdate

FatemeGhasemi:
Write test cases for addProjectUpdate
Related to #315


FatemeGhasemi:
@mohammadranjbarz can you please review this?

src/resolvers/projectResolver.test.ts
```diff
@@ -2964,3 +2966,192 @@ function similarProjectsBySlugTestCases() {
...
+  });
+  it('should can not add project update because user not found ', async () => {
+    const user = await User.create({
+      walletAddress: generateRandomEtheriumAddress(),
+      loginType: 'wallet',
+      firstName: 'testEditProjectUpdateFateme',
+    }).save();
+    const accessToken = await generateTestAccessToken(user.id);
+    await User.delete({ id: user.id });
+    const projectCount = await Project.count();
```
mohammadranjbarz, line 3189: @FatemeGhasemi  get projectId from create project response, instead of getting id with getting count of projects

src/resolvers/projectResolver.test.ts
```diff
@@ -2964,3 +2966,192 @@ function similarProjectsBySlugTestCases() {
...
+        title: 'testProjectUpdateFateme2',
+      },
+    });
+
+    assert.equal(
+      result.data.errors[0].message,
+      errorMessages.AUTHENTICATION_REQUIRED,
+    );
+  });
+  it('should can not add project update because user not found ', async () => {
```
mohammadranjbarz, line 3181: It's good to have a test case like this, for when adding project update, but projected is invalid ( project doesn't exists)

mohammadranjbarz (changes requested):
Thanks @FatemeGhasemi  , everything seems fine, just please fix requested changes

src/resolvers/projectResolver.test.ts
FatemeGhasemi: In this test I don't create new project so I can't get id from it.

src/resolvers/projectResolver.test.ts
FatemeGhasemi: I wrote it. did you mean 
https://github.com/FatemeGhasemi/impact-graph/blob/5c54fee14ed94887ab25ad88a4824f5340bf2e6c/src/resolvers/projectResolver.test.ts#L3083

src/resolvers/projectResolver.test.ts
mohammadranjbarz: @FatemeGhasemi  Oy you are right, my mistake

src/resolvers/projectResolver.test.ts
mohammadranjbarz: Good job, thank you

FatemeGhasemi:
@mohammadranjbarz Conflicts are resolved :)

PR #464, opened 2022-04-05
Improve async query for retrieving addresses

CarlosQ96:
If you think this is good you can merge!
I think the rest of the PR lgtm!


mohammadranjbarz (approved):
Nice job, it made our code more clear, thanks

Issue #463, opened 2022-04-04
Add an adminbro tab to create new tokens that are accepted and relate it to organizations.


CarlosQ96:

mohammadranjbarz:
@MoeNick @WhyldWanderer 
Now the PR of @CarlosQ96 is depoyed on UAT, 
you can test it now, I made a checklist for anyone who wants to test
- [x] Relate/Unrelate organizations to tokens (After these changing you should see the changes of token list in donate page of projects)
- [x] Add new tokens
- [x] Edit existing tokens
- [x] Only admin can edit tokens ( Operators  can't)

divine-comedian:
testing on `https://serve.giveth.io/admin`

I am on the `Token` tab I try to add a new token - I enter all the information and hit save and this error appears: 

![image](https://user-images.githubusercontent.com/67759413/163633580-c1bf3026-7293-420f-a211-1269522f6323.png)

however the token appears to be successfully added to the list 
![image](https://user-images.githubusercontent.com/67759413/163633650-da04e693-cb13-42c4-8791-39122d32edc7.png)

and it shows up on the donate page and I can successfully donate to a project with it - 
![image](https://user-images.githubusercontent.com/67759413/163634187-fd43a293-de3a-421b-9f7a-ac32e95a32bb.png)
the donation shows up appropriately on the donations tab of the project's page

CarlosQ96:
Funny it says, Error 200, when 200 is status OK

CarlosQ96:
I'll take a look

divine-comedian:
More Testing - I can successfully edit the info of a token. changing the name and changing which platform it can be used on. when i remove it from giveth it no longer shows on the donate page. however when I add it to Trace I can not find the token anywhere on the list to donate - however we can see there are many tokens eligible in trace to be donated 

![image](https://user-images.githubusercontent.com/67759413/163635141-7b375769-780f-4b9b-980c-918334f962b5.png)

![image](https://user-images.githubusercontent.com/67759413/163635298-1d11db3f-fe6e-4948-9488-2a52b77f3670.png)

also on the first photo above you can see I have two tokens with separate names but they have the same token address - this shouldn't be allowed.

divine-comedian:
I only have access to an admin account - do we have a dummy operator account I can test with @CarlosQ96 ?

CarlosQ96:
I don't but you can create it in the user admin page if you are an admin

CarlosQ96:
[...]

PR #467, opened 2022-04-07
Add token page to adminbro

CarlosQ96:
related to #463


src/entities/organization.ts
```diff
@@ -43,7 +43,6 @@ export class Organization extends BaseEntity {
   @Field(type => [Token], { nullable: true })
   @ManyToMany(type => Token, token => token.organizations, {
     nullable: true,
-    eager: true,
```
CarlosQ96, line 46: Let me know if this is required, Mohammad, I needed it for the admin bro on the token side, but typeorm wont let you have it in both side.

src/server/adminBro.ts
```diff
@@ -405,6 +408,104 @@ const getAdminBroInstance = () => {
...
+            organizations: {
+              isVisible: {
+                show: true,
+                edit: true,
+                new: true,
+                list: true,
+              },
+              components: {
+                show: AdminBro.bundle('./components/ListOrganizationsNames'),
+                list: AdminBro.bundle('./components/ListOrganizationsNames'),
```
CarlosQ96, line 447: Adminbro doesn't render many to many, so I had to create a custom component.

src/entities/organization.ts
mohammadranjbarz: No carlos, it's not required, currently we dont need `eager` in here

test/pre-test-scripts.ts
```diff
@@ -96,7 +96,7 @@ async function seedOrganizations() {
 }
 
 async function relateOrganizationsToTokens() {
-  const tokens = await Token.find({});
+  const tokens = await Token.createQueryBuilder('token').getMany();
```
CarlosQ96, line 99: Since I added the eager:true to the entity. This was not allowing the tokens to save with the organization because of the null loaded org.
We have to be careful when using find() methods, as they are the only ones affected by cascade: true, eager: true and other similar flags. 
createQueryBuilder does not call those methods. They are useful but might damage our tests.

CarlosQ96: Just saying it to clarify, because in most/all of our mutations/queries we use always queryBuilder.
And I think the queryBuilder is 100x more powerful than the entity find({}) methods, but we can use them being careful.

CarlosQ96:
image
Added organizations to the table and a dropdown of organizations in the creation.

src/server/adminBro.ts
```diff
@@ -405,6 +408,105 @@ const getAdminBroInstance = () => {
...
+              },
+              availableValues: [
```
CarlosQ96: Honestly found no other way to do it.
Tried the { isMulti: true } but adminbro ignores the array of values.

test/pre-test-scripts.ts
mohammadranjbarz: Good point, I didn't know about the performance difference, thanks for mentioning that, I love to hear these type of advice

mohammadranjbarz: BTW I prefer using raw queries because it's easier to copy those queries and test them in Db viewer applications (like pg admin), and its easier to find solutions in Postgres query instead of finding solutions in TypeOrm

src/server/adminBro.ts
mohammadranjbarz: Good job man

src/server/components/ListOrganizationsNames.jsx
```diff
@@ -0,0 +1,23 @@
...
+const ListOrganizationsNames = (props) => {
+  const token = props.record.params;
+
+  // Filter organizations from params format and delete null values
+  const organizations = [
+    token['organizations.0.name'],
+    token['organizations.1.name'],
+    token['organizations.2.name'],
+    token['organizations.3.name'],
+  ].filter(o => o);
```
mohammadranjbarz: Can you make this part more dynamic?
now if we add more organizations we should change these part, for instance I added an organization in my DB and it didn't show that to me

Screen Shot 1401-01-22 at 13 31 05
Screen Shot 1401-01-22 at 13 31 16

mohammadranjbarz (changes requested):
Thanks @CarlosQ96 
It was a great job that you did, it will save lots of backend team members and give the power and flexibility to admins, thank you

src/server/components/ListOrganizationsNames.jsx
CarlosQ96: I'll try to make up a solution for the component, but for the form I couldn't pass a async list (or fetched from the db) to the adminbro avaliable values. 
Ill let you know soon.

CarlosQ96: Ok I think I got it for the All Organizations option and the component. 
Missing the permutations list.

src/server/adminBro.ts
mohammadranjbarz: I had this problem before

mohammadranjbarz (approved):
I tested  it and it's working for me

Screen Shot 1401-01-24 at 10 48 03

Thank you so much, your test cases were great 
LGTM
Mucho gracias

Issue #445, opened 2022-03-26
Issue add uniqueness to emails


CarlosQ96:
This is causing issues in the adminbro login, multiple people with the same email. OR we move the adminbro login to another entity

CarlosQ96:
This is serious as we are having duplicated users with different wallets but same email, this can also conflict many other log ins.

@MoeNick

CarlosQ96:
This could be the reason people can't see their projects.
They log in, but the db matches a duplicate without their project, the UI says they are logged in with correct address/email but in reality they aren't. So they can't see their project. I remember there was 1 report of this issue before.

CarlosQ96:
Maybe adding uniqueness RIGHT AWAY isn't the solution. But moving the adminUser to another entity would reduce login issues.

MoeNick:
I didn't get the point, sorry
we added uniqueness of email for admin bro login?

CarlosQ96:
No. Look at the image:
![image](https://user-images.githubusercontent.com/92376054/160285300-3fdf85b5-f3ce-4ea8-9254-e7c33850d0f9.png)

CarlosQ96:
So while testing with Ashley some other stuff. We found she had in prod 6 accounts. 1 of them is the admin account too.

So when trying to log in by email in the adminbro. The db was selecting the wrong user (that doesn't have password) and returning error that it didn't match her password.

CarlosQ96:
Admins from adminbro use the same entity as the normal Users, so they can overlap, we should separate them.

CarlosQ96:
And as we can see, emails are not unique. They can be repeated.

MoeNick:
Oh thanks @CarlosQ96 I got it.

MoeNick:
Thank for making it clear @CarlosQ96, another solution might be adding roles to the user table, so only admins are considered as AdminBro users. 
We may need this for further multi-level account feature in the future.

PR #468, opened 2022-04-07
Fix login to admin bro when multiple user have same email address

mohammadranjbarz:
related to #445


CarlosQ96 (approved):
LGTM! did some manual testing and tests look good!
Thanks Mohammad.

Issue #470, opened 2022-04-09
Refactor DB queries ( by moving them from resolvers and reuse them)


mohammadranjbarz:
We use queries directly in our resolvers , and it makes our code messy, I suggest we should move the queries to a separated file like https://github.com/Giveth/impact-graph/blob/staging/src/repositories/userRepository.ts in this way we can write unit tests for every function(query) and we can reuse them ( currently we have written some queries in several resolvers and it makes it hard to maintain), besides doing this will make it easier to change our ORM/DB in the future , and it's good practice to encapsulate DB queries 

PS: I thinks it's not responsibility of resolvers to call DB queries
https://en.wikipedia.org/wiki/Single-responsibility_principle

- [ ] Move queries from categoryResolver
- [ ] Move queries from donationResolver
- [ ] Move queries from projectResover
- [ ] Move queries from reactionResolver
- [ ] Move queries from statusReasonResolver
- [ ] Move queries from userResolver

Please pay attention to these items during doing these stuffs :
1. Write some sucess/fail scenario test cases for every function that you write for queries
2. Check https://github.com/Giveth/giveth-dapps-v2/tree/develop/src/apollo/gql to see frontend uses what fields in response, maybe some `joins` are unnecessary ( @MohammadPCh  All the query/mutations that you use are in here or we should check somewhere else as well?)
3. when writing test cases, for function that use join , be aware to check wether join works correctly or not

@CarlosQ96  @aminlatifi  Correct me if you have another opinion about what I said

divine-comedian:
@mohammadranjbarz  is this issue still relevant?

mohammadranjbarz:
> @mohammadranjbarz is this issue still relevant?

We can close it

PR #473, opened 2022-04-11
F 470 refactor db queries status reasons resolver

FatemeGhasemi:
Relate #470

### DIFF
--- a/src/repositories/statusReasonRepository.test.ts
+++ b/src/repositories/statusReasonRepository.test.ts
@@ -1,5 +1,4 @@
 import { assert } from 'chai';
-import { ProjectStatus } from '../entities/projectStatus';
 import { ProjectStatusReason } from '../entities/projectStatusReason';
 import {
   findAllStatusReasons,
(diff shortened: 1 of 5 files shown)


FatemeGhasemi:
Hey @CarlosQ96 , can you review this? ( Because I dont have access to repository, I cant set reviewer for my PRs)

CarlosQ96:
@FatemeGhasemi Sure, ill review it soon!

src/repositories/statusReasonRepository.ts
```diff
@@ -0,0 +1,18 @@
...
+    .leftJoinAndSelect('project_status_reason.status', 'status')
+    .getMany();
+};
+
+export const findStatusReasonsByStatusId = async (
+  statusId: number,
+): Promise<ProjectStatusReason[]> => {
+  return ProjectStatusReason.createQueryBuilder('project_status_reason')
+    .leftJoinAndSelect('project_status_reason.status', 'status')
+    .where(`"statusId" = ${statusId}`)
```
CarlosQ96: Small change here, now that we are refactoring. Interporlation like this prone to SQL injection, maybe not in this case. But the best practices for ORMs and queries, is letting the ORM interpolate and escape the characters, preventing sql injection.
We are switching any interpolation to this format:
```javascript
.where('project_status_reason."statusId" = :statusId', { statusId: statusId })
```

src/repositories/statusReasonRepository.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const status = await ProjectStatus.findOne({
+      id: ProjStatus.active,
+    });
+    await ProjectStatusReason.create({
+      status,
+      description: 'test',
+    }).save();
+    const allStatusReasons = await findAllStatusReasons();
+    assert.isOk(allStatusReasons);
+    assert.notEqual(allStatusReasons.length, 0);
```
CarlosQ96: Instead of notEqual, to make the test direct and clear you should do isAbove.
```javascript
assert.isAbove(allStatusReasons.length, 0)
```

CarlosQ96 (changes requested):
Good job! LGTM.
Just two small comments since we are trying to improve the code base as much as posible following best practices. After those two small changes, ill merge!

src/repositories/statusReasonRepository.test.ts
CarlosQ96: There is huge list of assertions we can use in test with many functionalities: https://www.chaijs.com/api/assert/

src/repositories/statusReasonRepository.ts
FatemeGhasemi: Thanks @CarlosQ96  for your review, I didn't know it could be vulnerable, good point.
Done.

src/repositories/statusReasonRepository.test.ts
FatemeGhasemi: You are right, I found better assertion in the link that you sent, and changed to 
    `assert.isNotEmpty(allStatusReasons);`
instead of put assertion on length of arra

CarlosQ96 (approved):
Thanks LGTM!

PR #474, opened 2022-04-11
F 470 refactor db queries user resolver

FatemeGhasemi:
Relate #470

### DIFF
--- a/src/repositories/userRepository.ts
+++ b/src/repositories/userRepository.ts
@@ -9,3 +9,11 @@ export const findAdminUserByEmail = async (email): Promise<User | null> => {
           `);
   return users.length > 0 ? users[0] : null;
 };
+
+export const findUserByWalletAddress = (walletAddress: string) => {
+  return User.findOne({ walletAddress });
+};
+
+export const findUserById = (userId: number) => {
+  return User.findOne({ id: userId });
+};
--- a/src/repositories/userRepository.ts
+++ b/src/repositories/userRepository.ts
@@ -1,7 +1,7 @@
 import { User, UserRole } from '../entities/user';
 
 export const findAdminUserByEmail = async (
-  email,
+  email: string,
 ): Promise<User | undefined> => {
   const users = await User.query(`
             SELECT * FROM public."user"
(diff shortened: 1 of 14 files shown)


FatemeGhasemi:
@mohammadranjbarz  Can you review it?

src/resolvers/userResolver.ts
```diff
@@ -46,7 +52,8 @@ export class UserResolver {
 
   @Query(returns => User, { nullable: true })
   userByAddress(@Arg('address', type => String) address: string) {
-    return this.userRepository.findOne({ walletAddress: address });
+    // return this.userRepository.findOne({ walletAddress: address });
```
mohammadranjbarz: Please remove useless comments

src/resolvers/userResolver.ts
```diff
@@ -60,7 +67,8 @@ export class UserResolver {
     @Ctx() { req: { user } }: MyContext,
   ): Promise<boolean> {
     if (!user) throw new Error(errorMessages.AUTHENTICATION_REQUIRED);
-    const dbUser = await User.findOne({ id: user.userId });
+    // const dbUser = await User.findOne({ id: user.userId });
```
mohammadranjbarz: Please remove useless comments

mohammadranjbarz (changes requested):
Thanks @FatemeGhasemi 
Very good, I suggest in other resolvers that call user queries, it's better to change them to use your userRepository instead of calling queries directly, for instance this line
https://github.com/Giveth/impact-graph/blob/29ee74f68a15f55b4e45ef3aaffaf4970bed1f07/src/resolvers/projectResolver.ts#L697

src/repositories/userRepository.ts
```diff
@@ -1,6 +1,8 @@
 import { User, UserRole } from '../entities/user';
 
-export const findAdminUserByEmail = async (email): Promise<User | null> => {
+export const findAdminUserByEmail = async (
+  email,
```
mohammadranjbarz: specify type of `email` in input parameters

FatemeGhasemi: done!

mohammadranjbarz (approved):
@FatemeGhasemi Thanks, it seems good to me

Issue #289, opened 2022-01-06
Revise checking amount of donation when verifying


mohammadranjbarz:
There are some donations in our db that the amount of donation is a little different with real amount like https://blockscout.com/xdai/mainnet/tx/0xfe2c49f7462f13d07731285f6c39cede54c5859740d18988284dbfab046efc17 with `300.164094` amount but in our DB the amount is `300.1641` and we assume the amount is wrong, so the donation wound get verified, we should change the logic of comparison of amount  in verifying donations

CarlosQ96:
Got it

mohammadranjbarz:
<img width="1041" alt="Screen Shot 1401-01-30 at 09 49 35" src="https://user-images.githubusercontent.com/9850545/163925294-45e8ab39-8bc0-4ac5-a7ef-d45cd41cd0ee.png">

I deleted the poignArt donations and the new created donations have correct `priceUsd` and `valueUsd`
@MoeNick  so we can move it to done

MoeNick:
@jeevanplus007 
pls take a look. 
@mohammadranjbarz would u pls provie some test scenarios?

mohammadranjbarz:
@MoeNick  @jeevanplus007 
You should donate an amount that have 6 digits after `.` like `1.1223456` and expected behaviour is that donation should get verified ( succeed) before this PR they were `pending`

jeevanplus007:
**Test Snapshot :**  Tested with 6 digits after Decimal point. 
![image](https://user-images.githubusercontent.com/107279959/178103994-40e8b24b-d7c0-4cf5-9eca-8b81d6f45435.png)

PR #488, opened 2022-04-14
f_289 Ignore small difference in transaction and donation amounts

mohammadranjbarz:
related to #289


mohammadranjbarz:
@CarlosQ96 After merging this PR please let me know to delete existing poignArt donations from db

CarlosQ96 (approved):
LGTM! So we ignore small differences only as they throw error

PR #489, opened 2022-04-14
Add website to exportCSV data

mohammadranjbarz:
This came from Popcorn guys:

> Is there any way you would be able to add the projects websites to this sheet? We need to import it into our CRM and we can't do that without the sites (and there wouldn't be much point).

### DIFF
--- a/src/server/adminBro.ts
+++ b/src/server/adminBro.ts
@@ -978,6 +978,7 @@ const sendProjectsToGoogleSheet = async (
       listed: Boolean(project.listed),
       totalDonations: project.totalDonations,
       totalProjectUpdates: project.totalProjectUpdates,
+      website: project.website || '',
     };
   });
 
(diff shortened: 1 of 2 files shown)


src/server/adminBro.ts
```diff
@@ -978,6 +978,7 @@ const sendProjectsToGoogleSheet = async (
       listed: Boolean(project.listed),
       totalDonations: project.totalDonations,
       totalProjectUpdates: project.totalProjectUpdates,
+      website: project.website || '',
```
CarlosQ96, line 982: Add to the headers and the params of the addSheetWithRows

CarlosQ96 (changes requested):
After that you can merge

src/server/adminBro.ts
mohammadranjbarz: Thanks for mentioning, Done

Issue #485, opened 2022-04-11
priceUsd of imported poignArt donations is invalid


mohammadranjbarz:
I figured it out that donaiton has invalid price

query:
```
select id,"priceUsd",amount,currency, "priceEth", "valueEth" from donation
where "donationType"='poignArt'
```

* Produciton Donation
![Screen Shot 1401-01-22 at 18.21.18.png](https://images.zenhubusercontent.com/5fa7ed7fb0994323d1a66034/a7d906ae-d77d-43a9-a268-44c39ab55f65)

* Staging donation
![Screen Shot 1401-01-22 at 18.24.37.png](https://images.zenhubusercontent.com/5fa7ed7fb0994323d1a66034/93af69f1-cdb5-4616-93d9-d951cbeaca3c)

maryjaf:
Can we move this issue to Done? @mohammadranjbarz

mohammadranjbarz:
@maryjaf Yeah it's fixed now

PR #490, opened 2022-04-17
Fix valueUsd for poignArt donations

mohammadranjbarz:
related to #485


CarlosQ96:
I missed this completely when finishing the task. lgtm to merge too.

CarlosQ96:
I missed this completely when finishing the task. lgtm to merge too.

PR #506, opened 2022-05-04
Fix userId not present in anonymous donations

CarlosQ96:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/638

### DIFF
--- a/migration/1651611664220-fixUserIdForDonations.ts
+++ b/migration/1651611664220-fixUserIdForDonations.ts
@@ -1,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class fixUserIdForDonations1651611664220 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const donationTableExists = await queryRunner.hasTable('donation');
+
+    if (donationTableExists) {
+      // only for anonymous donations
+      await queryRunner.query(`
+            UPDATE "donation"
+            SET "userId" = u.id
+            FROM (
+                SELECT id, lower("walletAddress") as "userAddress"
+                FROM "user"
+            ) AS u
+            WHERE u."userAddress" = lower("fromWalletAddress")
+            AND "anonymous" = true
+        `);
+    }
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/src/resolvers/donationResolver.test.ts
+++ b/src/resolvers/donationResolver.test.ts
@@ -10,6 +10,7 @@ import {
   generateRandomEtheriumAddress,
   saveDonationDirectlyToDb,
   createDonationData,
+  generateUserIdLessAccessToken,
 } from '../../test/testUtils';
 import axios from 'axios';
 import { errorMessages } from '../utils/errorMessages';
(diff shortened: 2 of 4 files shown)


migration/1651611664220-fixUserIdForDonations.ts
```diff
@@ -0,0 +1,23 @@
...
+      // only for anonymous donations
+      await queryRunner.query(`
+            UPDATE "donation"
+            SET "userId" = u.id
+            FROM (
+                SELECT id, lower("walletAddress") as "userAddress"
+                FROM "user"
+            ) AS u
+            WHERE u."userAddress" = lower("fromWalletAddress")
+            AND "anonymous" = true
```
CarlosQ96, line 17: Non anonymous donations are fine, I checked in the db directly.

src/resolvers/donationResolver.test.ts
```diff
@@ -783,7 +784,12 @@ function saveDonationTestCases() {
     });
     assert.isTrue(donation?.isTokenEligibleForGivback);
     assert.isOk(donation);
-    assert.isNull(donation?.userId);
+    const user = await User.findOne({
+      walletAddress: SEED_DATA.FIRST_USER.walletAddress,
+    });
+    // there will always be an userId present
+    // login resolvers always create an user even if profile not completed
```
CarlosQ96, line 791: When we login with wallet it will always have a user id related, if not it will create it. (LoginWallet Mutation). LoginResolver.

I don't think I missed any instance the user is not created. But,
Correct me if I am wrong here @mohammadranjbarz , if you find any instance of user not created when logged in. I think it always goes through the loginWallet mutation.

src/resolvers/donationResolver.ts
```diff
@@ -399,11 +402,9 @@ export class DonationResolver {
...
       }
 
-      if (userId) {
-        donorUser = await findUserById(ctx.req.user.userId);
-      } else {
-        donorUser = null;
-      }
+      donorUser =
+        (await findUserById(ctx.req.user?.userId)) ||
+        (await findUserByWalletAddress(fromAddress.toString()));
```
CarlosQ96, line 407: Login resolver always has a user present or creates it.

migration/1651611664220-fixUserIdForDonations.ts
mohammadranjbarz: Nice

mohammadranjbarz (approved):
Thanks @CarlosQ96  LGTM
I ran migration locally and it filled `userId` for my donations
But we should be aware in `saveDonation` mutation `accessToken` is not necessary so users can create donation with any `fromWalletAddress` without being logged-in but as this webservice needs lots of changes through 
https://app.zenhub.com/workspaces/all-devs-61644589b92edf0016f6999c/issues/giveth/giveth-dapps-v2/629

and we might need to deprecate this endpoint and implement new one ( this mutation is already very messy)
So TL;DR IMO you can merge this now

CarlosQ96:
> Thanks @CarlosQ96 LGTM I ran migration locally and it filled `userId` for my donations But we should be aware in `saveDonation` mutation `accessToken` is not necessary so users can create donation with any `fromWalletAddress` without being logged-in but as this webservice needs lots of changes through https://app.zenhub.com/workspaces/all-devs-61644589b92edf0016f6999c/issues/giveth/giveth-dapps-v2/629
> 
> and we might need to deprecate this endpoint and implement new one ( this mutation is already very messy) So TL;DR IMO you can merge this now

Yeah I agree, next is improving the auth with the microservice at the very least.

PR #508, opened 2022-05-12
hotfix Fill value usd for old donations that dont have price already

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#667


mohammadranjbarz:
@CarlosQ96 Please Dont merge it, just approve it if it was ok, I will merge it myself

These PR do two things:
1. Change existing job for filling `valueUsd` stable coin donations 
2. Write migration for filling valuUsd For Other donations in https://github.com/Giveth/giveth-dapps-v2/issues/667#issuecomment-1118470909

src/services/donationService.ts
```diff
@@ -132,3 +134,68 @@ export const isTokenAcceptableForProject = async (inputData: {
...
+    'updateOldStableCoinDonationPrice donations count',
+    donations.length,
+  );
+  for (const donation of donations) {
+    logger.debug(
+      'updateOldStableCoinDonationPrice() updating accurate price, donationId',
+      donation.id,
+    );
+    try {
+      donation.priceUsd = 1;
```
CarlosQ96, line 193: In this case we are setting values to 1, and ignoring small differences?, for instance if in that moment the value was 0.98?

CarlosQ96 (approved):
Code seems good! Great job. Just that question I have. But I assume there is no easy way to get the stable coins historic amount, so we can take a negligible amount difference, in the end the tx in the blockchain is what matter.

CarlosQ96:
Ill be making a PR to this PR with the adminbro changes.

CarlosQ96:
You can merge! 

I linked this issue! https://github.com/Giveth/impact-graph/issues/509 

for later!!! I tackle it when I have more bandwidth, the effort is quite a lot after trying some implementation. You know the adminbro how limited is when it comes to customization 😮‍💨

src/services/donationService.ts
mohammadranjbarz: You know as you  said it's hard to get accurate price, and it's not necessary to do that, in trace.giveth.io we assume every stable coin is 1$ as well ( I didn't include **UST** here :))) )

Issue #515, opened 2022-05-16
Use Authentication Service to Verify JWT


aminlatifi:
After having the authentication service up, impact-graph can use this service to verify user JWT

aminlatifi:
https://github.com/Giveth/impact-graph/issues/514
https://github.com/Giveth/giveth-dapps-v2/issues/701

Issue #514, opened 2022-05-16
Setting up Authentication Service


aminlatifi:
New Authentication Service is implemented by @CarlosQ96 
Now it's time to:

- [x]  Deploy Service on kubernetese https://github.com/Giveth/SiweAuthMicroservice/issues/1
- [x] Define a separate route and API for authentication (like giveth.io/v1/auth) on Gateway
- [x] Develop frontend to use the new service https://github.com/Giveth/giveth-dapps-v2/issues/701
- [x] Develop impact-graph to use the new Authentication Service to verify issues JWT #515

divine-comedian:
@aminlatifi  is this epic still relevant or can it be closed?

PR #518, opened 2022-05-19
Feature: Implement auth microservice authorization

CarlosQ96:
Related to https://github.com/Giveth/impact-graph/issues/515

And the tasks Amin created https://github.com/Giveth/impact-graph/issues/514

The test wallets from mainnet are taken from the @ethersproject/testcases  -> accounts.json.gz file
which contains tons and tons of testing wallets.


src/server/bootstrap.ts
```diff
@@ -80,55 +81,27 @@ export async function bootstrap() {
...
+            user = await authorizationHandler(authVersion, token);
             req.user = user;
           }
-
-          const userWalletAddress = headers['wallet-address'];
-          if (userWalletAddress) {
-            req.userwalletAddress = userWalletAddress;
-          }
         } catch (error) {
-          // console.error(
```
CarlosQ96, line 121: Unused comment.

src/server/bootstrap.ts
```diff
@@ -80,55 +81,27 @@ export async function bootstrap() {
...
           if (headers.authorization) {
             token = headers.authorization.split(' ')[1].toString();
-            const secret = config.get('JWT_SECRET') as string;
-
-            const decodedJwt: any = jwt.verify(token, secret);
-
             let user;
-            if (decodedJwt.nextAuth) {
-              user = {
-                email: decodedJwt?.nextauth?.user?.email,
```
CarlosQ96, line 100: removed this nextauth, honestly we do not use any user object without ID. I don't know what the nextauth is.

src/services/authorizationServices.ts
```diff
@@ -29,3 +36,80 @@ export const getLoggedInUser = async (ctx: MyContext) => {
...
+      user = await validateImpactGraphJwt(token);
+      break;
+    case '2':
+      user = await validateAuthMicroserviceJwt(token);
+      break;
+    default:
+      throw new Error(errorMessages.INVALID_AUTHORIZATION_VERSION);
+  }
+  return user;
+};
```
CarlosQ96, line 63: So to handle backward compatibility I added versions. So version 1 will be our current one and the version 2 is the auth microservice. After we fully implement both I can remove the first one. 
This is to not break anything Ramin or Cherik might be doing. @mateodaza suggested.

src/repositories/userRepository.ts
```diff
@@ -22,3 +22,13 @@ export const findUserByWalletAddress = async (
 export const findUserById = (userId: number): Promise<User | undefined> => {
   return User.findOne({ id: userId });
 };
+
+export const createUserWithPublicAddress = async (
```
aminlatifi, line 29: This function should include `analytics.identifyUser` too

package.json
```diff
@@ -41,7 +42,7 @@
     "cors": "^2.8.5",
     "dotenv": "^8.2.0",
     "eth-sig-util": "^3.0.1",
-    "ethers": "^5.5.4",
+    "ethers": "5.5.4",
```
aminlatifi: Is this change by intention?

aminlatifi: Is this change by intention?

src/services/authorizationService.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      lastName: 'lastName',
+      email: `${new Date().getTime()}-giveth@giveth.com`,
+      url: 'website url',
+      loginType: 'wallet',
+      walletAddress: publicKey,
+    };
+    const user = await User.create(userData).save();
+    const nonceRoute = config.get('AUTH_MICROSERVICE_NONCE_URL') as string;
+    const nonceResult = await Axios.get(nonceRoute);
+    const wallet = new ethers.Wallet(privateKey, provider);
```
aminlatifi: AFAIK, provider is optional and is not needed here.
```suggestion
    const wallet = new ethers.Wallet(privateKey);
```

aminlatifi: AFAIK, the second argument (provider) is optional and is not needed here.
```suggestion
    const wallet = new ethers.Wallet(privateKey);
```

src/services/authorizationService.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const accessToken = authenticationResult.data.jwt;
+    const jwtUser = await authorizationHandler('2', accessToken);
+    assert.equal(jwtUser.userId, user.id);
+  });
+  it('should decode jwt and create user if it is nonexistent', async () => {
+    const privateKey = process.env.PRIVATE_ETHERS_SECONDARY_TEST_KEY as string;
+    const publicKey = process.env.PUBLIC_ETHERS_SECONDARY_TEST_KEY as string;
+    const nonceRoute = config.get('AUTH_MICROSERVICE_NONCE_URL') as string;
+    const nonceResult = await Axios.get(nonceRoute);
+    const wallet = new ethers.Wallet(privateKey, provider);
```
aminlatifi: Like above
```suggestion
    const wallet = new ethers.Wallet(privateKey);
```

aminlatifi: Like above
```suggestion
    const wallet = new ethers.Wallet(privateKey);
```

aminlatifi (changes requested):
Greate @CarlosQ96 , just few little things to fix.

package.json
CarlosQ96: Reverting this, seems I had an error but I can't reproduce it anymore, maybe it was my node version.

CarlosQ96: Fixed.

PR #523, opened 2022-05-30
Handle failed transactions when verifying donations

mohammadranjbarz:
Giveth/giveth-dapps-v2#795

### DIFF
--- a/src/services/transactionService.ts
+++ b/src/services/transactionService.ts
@@ -176,8 +176,14 @@ async function getTransactionDetailForNormalTransfer(
   const transaction = await getNetworkWeb3(networkId).eth.getTransaction(
     txHash,
   );
+  const receipt = await getNetworkWeb3(networkId).eth.getTransactionReceipt(
+    txHash,
+  );
   if (!transaction) {
     return null;
+  }
+  if (!receipt?.status) {
+    throw new Error(errorMessages.TRANSACTION_STATUS_IS_FAILED_IN_NETWORK);
   }
   const block = await getNetworkWeb3(networkId).eth.getBlock(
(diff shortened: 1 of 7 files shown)


src/services/transactionService.ts
```diff
@@ -176,9 +176,15 @@ async function getTransactionDetailForNormalTransfer(
   const transaction = await getNetworkWeb3(networkId).eth.getTransaction(
     txHash,
   );
+  const receipt = await getNetworkWeb3(networkId).eth.getTransactionReceipt(
+    txHash,
+  );
```
mohammadranjbarz, line 181: @CarlosQ96 
```
getNetworkWeb3(networkId).eth.getTransaction(
    txHash,
  );
  ```
  return the tx info and it doesn't matter tx is failed or not, so we have to use
  ```
   getNetworkWeb3(networkId).eth.getTransactionReceipt(
    txHash,
  );
  ```
  for this purpose

File: src/services/transactionService.ts [hunk 2]
```diff
@@ -216,6 +224,9 @@ async function getTransactionDetailForTokenTransfer(
   if (!transaction) {
     return null;
   }
+  if (!receipt?.status) {
```
CarlosQ96, line 227: So does receipt status is missing or the whole receipt doesn't get fetched?

CarlosQ96 (approved):
Hey Mohammad in general all looks good to me. Only the question I left.

src/services/transactionService.ts
mohammadranjbarz: The `status` for failed transactions is `false` otherwise it's `true`

Issue #521, opened 2022-05-29
Should just return all field of users to the user himself  ( security issue)


mohammadranjbarz:
Someone else should not be able to get another person's email ( and maybe other fields)

mohammadranjbarz:
@MohammadPCh  @RamRamez  @mateodaza 

Does it make problem in frontend?

RamRamez:
@mohammadranjbarz Yeah, for the public profile view we need these information:

![image](https://user-images.githubusercontent.com/73441857/170857482-8b24a9bc-b59a-4e07-b634-dda5d8a5355f.png)

RamRamez:
@mohammadranjbarz If this is a security issue related to email address, we can talk with design team and remove it but we need the user name.
which end-point you mean? just `userByAddress` or all end-points related to user info?

mohammadranjbarz:
> @mohammadranjbarz If this is a security issue related to email address, we can talk with design team and remove it but we need the user name. which end-point you mean? just `userByAddress` or all end-points related to user info?

I think all end-points related to user info makes sense,  I think it's ok to have name there

maryjaf:
I tested this issue on staging and the email address of other user has not been displayed 
for example(https://staging.giveth.io/user/0x2a3522042d44278689a7a15aa2ef5fabed1a9c8e)

But when I select  my user(maryjaf), I don't see my email address. This is true? @mohammadranjbarz 
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/111529185/192085836-86376c3c-1e38-456b-967d-24ae41a46ece.png">

RamRamez:
> I tested this issue on staging and the email address of other user has not been displayed for example(https://staging.giveth.io/user/0x2a3522042d44278689a7a15aa2ef5fabed1a9c8e)
> 
> But when I select my user(maryjaf), I don't see my email address. This is true? @mohammadranjbarz <img alt="image" width="1440" src="https://user-images.githubusercontent.com/111529185/192085836-86376c3c-1e38-456b-967d-24ae41a46ece.png">

This page (`/user/[wallet]`) is public profile. In public profile email address shouldn't be shown (that is related to this issue).
In my account (`/account`) page, email (if associated with user), will be shown.

PR #525, opened 2022-05-30
Dont return all fields for users ( security issue)

mohammadranjbarz:
related to #521

### DIFF
--- a/src/entities/project.ts
+++ b/src/entities/project.ts
@@ -8,6 +8,7 @@ import {
   Column,
   Entity,
   Index,
+  JoinColumn,
   JoinTable,
   LessThan,
   ManyToMany,
@@ -201,6 +202,11 @@ class Project extends BaseEntity {
   @ManyToOne(type => ProjectStatus, { eager: true })
   status: ProjectStatus;
 
+  @Index()
+  @Field(type => User, { nullable: true })
+  @ManyToOne(() => User, { eager: true })
+  adminUser?: User;
+
   @Field(type => [ProjectStatusHistory], { nullable: true })
   @OneToMany(
     type => ProjectStatusHistory,
@@ -230,10 +236,6 @@ class Project extends BaseEntity {
   @Field(type => Boolean, { nullable: true })
   @Column({ type: 'boolean', default: null, nullable: true })
   listed?: boolean | null;
-
-  // Virtual attribute to subquery result into
-  @Field(type => User, { nullable: true })
-  adminUser?: User;
 
   // Virtual attribute to subquery result into
   @Field(type => Int, { nullable: true })
--- a/src/entities/project.ts
+++ b/src/entities/project.ts
@@ -202,10 +202,16 @@ class Project extends BaseEntity {
   @ManyToOne(type => ProjectStatus, { eager: true })
   status: ProjectStatus;
 
+  @RelationId((project: Project) => project.status)
+  statusId: number;
+
   @Index()
   @Field(type => User, { nullable: true })
   @ManyToOne(() => User, { eager: true })
   adminUser?: User;
+
+  @RelationId((project: Project) => project.adminUser)
+  adminUserId: number;
 
   @Field(type => [ProjectStatusHistory], { nullable: true })
   @OneToMany(
@@ -213,9 +219,6 @@ class Project extends BaseEntity {
     projectStatusHistory => projectStatusHistory.project,
   )
   statusHistory?: ProjectStatusHistory[];
-
-  @RelationId((project: Project) => project.status)
-  statusId: number;
 
   @Field(type => Float)
   @Column({ type: 'real' })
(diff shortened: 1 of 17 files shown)


mohammadranjbarz:
I have deleted some webservices that we don't use it ( I have checked it with @RamRamez )
* `project`
* `projectByAddress`
* `user`

mohammadranjbarz:
@CarlosQ96  I changed all resolvers to not return sensitive data of users, and I wrote the test cases, but I couldn't add selection for
```
 .innerJoinAndMapOne(
        'project.adminUser',
        User,
        'user',
        'user.id = CAST(project.admin AS INTEGER)',
      )
    ```  
    in some resolvers, as you have more typeorm experience than me, can you help me on that? ( if we fix that the test cases should pass)

**Review by CarlosQ96 — APPROVED** [review]
Hey Mohammad I reviewed your changes and tested and they are fine! I reused your tests for my changes.

Please review yourself the part I did, that no info is deleted and everything runs fine. Just to be sure. And then we can merge.

**Comment by CarlosQ96** [comment]
Code looks much cleaner and more optimal as we don't map data separately but the database and the database parser takes care directly.

TODO:

On another separate pr after we release to production. We have to remove the admin string column that is deprecated.
We cannot do it on this PR or same release to staging/prod, because the db sync will get in the way.

File: src/resolvers/projectResolver.ts [hunk 1]
```diff
@@ -814,10 +793,10 @@ export class ProjectResolver {
       totalProjectUpdates: 1,
       verified: false,
       giveBacks: false,
+      adminUser: user,
     });
 
     const newProject = await this.projectRepository.save(project);
-    newProject.adminUser = await findUserById(Number(newProject.admin));
```
CarlosQ96, line 820: No more virtual attribute, this is not required. Its already set on creation for adminUser in the createProject Mutation. (just clarifying)

src/entities/project.ts
```diff
@@ -201,16 +202,24 @@ class Project extends BaseEntity {
   @ManyToOne(type => ProjectStatus, { eager: true })
   status: ProjectStatus;
 
+  @RelationId((project: Project) => project.status)
+  statusId: number;
+
+  @Index()
+  @Field(type => User, { nullable: true })
+  @ManyToOne(() => User, { eager: true })
```
aminlatifi, line 210: @CarlosQ96 @mohammadranjbarz Why do we have `eager` true?

src/entities/user.ts
```diff
@@ -56,11 +64,9 @@ export class User extends BaseEntity {
   @Column({ unique: true })
   walletAddress?: string;
 
-  @Field({ nullable: true })
```
aminlatifi, line 59: @mohammadranjbarz @CarlosQ96 Why we don't remove `password` and `encryptedPassword` field completely? They remained from impact-graph legacy project and don't have any meaning in Giveth.

src/entities/project.ts
aminlatifi: The user maybe loaded with its sensitive data whenever the a project instance is queried!

src/resolvers/donationResolver.ts
```diff
@@ -175,12 +176,13 @@ export class DonationResolver {
...
-        fromWalletAddress: In(fromWalletAddressesArray),
-      },
-    });
-    return donations;
+    return this.donationRepository
+      .createQueryBuilder('donation')
+      .where({ fromWalletAddress: In(fromWalletAddressesArray) })
+      .leftJoin('donation.user', 'user')
+      .addSelect(publicSelectionFields)
+      .leftJoinAndSelect('donation.project', 'project')
```
aminlatifi, line 184: I am worried project admin sensitive data be loaded here because of that `{eagher: true}` option in src/entities/project.ts

src/entities/user.ts
```diff
@@ -12,9 +12,18 @@ import { Project, ProjStatus } from './project';
...
 
+export const publicSelectionFields = [
+  'user.id',
+  'user.walletAddress',
+  'user.name',
+  'user.firstName',
+  'user.lastName',
+  'user.url',
+  'user.avatar',
+];
```
aminlatifi, line 25: We can keep it as it is, but the user entity always must be mapped to `user` name in the query.

aminlatifi (changes requested):
Thanks @mohammadranjbarz  and @CarlosQ96 

Nothing special, just I am suspicious about `eager` value

src/entities/user.ts
mohammadranjbarz: We use `encryptedPassword` for admin users, and about password I thin if delete it now we should change admin bro create admin page to replace password field with another field

src/resolvers/donationResolver.ts
CarlosQ96: Hey Amin. I guess the ``` { eager: true } ``` is being used on our tests for easy loading. They are only loaded when using repository like ``` .find() ``` methods.

Since we use querybuilder on the mutations it won't load the extra data.

https://orkhan.gitbook.io/typeorm/docs/eager-and-lazy-relations

```
Eager relations only work when you use find* methods. If you use QueryBuilder eager relations are disabled and have to use leftJoinAndSelect to load the relation. Eager relations can only be used on one side of the relationship, using eager: true on both sides of relationship is disallowed.
```

PR #530, opened 2022-06-01
Restart Backend Automatically if impact-graph process failed

mhmdksh:
After testing, I noticed that some containers are not restarted automatically when anything wrong happens inside the process that the container is running.

I've added a simple fix to how the docker-compose should be run, hopefully this will fix the sudden downtime issue.


CarlosQ96 (approved):
Thanks man ill redo this for ApiGiv and auth microservice.

PR #531, opened 2022-06-02
F 703 support multi recipient

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#703


CarlosQ96 (approved):
Ignore this hahah

mohammadranjbarz:
@MoeNick  @CarlosQ96 @MohammadPCh 

After merging this PR ( we are waiting for current release after that we would review and merge this)
The frontend should change two some web services:
### **createProject** and **updateProject**
Instead of sending `walletAddress` you should send this in the body of request

```
 addresses: [
        {
          address: walletAddress,
          networkId: 1 (for mainnet) or 3( for ropsten) or 100 (for xDai),
        },
      ]
```
### Getting projects
In every service that frontend gets project list or project detail , it would get address infoes as well with adding this to graphql query

```
 addresses {
          address
          isRecipient
          networkId
        }
```
And then based on each network that user selected on metamask frontend should use the address of that network, and make sure that the `isRecipient` is `true` for that address

mohammadranjbarz:
@CarlosQ96 Please review it, but I keep it draft to not merge it until the frontend guys @MohammadPCh  or @MoeNick  tell us to merge it, because this PR changes are not backward compatible and staging may break if we merge it without implementing frontend side

CarlosQ96 (changes requested):
https://github.com/Giveth/impact-graph/pull/525#issuecomment-1141686937 I think you mentioned the projectByAddress mutation is no longer used. We should delete it as it would be incomplete with this Pr, it would need to search by related addresses too.

Same for project query. As we do not user repository to fetch project.

src/resolvers/projectResolver.ts
```diff
@@ -1056,47 +1099,15 @@ export class ProjectResolver {
...
-    const purpleAddress = await PurpleAddress.createQueryBuilder()
-      .where(`LOWER(address) = :walletAddress`, {
-        walletAddress: address.toLowerCase(),
-      })
-      .getOne();
-
-    if (purpleAddress) return true;
-
-    return false;
+    return isWalletAddressInPurpleList(address);
```
CarlosQ96, line 1106: Sql injection can be done here, escape the address on the raw query.

src/repositories/projectAddressRepository.ts
```diff
@@ -0,0 +1,91 @@
...
+export const isWalletAddressInPurpleList = async (
+  address: string,
+): Promise<boolean> => {
+  const projectAddress = await ProjectAddress.query(
+    `
+          SELECT "projectId", LOWER(address) as "projectAddress"
+          FROM project_address
+          JOIN project
+          on project.id="projectId" and "verified"=true
+          where address='${address}'
```
CarlosQ96: We have an sql injection problem here. Specially since the query ```walletAddressIsPurpleListed``` accepts the address directly. We have to escape it!!

CarlosQ96 (changes requested):

src/repositories/projectAddressRepository.ts
CarlosQ96: Did not find a clean way with raw queries... but maybe this can help.

https://github.com/typeorm/typeorm/issues/556#issuecomment-556042689

CarlosQ96:
Quick question, is the previous purple address list lost? gotta reintroduce them in the adminbro?

mohammadranjbarz:
> Quick question, is the previous purple address list lost? gotta reintroduce them in the adminbro?

Yeah we would lose them, but now there isn't any purple addresses in our DB and I told Ashley to not add any addresses until we merge this, so it would not make a problem

CarlosQ96:
I think we can skip deleting the other mutations for now. The sql injection is fixed.

src/resolvers/projectResolver.ts
```diff
@@ -797,6 +828,20 @@ export class ProjectResolver {
     });
 
     const newProject = await this.projectRepository.save(project);
+    const adminUser = (await findUserById(Number(newProject.admin))) as User;
+    newProject.adminUser = adminUser;
+    for (const relatedAddress of projectInput?.addresses) {
```
CarlosQ96: We have a N+1 queries issue here as we are saving entities separately. It's better to unify the jsons and insert once into the entity.

CarlosQ96 (changes requested):

src/resolvers/projectResolver.ts
mohammadranjbarz: @CarlosQ96 Thanks for mentioning that, I fixed it here
https://github.com/Giveth/impact-graph/pull/531/commits/6fc6f71cac105fa28293b135418a3b9e1ee63517

CarlosQ96 (approved):
Awesome job man! All issues fixed. I think we only have to coordinate with @mateodaza to merge. He is wanting to integrate.

src/repositories/projectAddressRepository.ts
CarlosQ96: Fixed.!

mohammadranjbarz: Thanks so much

PR #538, opened 2022-06-13
Feature verification from adminbro

CarlosQ96:


src/repositories/projectRepository.ts
```diff
@@ -17,6 +21,81 @@ export const findProjectBySlug = (
...
+        address: relatedAddress.address,
+        networkId: relatedAddress.networkId,
+        projectId: verificationForm.projectId,
+        userId: verificationForm.userId,
+        isPrimaryAddress: false,
+      }),
+    );
+  }
+  project.relatedAddresses = relatedAddresses;
+  return project.save();
```
mohammadranjbarz: @CarlosQ96  Why you dont copy the data to project, and you are just copying that in projectUpdate?

CarlosQ96: Hey Mohammad, I just looked at Milestone and the attributes you added them to the projectupdate. Should I add them too to the project? I feel like most info is related to a project update.

PR #539, opened 2022-06-15
Add index for createdAt field in donation entity

mohammadranjbarz:
As @mendesfabio mentioned
> I'm querying giveth graphql API
but it's taking too much time to return donations, even when i apply the filter fromDate and toDate
looking at the code i noticed we don't have an index on createdAt
could we add it there? it would probably help these kind of query

### DIFF
--- a/src/entities/donation.ts
+++ b/src/entities/donation.ts
@@ -135,6 +135,7 @@ export class Donation extends BaseEntity {
   @RelationId((donation: Donation) => donation.user)
   userId: number;
 
+  @Index()
   @Field(type => Date)
   @Column()
   createdAt: Date;


mendesfabio:
hey, thanks for this @mohammadranjbarz - isn't missing the migration file here?

CarlosQ96:
Hey @mendesfabio! yeah ideally a migration is required!! But the project before we joined the team,  was setup with typeorm Synchronize entities to true. This added some complexity, quite some work and the project entities was quite numerous already.
(like creating all migrations from scratch and not overwrite)

We have to do the migrate to using just-migrations, but priorities haven't allow it. 

As of now a migration is not required.

CarlosQ96 (approved):
Let's merge. 
I investigated a bit and typeorm doesn't support custom indexes. But I think this should be enough for now.

PR #542, opened 2022-06-22
Fix project social Profiles displaying in adminbro show view

CarlosQ96:
Simplified the show view with a custom function that modifies the response object. That way no weird db or columns need to be added.

If a social profile is removed it will be reflected in both the project and projectVerificationForm.


.adminbro/.entry.js
```diff
@@ -1,9 +1,7 @@
 AdminBro.UserComponents = {}
-import Component1 from '../src/server/components/SocialProfiles'
+import Component1 from '../src/server/components/ListOrganizationsNames'
 AdminBro.UserComponents.Component1 = Component1
 import Component2 from '../src/server/components/ListOrganizationsNames'
 AdminBro.UserComponents.Component2 = Component2
```
CarlosQ96, line 5: this was auto generated by adminbro.

mohammadranjbarz (approved):
Thanks @CarlosQ96 

Your approach was great and easy to understand ( better than changing DB columns,...)

PS: I just changed the text ( to include social username and verified status), and I added some test cases for your repository function

mohammadranjbarz:
@MoeNick 
Now we show social profiles in admin-bro, ( this was missed in weekly demo)
Screen Shot 1401-04-01 at 14 53 55

PR #543, opened 2022-06-23
Fix similar projects query for projects that doesnt have any category

mohammadranjbarz:
related #668

### DIFF
--- a/src/resolvers/projectResolver.test.ts
+++ b/src/resolvers/projectResolver.test.ts
@@ -3641,6 +3641,22 @@ function similarProjectsBySlugTestCases() {
     );
     assert.isUndefined(currentViewedProject);
   });
+  it('should not throw error if project doesnt have any category ', async () => {
+    const viewedProject = await saveProjectDirectlyToDb({
+      ...createProjectData(),
+      title: String(new Date().getTime()),
+      slug: String(new Date().getTime()),
+      categories: [],
+    });
+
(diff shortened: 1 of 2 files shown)


CarlosQ96 (approved):
LGTM thanks!

src/resolvers/projectResolver.test.ts
```diff
@@ -3641,6 +3641,22 @@ function similarProjectsBySlugTestCases() {
...
+      categories: [],
+    });
+
+    const result = await axios.post(graphqlUrl, {
+      query: fetchSimilarProjectsBySlugQuery,
+      variables: {
+        slug: viewedProject.slug,
+      },
+    });
+    assert.isArray(result.data.data.similarProjectsBySlug.projects);
```
CarlosQ96, line 3658: Actually test that there is at least 1 element.
Because if its empty it returns [] => which matches the array condition.

CarlosQ96 (changes requested):

src/resolvers/projectResolver.test.ts
mohammadranjbarz: You know I dont care about the result item here, we just need to get array, because before fixing the issue it was returning `null` for data and was throwing exception, so I just need to make sure it doesn't throw exception anymore and the result number is not important in this test case

CarlosQ96 (approved):
Ok thanks!

PR #562, opened 2022-07-13
F 655 add main categories

mohammadranjbarz:
related to Giveth/GIVeconomy#655

### DIFF
--- a/src/resolvers/projectResolver.ts
+++ b/src/resolvers/projectResolver.ts
@@ -182,6 +182,9 @@ class GetProjectsArgs {
   @Field({ nullable: true })
   category: string;
 
+  @Field({ nullable: true })
+  mainCategory: string;
+
   @Field(type => FilterBy, {
     nullable: true,
     defaultValue: { field: null, value: null },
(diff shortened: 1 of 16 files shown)


src/resolvers/projectResolver.ts
```diff
@@ -235,6 +238,24 @@ export class ProjectResolver {
...
+
+  static addMainCategoryQuery(
+    query: SelectQueryBuilder<Project>,
+    mainCategory: string,
+  ) {
+    if (!mainCategory) return query;
+
+    query = query
+      .leftJoinAndSelect('project.categories', 'categoryForMainCategorySearch')
+      .innerJoinAndSelect(
```
CarlosQ96: TypeOrm only serializes columns or virtual columns that exist in the entity, so this attributes categoryForMainCategorySearch  and filteredMainCategory, won't be selectable. However the ```JoinAndSelect``` is attempting to load the data. 
We should use ```leftJoin```and ```ìnnerJoin```to avoid extra data loading.

CarlosQ96: Test Pass with those too!

CarlosQ96 (changes requested):
Hey Mohammad, I think the Pr is good I only found a small optimization regarding left/inner joins.

src/resolvers/projectResolver.ts
mohammadranjbarz: Thanks @CarlosQ96 , I didn't know that, thanks for finding and fixing this

CarlosQ96 (approved):
Thanks Mohammad, great work!

Issue #565, opened 2022-07-19
Allow projects to have just one recipient address


mohammadranjbarz:
@RamRamez  said we need to have it

mohammadranjbarz:
@MoeNick  @RamRamez 
We deployed it on staging

divine-comedian:
how can this issue be tested?

RamRamez:
@divine-comedian you test multi recipient address after [this PR](https://github.com/Giveth/giveth-dapps-v2/issues/1135) is merged.

MoeNick:
Moving it back to code review due to the depency.

MoeNick:
The dependency merged, so it's now testable with the same scenario of #1150

PR #566, opened 2022-07-19
Allow projects to have one recipient address

mohammadranjbarz:
related to #565


CarlosQ96 (approved):
Thanks for explaining Mohammad! Looks good to me!

PR #574, opened 2022-07-27
Add notification-adapter interface

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#709


src/resolvers/projectResolver.ts
```diff
@@ -988,8 +976,19 @@ export class ProjectResolver {
 
     await pubSub.publish('NOTIFICATIONS', payload);
 
-    if (config.get('TRIGGER_BUILD_ON_NEW_PROJECT') === 'true')
+    if (config.get('TRIGGER_BUILD_ON_NEW_PROJECT') === 'true') {
+      // TODO whats this? I think we can delete it
```
CarlosQ96: Yes please delete this. This is netflify integration, when we had gatsby in the frontend, this triggered a full rerender of the app, super inneficient and we scrapped that after our rewrites of the frontend.

CarlosQ96 (approved):
LGTM Mohammad, awesome work. Just delete the netflify stuff as we are not using that.

We also have in the bootstrap.ts netflify routes I think that can be deleted too, we don't use that or have any idea of the service credentials. That doesn't even have tests and I haven't touched that old code.

src/resolvers/projectResolver.ts
mohammadranjbarz: Thanks @CarlosQ96  , I removed al the thing related to netlify

PR #575, opened 2022-07-27
Allow currencies to have number in their name

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#737


mohammadranjbarz:
@CarlosQ96 I have checked the logs and see the donations on drgiv3 is failing because it has number in it's name

CarlosQ96 (approved):
LGTM! Tested the cases and with numbers.

Issue #567, opened 2022-07-19
verification - show checked icon when email is verified


RamRamez:
When user clicks verify email, personal info get a checked badge. The checked badge should be shown when user has verified the email. We show checked badge based on last step so when user hasn't verified email yet, last step should be `before you start`

@mohammadranjbarz @CarlosQ96 

![image](https://user-images.githubusercontent.com/73441857/179787076-2a2c4978-6192-4ed5-b0a6-7ba660e68087.png)

mohammadranjbarz:
After sending each step data we marked that step as done and we change the last step but in personal info after filling the email and sending personal info data you need to verify the mail, so we should change flow of updating lastStep, it may affect the whole flow, so if we want to release these days I recommend don't change it and fix it for next release, because it needs lots of testing(maybe writing more automatic tests) to make sure doesn't break other things,
@MoeNick  what do you think?

MoeNick:
Let's have a talk about this in the daily @RamRamez @RamRamez @MohammadPCh

RamRamez:
@mohammadranjbarz The other problem that I have is the `managing funds` step doesn't have the checked badge. I think backend doesn't update last step for that.

![image](https://user-images.githubusercontent.com/73441857/179956253-d1cb6c1c-48a2-43ed-9cb2-b046904c419a.png)

mohammadranjbarz:
> @mohammadranjbarz The other problem that I have is the `managing funds` step doesn't have the checked badge. I think backend doesn't update last step for that.
> 
> ![image](https://user-images.githubusercontent.com/73441857/179956253-d1cb6c1c-48a2-43ed-9cb2-b046904c419a.png)

@RamRamez  Thanks for reporting this issue , I fixed it, it's deployed on staging ( the reason is that in the backend `managingFunds` step was before milestones, so updating `managingFunds` didn't update`lastStep`)

<img width="1857" alt="Screen Shot 1401-04-29 at 16 53 42" src="https://user-images.githubusercontent.com/9850545/179981083-89580c53-d37a-410b-a451-f60f81528913.png">

mohammadranjbarz:
@RamRamez  Now if user is not confirmed the email we should set `null` as their step so frontend will redirect it to first step ( personal info)
It's on staging now, you can test it

RamRamez:
@mohammadranjbarz after user confirmed the email, you should change last step, but now it doesn't change

[...]

PR #577, opened 2022-07-27
Dont update lastStep if email is not confirmed

mohammadranjbarz:
related to #567

### DIFF
--- a/src/services/projectVerificationFormService.ts
+++ b/src/services/projectVerificationFormService.ts
@@ -32,6 +32,30 @@ import { errorMessages } from '../utils/errorMessages';
 import { ProjectVerificationUpdateInput } from '../resolvers/types/ProjectVerificationUpdateInput';
 import { removeUndefinedFieldsFromObject } from '../utils/utils';
 
+const updateLastStep = async (params: {
+  projectVerificationForm: ProjectVerificationForm;
+  step: string;
+}): Promise<ProjectVerificationForm> => {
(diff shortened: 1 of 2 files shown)


src/services/projectVerificationFormService.ts
```diff
@@ -32,6 +32,30 @@ import { errorMessages } from '../utils/errorMessages';
 import { ProjectVerificationUpdateInput } from '../resolvers/types/ProjectVerificationUpdateInput';
 import { removeUndefinedFieldsFromObject } from '../utils/utils';
 
+const updateLastStep = async (params: {
+  projectVerificationForm: ProjectVerificationForm;
+  step: string;
+}): Promise<ProjectVerificationForm> => {
+  const { step, projectVerificationForm } = params;
+  if (
+    step === PROJECT_VERIFICATION_STEPS.PERSONAL_INFO &&
```
CarlosQ96, line 41: Can you skip steps or frontend validates? Maybe this validation of email_confirmed, should be present in future steps or in the submit step too.

CarlosQ96: Just in case, of some wacky url manipulation.

CarlosQ96: Thanks for explaning me by DM :D

mohammadranjbarz: Thank you for paying attention to details and security issues

Issue #578, opened 2022-07-27
Add multiple addresses to csv export


CarlosQ96:
Right now there is no way of knowing if the wallet address of the project is mainnet/xdai on the project. We gotta use the added relations when exporting.

CarlosQ96:
@WhyldWanderer The multiple addresses csv was merged! you can test!

PR #584, opened 2022-08-02
Add wallet addresses to export csv

CarlosQ96:
Related to https://github.com/Giveth/impact-graph/issues/578

### DIFF
--- a/src/server/adminBro.ts
+++ b/src/server/adminBro.ts
@@ -43,7 +43,7 @@ import {
 } from '../entities/projectStatusHistory';
 import { Organization, ORGANIZATION_LABELS } from '../entities/organization';
 import { Token } from '../entities/token';
-import { NETWORK_IDS } from '../provider';
+import { NETWORKS_IDS_TO_NAME, NETWORK_IDS } from '../provider';
 import { ProjectAddress } from '../entities/projectAddress';
 import {
   findAdminUserByEmail,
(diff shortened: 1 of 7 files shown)


src/server/adminBro.ts
```diff
@@ -1602,6 +1621,8 @@ const sendProjectsToGoogleSheet = async (
 
   // parse data and set headers
   const projectRows = projects.map((project: Project) => {
```
mohammadranjbarz, line 1622: I think we should set a filter for addresses that have `isRecipient:true` , because we may have more than two addresses ( a project can have multi related addresses but now just one or two of them are recipient)

mohammadranjbarz (changes requested):
Screen Shot 1401-05-11 at 17 16 49


I tested admin-bro and it works, and I think we would not have any problem on exporting on googlesheet, I just had one feedback other things are fine , thanks

src/server/adminBro.ts
CarlosQ96: Oh ok, I thought since there was a restriction You couldn't have more than 2.
Ill add the filter for isRecipient

CarlosQ96: Thanks Mohammad! I added the isRecipient filter in query and in the component.

mohammadranjbarz: Thanks Carlos, I tested it and it works,

mohammadranjbarz (approved):
Thanks @CarlosQ96  LGTM

Issue #597, opened 2022-08-18
Migrate test network from Kovan/Ropsten to Goerli on backend


mateodaza:
Specially for features like

- [x]  All Donating flow, from fetching prices and saving right amounts
- [x]  Token list given chainID: 5 (Goerli)
     Test tokens can be:
     goerliDAI: 0xdc31Ee1784292379Fbb2964b3B9C4124D8F89C60
     DRGIV3: 0xA2470F25bb8b53Bd3924C7AC0C68d32BF2aBd5be
- [x]  (Fill here if I miss something)

@CarlosQ96 @mohammadranjbarz

mohammadranjbarz:
Thanks @mateodaza for creating this issue and add information to this, I'm working on it

mohammadranjbarz:
@mateodaza  We merged the impact-graph on staging, can you check the frontend?

PR #599, opened 2022-08-22
Add Goerli network

mohammadranjbarz:
related to #597


mohammadranjbarz:
@mateodaza  have you done the frontend side?
is it safe to merge backend side? ( although I tried to make it backward compatible to still work with Ropsten network)

MoeNick:
I think the donation flow also should be considered.

mateodaza:
@mohammadranjbarz this is essentially for the donation flow. GIVeconomy network is already migrated - So I think it's safe to merge as most of the config comes from the backend

CarlosQ96 (approved):
Checked tests and migrations, everything seems ok! LGTM the code. 

I think we can skip giving blocks for tests/goerli as we really don't see much use from them, we are kind of wanting to unintegrate, so not worth the effort.

Issue #593, opened 2022-08-16
Boosting api - set user-project allocation


aminlatifi:
Set user-project boosting allocation percentage.

This API should accept an array of <projectId, percentages> tuples. So it should enable the user to set percentages of one or multiple projects(for dashboard use case) in one query

**Validations:**
Each user can allocate to a limited number of projects. Consider the limitation of 20 projects.
The allocation percentage precision is 0. (i.e. numbers like 12.1 are not accepted, only 12 or 13)

**Consistency:**
We should check the sum of the user allocation percentages is not more than 100%
We must update the user totalAllocatedGIVpower percentage after each update.
These can be implemented on DB layer by the user of triggers.

aminlatifi:
@mohammadranjbarz and I concluded
1. We will have an API for setting single project boosting. In this scenario, the user sets boosting for a single project and the rest of the positive boosting adapt their values as the sum of all user boosting values becomes 100%
2. Another API for setting multiple boosting values. The sum of these values must be 99% to 100%. A successful full call to this API causes all other boosting values to become zero and only data passed be persisted as user boosting.

maryjaf:
1-

Plz check this scenario @MohammadPCh 
Select the edit option
Change one of the percentages to 100 so that the total number of allocutions is above 100
The message is displayed correctly: You can’t exceed 100%
But when cancel or reset is selected, this message is still displayed
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/2121935b-2840-4063-9d12-cc830001086f)

maryjaf:
2-

![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/bb84f888-928f-4ca4-85ce-b452ec5e091c)
@mohammadranjbarz 
In the description, there is a limit of 20 boosted projects for the user, but when I select the 21st project, the appropriate error message is not displayed and the boosting operation remains loading.

maryjaf:
[...]

Issue #591, opened 2022-08-16
GIVpower boosting database schema


aminlatifi:
These values should be saved in DB
- givpower allocation percentage per user and project
- Accumulated user power in the last two weeks
- ~~Total percentage of givpower allocated by user~~
- Total givPower allocated to each project

Note: I have searched about Postgres triggers. I think we can use this feature to keep data consistency (not more than 100% allocation per user) and update user total allocation percentages

mohammadranjbarz:
* **Power boosting**

| name       | type    | required | validation     |
|------------|---------|----------|----------------|
| projectId  | integer | yes      |                |
| userId     | integer | yes      |                |
| percentage | integer | yes      | min=0, max-100 |

* **Power boosting snapshot**

| name       | type    | required | description                 |
|------------|---------|----------|-----------------------------|
| userId     | integer | yes      |                             |
| date       | Date    | yes      |                             |
| userPower  | integer | yes      | user power at specific time |

aminlatifi:
> 

By new design (givpower of a user is used 100% or 0%), it doesn't make sense to have this _Total percentage of givpower allocated by user_

MoeNick:
Ranking by the user service is still missing @mohammadranjbarz

aminlatifi:
> Ranking by the user service is still missing @mohammadranjbarz

@mohammadranjbarz @MoeNick We have a view to mapping tuple of userId and projectId to power (corresponding percentage multiplied by user average power in last round)
This view is of materialized one and has appropriate indexes for filtering and sorting.

aminlatifi:
The design of computing a project's givpower must be changed. Based on the new design, we should take snapshots of users' power balance, and the allocated percentages every 4 hours.
In order to implement project average allocated givpower we need to change the design and define new tables.

**Snapshot time and blocknumber**
| name       | type    | required | validation     | description |
|------------|---------|----------|----------------|----------------|
| id  | integer | yes      |                ||
| time     | Date | yes      |                |time of snapshot (every 4 hours). Will be filled by a job run by a cron job|
| blockNumber | integer |no     | | Will be filled later |
| roundNumber| integer| no|| The round snapshot belongs to. Can be filled later by app (not by db cron job)|

**User power balance at snapshot**
| name       | type    | required | validation     | description |
|------------|---------|----------|----------------|----------------|
|id| integer| yes|||
| userId| integer | true|||
|snapshotId| integer| true| | |
| balance| ineger| true| ||

[...]

PR #600, opened 2022-08-24
Feature power boosting

mohammadranjbarz:
related to #593 #591


migration/1662877385100-createUserTable.ts
```diff
@@ -0,0 +1,45 @@
...
+                  location character varying COLLATE pg_catalog."default",
+                  "loginType" character varying COLLATE pg_catalog."default" NOT NULL,
+                  "dId" character varying COLLATE pg_catalog."default",
+                  confirmed boolean NOT NULL DEFAULT false,
+                  "segmentIdentified" boolean NOT NULL DEFAULT false,
+                  "totalDonated" real DEFAULT 0,
+                  "totalReceived" real DEFAULT 0,
+                  CONSTRAINT "PK_cace4a159ff9f2512dd42373760" PRIMARY KEY (id),
+                  CONSTRAINT "UQ_efbd1135797e451d834bcf88cd2" UNIQUE ("walletAddress")
+              )
```
CarlosQ96, line 38: Last week MoeNick and Me were doing customer support and debugging an issue but we couldn't properly do it because we did not know when was the information of the user was added or when was he created.

I think we should add the createdAt and updatedAt columns here and on the entity.

CarlosQ96: Updated at can use this decorator in the entity for it to automatically update:   
```
 @UpdateDateColumn()
 updatedAt: Date;
```

I think it's important to always keep track of created/update dates.

mohammadranjbarz: Thanks @CarlosQ96  you are totally right I added 

```
  @UpdateDateColumn()
  updatedAt: Date;

  @CreateDateColumn()
  createdAt: Date;
```

to user entity

PR #603, opened 2022-08-28
F 334 test case for category resolver

FatemeGhasemi:


FatemeGhasemi:
@CarlosQ96 
Can you plz review this?

CarlosQ96:
Thanks Will review now!

src/resolvers/categoryResolver.test.ts
```diff
@@ -1,2 +1,67 @@
...
+      value: 'Agriculture',
+      isActive: true,
+    });
+    const categoryResponse = await axios.post(graphqlUrl, {
+      query: getCategoryData,
+      variables: {},
+    });
+    const result = categoryResponse?.data?.data?.categories;
+    assert.isNotEmpty(result);
+    categoryResponse.data.data.categories.forEach(item => {
```
CarlosQ96: I think you should remove this loop, because if another test adds a category without maincategory it will fail.
and just validate the mainCategory with the one you created.
```
    assert.equal(result[result.length - 1].mainCategory.title, mainCategory.title);
```

CarlosQ96 (changes requested):
Just this tiny change, and we can merge let me know when you do it! 
Thanks for the tests, great work!

src/resolvers/categoryResolver.test.ts
CarlosQ96: Also just to clarify, it's not that I think it's wrong, but it might be too general of test, checking if it has a relation or if it has a title present, because the title could be empty. (' ')
We better off encapsulating things to their own context so no side effects occur, and keeping things as explicit as posible.

FatemeGhasemi: You are right, it's better to encapsulate this, thanks for your feedback, I did it

CarlosQ96 (approved):
Thanks LGTM!

PR #605, opened 2022-08-30
Feat/givpower boosting api

aminlatifi:
- [x] Set single project boost
- [x] Set multiple project boost
- [x] Test set single project boost
- [x] Test set multiple  project boost
- [ ] ~~Add database triggers~~


aminlatifi:
@mohammadranjbarz I skipped the database trigger feature, it seems implementing this is not straightforward or even impossible. Because when we set multiple project boosts, multiple single update queries are run one by one and we cannot expect the database has a valid state after running each one individually.

mohammadranjbarz (approved):
Thanks @aminlatifi , your sollution for changing other projects boosting are a great, LGTM

PR #610, opened 2022-09-02
add userPower query for unsynced round

CarlosQ96:
Complement power boosting branch

Run npm run test:userPowerRepository


CarlosQ96:
Please review it when you can Mohammad! 
I added a test for validation and you can also confirm this by playing with the query.
```
      .where('userPowers.userId IS NULL')
```
if you remove this the join with return you those with the specified round.

mohammadranjbarz (approved):
Thanks @CarlosQ96 , LGTM

PR #611, opened 2022-09-02
Add revoke badge base cronjob

CarlosQ96:
WIP
Link to specs: https://docs.google.com/document/d/1oNMnueC2hFxcncnmsaqscJFJpnaTEbTpJAPx9L2gD44/edit


src/services/cronJobs/checkProjectVerificationStatus.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  // First email for reminding to add an update
+  if (
+    project.projectUpdate.createdAt <= maxDaysForSendingUpdateReminder &&
+    project.projectUpdate.createdAt > maxDaysForSendingUpdateWarning
+  ) {
+    project.verificationStatus = RevokeSteps.Reminder;
+  }
+
```
mohammadranjbarz: Is it possible to go inside two `if` statement?
If it's not possible, I think it's better use `else if` statement to make it clear

mohammadranjbarz (changes requested):
Thanks @CarlosQ96 , it seems good and integration tests are great, I just requested a tiny request change, after that you can merge it

src/services/cronJobs/checkProjectVerificationStatus.ts
CarlosQ96: Technically is not posible, I tested manually the conditions and thats why they are ordered that way and the status is the limiting factor so it doesn't get into two. And the ifs are by time range.

But I think else if might be safer, just in case.

CarlosQ96:
Thanks Mohammad I did the feedback!

src/services/cronJobs/checkProjectVerificationStatus.ts
```diff
@@ -0,0 +1,183 @@
...
+    case RevokeSteps.Reminder:
+      return NOTIFICATIONS_EVENT_NAMES.PROJECT_BADGE_REVOKE_REMINDER;
+    case RevokeSteps.Warning:
+      return NOTIFICATIONS_EVENT_NAMES.PROJECT_BADGE_REVOKE_WARNING;
+    case RevokeSteps.LastChance:
+      return NOTIFICATIONS_EVENT_NAMES.PROJECT_BADGE_REVOKE_LAST_WARNING;
+    case RevokeSteps.Revoked:
+      return NOTIFICATIONS_EVENT_NAMES.PROJECT_BADGE_REVOKED;
+    case RevokeSteps.UpForRevoking:
+      return NOTIFICATIONS_EVENT_NAMES.PROJECT_BADGE_UP_FOR_REVOKING;
```
mohammadranjbarz, line 179: @CarlosQ96  @WhyldWanderer @divine-comedian 
have we created the autopilot journeys for these events?

src/analytics/analytics.ts
```diff
@@ -8,6 +8,10 @@ export enum NOTIFICATIONS_EVENT_NAMES {
   PROJECT_UNLISTED = 'Project unlisted',
   PROJECT_EDITED = 'Project edited',
   PROJECT_BADGE_REVOKED = 'Project badge revoked',
+  PROJECT_BADGE_REVOKE_REMINDER = 'Project badge revoke reminder',
+  PROJECT_BADGE_REVOKE_WARNING = 'Project badge revoke warning',
+  PROJECT_BADGE_REVOKE_LAST_WARNING = 'Project badge revoke last warning',
+  PROJECT_BADGE_UP_FOR_REVOKING = 'Project badge up for revoking',
```
mohammadranjbarz, line 14: @CarlosQ96 @WhyldWanderer @divine-comedian
have we created the autopilot journeys for these events?

PR #618, opened 2022-09-05
Feat/batch givpower average

aminlatifi:


src/services/cronJobs/syncUserPoewrs.ts
```diff
@@ -2,14 +2,16 @@ import Bull from 'bull';
 import config from '../../config';
```
CarlosQ96, line 2: Just noticed that this TS file is named wrong.

Currently: syncUserPoewrs.ts
Should be: syncUserPowers.ts

src/services/cronJobs/syncUserPoewrs.ts
```diff
@@ -62,41 +64,46 @@ async function addSyncUserPowerJobsToQueue() {
...
+      users,
       fromTimeStamp,
       toTimeStamp,
       givbackRound: previousGivbackRound,
     });
-  });
+
+    totalFetched += users.length;
+    if (totalFetched >= count) break;
+  }
```
CarlosQ96: I don't see this as an infinite loop, as the condition is clearly stated, could this be a good candidate for a do...while loop?
Something like
```
let totalFetched = 0;
let count = 0;
do {
 // logic and http calls
} while (totalFetched < count)
```

I ask because, (do...while) are not common. But seems like the code needs to execute at least once which works with this do...whiles

CarlosQ96 (changes requested):
The code looks good too me, good improvements over the last PR I reviewed.
I just left two comments, fixing a file name and if we could change the infinite loop for the (do..while) in this case.

src/services/cronJobs/syncUserPoewrs.ts
aminlatifi: Oh right

src/services/cronJobs/syncUserPoewrs.ts
aminlatifi: I replaced it.
It's a matter of taste though. Some modern languages like Golang and Rust don't have it anymore.

CarlosQ96:
Thanks a lot Amin, LGTM!

CarlosQ96 (approved):
APPROVED!

PR #619, opened 2022-09-05
Fix/typescript compile

aminlatifi:
After this patch, we can compile the typescript using `tsc` command. Executing js files needs less resources rather than executing direct typescript by ts-node

**Note:** I forked from `master` branch. So we can make a PR to production if needed


CarlosQ96 (commented):
Seems it brought my hotfix from prod to staging, which is weird as staging also has it.

Aside from that, everything seems in order. Ran the tsc and it generates the build fine.

 But I think we need to change the dockerfile? I don't think our current docker config is using the ./build folder or even generating it.

Correct me if I am wrong.

aminlatifi:
> Seems it brought my hotfix from prod to staging, which is weird as staging also has it.
> 
> Aside from that, everything seems in order. Ran the tsc and it generates the build fine.
> 
> But I think we need to change the dockerfile? I don't think our current docker config is using the ./build folder or even generating it.
> 
> Correct me if I am wrong.

Changing dockerfile is the next step. We must test that the service works correctly if we run it by executing js files.
@CarlosQ96, so do you accept this PR finally? I am asking again because you say it applies some changes again to staging.

CarlosQ96 (approved):
Yes I checked over the changes and I think my changes are non issue. You can merge.

PR #623, opened 2022-09-06
Only query for givpower of boosters

aminlatifi:


src/repositories/userPowerRepository.test.ts
```diff
@@ -103,4 +135,36 @@ function findUsersThatDidntSyncTheirPowerTestCases() {
     // usersWithoutPowerSync should not include user1
     assert.isNotOk(usersWithoutPowerSync.find(user => user.id === user1.id));
   });
+  it('should return users without a boosted any project', async () => {
```
CarlosQ96:
```suggestion
  it('should not return users without any boosted project', async () => {
```

CarlosQ96: Just a text suggestion to make it clear, I was getting confused as the headline was not what was ocurring on the code execution.

CarlosQ96 (approved):
Code is fine! Thanks for the updates. 

Minor suggestion: I would change the description of the test to make it clearer vs what it's doing.

After that you can merge. (also testing suggestions like you told me 😄 )

src/repositories/userPowerRepository.test.ts
aminlatifi: Thanks @CarlosQ96

PR #625, opened 2022-09-07
F 16 power boosting add more tests

mohammadranjbarz:


aminlatifi (commented):
LGTM!
But in some parts, we create entities that are prone to be created before. Like userPower for a specific round may have been filled before. As we creat and save multiple entities together, I am not sure failure of one will allow the rest be saved or not.

aminlatifi (approved):

Issue #572, opened 2022-07-24
Update typeorm and  admin-bro dependencies


mohammadranjbarz:

aminlatifi:
It seems the current version of typeorm doesn't support upsert. It's important now as we are implementing automatic cron jobs to insert to db. If any of them does a mistake and tries to insert an entity again one round of process would fail

mohammadranjbarz:
@MohammadPCh  gave this: https://giveth-dapps-v2-git-staging-with-dev-serve-backend-givethio.vercel.app 

A frontend version that uses https://dev.serve.giveth.io as backend

mohammadranjbarz:
@aminlatifi  @CarlosQ96  
I deployed our PR on dev ENV.

@maryjaf  can you test our functionality, especially in admin panel, and if it's ok tell us to merge it to staging
https://giveth-dapps-v2-git-staging-with-dev-serve-backend-givethio.vercel.app/
admin panel: https://dev.serve.giveth.io/admin

maryjaf:
by selecting **project** in admin I got 502  bad gateway, but for other menu like donation i didn't see this problem
@mohammadranjbarz 
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/111529185/212095250-73614f71-fe39-4476-88d3-fff447512b2d.png">
<img width="1437" alt="image" src="https://user-images.githubusercontent.com/111529185/212095409-0721bdb6-c8ab-4476-bde9-e0dd98dcd19f.png">

mohammadranjbarz:
@maryjaf Can you test it again?
maybe you have had some internet problem, because it's working for me

https://dev.serve.giveth.io/admin/resources/Project/records/37/show
<img width="1433" alt="Screen Shot 1401-10-25 at 09 49 26" src="https://user-images.githubusercontent.com/9850545/212526092-a5fd3573-e040-47e9-988b-e8699b803f66.png">

mohammadranjbarz:
https://github.com/Giveth/impact-graph/issues/813

maryjaf:
Thanks @mohammadranjbarz  the problem related to edit project has been resolved
Plz check this scenario
the status of verification form of project is submitted
when the admin user change this status to verified
error message is shown as below pic but the status of verification form changed into verified correctly
<img width="958" alt="image" src="https://user-images.githubusercontent.com/111529185/213917235-3cc23686-b96d-4064-b7e8-853b78c955fa.png">

[...]

PR #630, opened 2022-09-09
Update admin-bro and typeorm

mohammadranjbarz:
related to #572


mohammadranjbarz:
@aminlatifi  I updated typeorm and admin-bro, all tests pass but when I opened the admin bro I saw some issues that may take too much time but are necessary before merging

Screen Shot 1401-06-30 at 16 24 49

Screen Shot 1401-06-30 at 16 24 55

The relations don't work properly and when opening project page we get error, I think our custom views need to have some changes, so you tell me whether  it worth to put more time to finish that or not

PS: @CarlosQ96  Wants to take a look to see if can find a good solution or not

CarlosQ96:
Ill take a look

CarlosQ96 (approved):
LGTM, ran tests and verified data wasn't being deleted.

Lets merge develop first then go for staging!

PR #632, opened 2022-09-12
Fix loop of projects and change october revoke if

CarlosQ96:
Related to revoke badge.
Adjusting the projects to be revoked on the 15th of october.

Fixed a bug I found related to the loop of projects.


src/services/cronJobs/checkProjectVerificationStatus.ts
```diff
@@ -80,19 +84,18 @@ export const checkProjectVerificationStatus = async () => {
     maxDaysForSendingUpdateReminder,
   );
 
-  for (const project of projects) {
```
CarlosQ96, line 83: This iteration was only running the first promise and the rest sometime in the future so latest test wasn't working. This could lead to false positives or unexecuted code.
Fixed with Promise.all

mohammadranjbarz (approved):
Thanks @CarlosQ96  Seems fine to me

PR #636, opened 2022-09-12
Refresh user project power views

mohammadranjbarz:
- [x] After users powers are fetched
- [x] Set interval for refresh views periodically
- [x] After setting current round -  as we still don't know how update givback round, this item is missing

### DIFF
--- a/src/services/cronJobs/syncUserPowers.ts
+++ b/src/services/cronJobs/syncUserPowers.ts
@@ -12,6 +12,7 @@ import {
 import { User } from '../../entities/user';
 import { refreshUserProjectPowerView } from '../../repositories/userProjectPowerViewRepository';
 import { refreshProjectPowerView } from '../../repositories/projectPowerViewRepository';
+import { setPowerRound } from '../../repositories/powerRoundRepository';
 
 const syncUserPowersQueue = new Bull<SyncUserPowersJobData>(
   'verify-userPower-queue',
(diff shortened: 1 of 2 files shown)


aminlatifi:
LGTM! but setting the current givback round is too important! because without it APIs will return an empty list!!

aminlatifi:
I suggest adding setting the current round to the same cron job which refreshes as a first resort

aminlatifi (changes requested):
We must set current round.

src/services/cronJobs/syncUserPowers.ts
```diff
@@ -129,7 +142,18 @@ export function processSyncUserPowerJobs() {
           givbackRound,
           users,
         });
-        logger.debug('inserting userPowers...', users.length);
+        const remainingJobsCount = await syncUserPowersQueue.count();
+        if (remainingJobsCount === 0) {
+          await setPowerRound(givbackRound);
```
mohammadranjbarz, line 147: @aminlatifi  I think it's the best place for updating current round because all user powers are synced and we calculated round well, 
what do you think?

mohammadranjbarz:
@aminlatifi Can we merge it?

aminlatifi:
Yes, Thanks @mohammadranjbarz

aminlatifi (approved):

PR #638, opened 2022-09-13
Tested projectPowerView join

aminlatifi:


aminlatifi:
@mohammadranjbarz Note the projectPowerView create script had a problem. So we need to rerun its migration everywhere it was created.

mohammadranjbarz (approved):
Thanks @aminlatifi  LGTM
I would delete the view and corresponding migration file in DEV db

PR #641, opened 2022-09-14
Feat/project power rank and update time

aminlatifi:
This PR implements some remaining givpower related features
 - Project ranks based on total boosted power
 - Last update time of rank and boosted power


mohammadranjbarz (approved):
@aminlatifi Very nice usage of Postgres features, LGTM

Issue #620, opened 2022-09-06
Update dockerfile to execute js files rather than ts


aminlatifi:
After this https://github.com/Giveth/impact-graph/pull/619, we are able to compile typescript to javascript completely. 
In docker file we should apply these changes:
1. compile typescript files and have result js files in the `./build` directory
2. replace executing typescript files by executing js files

**Note:** We must be sure the server works correctly after this change

aminlatifi:
@CarlosQ96 What's the latest update? are we stuck with admin-bro?

CarlosQ96:
Yeah, I investigated a lot to fix adminbro compilations.
So I did:
1. Migrated the adminbro jsx components to tsx for they are compiled successfully by tsc.
2. Imported dom elements to the tsconfig so tsc compiled adminbro correctly.

However now that the build is done, adminbro js isn't resolving the components path correctly. So the admin runs, but custom components show as red squares:

<img width="1357" alt="image" src="https://user-images.githubusercontent.com/92376054/190577506-33974170-2e6d-4c1e-af6b-326d1a83b746.png">

CarlosQ96:
I found this article where someone encountered this issue and fixed it in a really.... weird way, but he explains adminbro is looking for the uncompiled tsx files by default.

Link: https://lightrun.com/answers/softwarebrothers-adminjs-nestjs-provide-a-more-explanatory-example-project

CarlosQ96:
That's my last blocker, if I fix it, once the component loads, the build-dev scripts I wrote can be modified to each env (staging/de/prod) and called on docker to run the app.

https://github.com/Giveth/impact-graph/pull/643 this is the draft pr, not ready for review yet as I am fixing this last issue.

CarlosQ96:
Ok I managed to fix it, but the only solution I found in forums, was copy the uncompiled tsx files to the build folder. 
That worked perfectly.

PR #643, opened 2022-09-16
Feature/docker tsc build scripts update

CarlosQ96:
Related to https://github.com/Giveth/impact-graph/issues/620


CarlosQ96:
@mohammadranjbarz can you take a look when running docker compose? I am having a env var issue. After that we can do a test deploy.

CarlosQ96:
Ready for review Mohammad. Let me know if anything needs to be changes, also if you feel like I ran the app without and with docker and adminbro runs, components render fine.
Let's do a final review and do a test deploy!

CarlosQ96:
Btw I felt the migrations took a bit of time to run, but maybe it's because they ran from zero. 

That's my only caveat. Aside from that everything was fine.

CarlosQ96:
Last question @mohammadranjbarz , now that I am reading a bit. Should we compile tests too? and run them from build?

I personally don't think it will speed up things, because most use the database directly, so the overhead we are having is db connections related.

If we want to optimize tests, we have to reduce/mock some database calls. But let me know what you think? and if so, we should make a separate issue for this. It's not high priority right now.

mohammadranjbarz:
> Btw I felt the migrations took a bit of time to run, but maybe it's because they ran from zero.
> 
> That's my only caveat. Aside from that everything was fine.
You know  I think maybe we need to convert migration files to js and run them as js files, it may solve the bootstrap speed problem, I created an issue for that
https://github.com/Giveth/impact-graph/issues/651

mohammadranjbarz (approved):
Thanks @CarlosQ96  for your work on this issue, LGTM

PR #646, opened 2022-09-19
Feature/docker tsc build scripts update revision

mohammadranjbarz:
Now with help of @aminlatifi , and changes that I made you can successfully build docker image locally and run it properly, you can run it with :
`docker-compose -f docker-compose-local.yml up  --build`
 **--build** flag is important here, because currently we don't support watching changes so when you change anything in code you need to build docker image again so you have to pass **--build**
 
 before we were using to postgres and redis the their port was bonded to local system ports, but now that we run application with docker we don't have access to local machine ports easily so it's better to connect to postgres and redis by their container info  so if you want to don't get error put/edit these lines in `config/development.env`
 
 ```
 TYPEORM_DATABASE_HOST=impact-graph-postgress
TYPEORM_DATABASE_PORT=5432

REDIS_PORT=6379
REDIS_HOST=redis-giveth

SHARED_REDIS_PORT=6379
SHARED_REDIS_HOST=redis-giveth
 ```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,9 @@ coverage
 # parcel cache
 .cache
 
-config/development.env
-config/production.env
+config/*
+!config/example.env
+!config/test.env
 .vscode/settings.json
 
 *.rdb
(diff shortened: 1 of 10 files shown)


CarlosQ96 (approved):
Tested the app with and without docker and it works! Components are rendering fine! Thanks a lot, lets merge to my original branch.

PR #647, opened 2022-09-19
Fixed userProjectPowerView rank problem on pagination

aminlatifi:


mohammadranjbarz (approved):
Thanks @aminlatifi 
I would deploy it on DEV server just now

PR #652, opened 2022-09-20
Fix running test without dropping db

mohammadranjbarz:


mohammadranjbarz:
@aminlatifi  dropSchema works fine we just need to create `cronJob` db before running tests ( but when you change this and moving to givethio db, we can remove that as well)

aminlatifi:
> @aminlatifi dropSchema works fine we just need to create `cronJob` db before running tests ( but when you change this and moving to givethio db, we can remove that as well)

I would put cronJob tables in the same db, not in a separate one. Though pg_cron extension creates the db iteself.

mohammadranjbarz:
@aminlatifi  I updated typeorm and admin-bro, all tests pass but when I opened the admin bro I saw some issues that may take too much time but are necessary before merging

Screen Shot 1401-06-30 at 16 24 49

Screen Shot 1401-06-30 at 16 24 55

The relations don't work properly and when opening project page we get error, I think our custom views need to have some changes, so you tell me wether  it worth to put more time to finish that or not

mohammadranjbarz:
@aminlatifi  I think we don't need this anymore, can we close it?

aminlatifi:
@mohammadranjbarz Yes, we can close it. Thanks.

Issue #568, opened 2022-07-20
Review and merge bump dependencies PRs


mohammadranjbarz:
I suggest we do it after our current release

- [x] https://github.com/Giveth/impact-graph/pull/533
- [x] https://github.com/Giveth/impact-graph/pull/534
- [x] https://github.com/Giveth/impact-graph/pull/559
- [x] https://github.com/Giveth/impact-graph/pull/561

PR #653, opened 2022-09-21
Security Updates: Update husky, moment-timezone, terser, parse-url, semver

CarlosQ96:
Related to https://github.com/Giveth/impact-graph/issues/568


CarlosQ96:
Ran server locally and with docker and everything seems in order. Cronjobs were running I was getting the same messages I got before the upgrade of bull dependencies.

Seems safe.

package.json
```diff
@@ -55,6 +55,7 @@
     "handlebars": "^4.7.7",
     "honeyswap-sdk": "^1.0.0",
     "html-pdf-node": "^1.0.7",
+    "husky": "^4.3.8",
```
mohammadranjbarz: How about setting exact version for these dependencies?
(for instance `4.3.8` instead of `^4.3.8` to prevent unwanted upgrades

CarlosQ96: You are right. I think there are no more updates actually. 4.3.8 is the last one. 
Thanks

CarlosQ96: Upgrading husky to 5.0.0 will break it FYI. So lets leave it in 4.3.8

mohammadranjbarz: Thanks @CarlosQ96

mohammadranjbarz:
@maryjaf  It was just about updating some packages, if main functionality of stage is ok you can move it to DONE ( they were not risky packages, so even if just staging is up we can move it to DONE without lots of testing)

Issue #664, opened 2022-09-27
Backend: Adding "Accept Funds on Gnosis" to filters


MoeNick:

mohammadranjbarz:
@maryjaf  The backend is ready , you can test it now with postman, so I move this to unit test

@MohammadPCh when frontend implement it we can move to UAT

sample request data for postman
```
query {
  projects(
    take: 10,
    skip: 0,
    searchTerm: "",
    filterBy: {
        field: AcceptFundOnGnosis,
        value: true
        },
    orderBy: {
        field: CreationDate,
        direction: DESC
         }
  ) {
    totalCount
    projects {
      traceCampaignId
      id
      addresses {
        networkId
        isRecipient
        address
      }
      categories {
        id
        name
      }
      title
      slug
      qualityScore
      walletAddress
      creationDate
      adminUser{
        name
      }
      users {
          name
      }
      updatedAt
      totalReactions
      totalDonations
      totalTraceDonations
    }

  }
}

```

every project that we get with this payload would have a
```
  {
                            "networkId": 100,
                            "isRecipient": true,
                            "address": "walletAddress..."
                        }
```
in the addresses array, it means it have a recipient address in gnosis(xdai) 

PS: `netowrkId:100` is for gnosis chain

mohammadranjbarz:
> @maryjaf The backend is ready , you can test it now with postman, so I move this to unit test
> 
> @MohammadPCh when frontend implement it we can move to UAT
> 
> sample request data for postman
> 
> ```
> query {
>   projects(
>     take: 10,
>     skip: 0,
>     searchTerm: "",
>     filterBy: {
>         field: AcceptFundOnGnosis,
>         value: true
>         },
>     orderBy: {
>         field: CreationDate,
>         direction: DESC
>          }
>   ) {
>     totalCount
>     projects {
>       traceCampaignId
>       id
>       addresses {
>         networkId
>         isRecipient
>         address
>       }
>       categories {
>         id
>         name
>       }
>       title
>       slug
>       qualityScore
>       walletAddress
>       creationDate
>       adminUser{
>         name
>       }
>       users {
>           name
>       }
>       updatedAt
>       totalReactions
>       totalDonations
>       totalTraceDonations
>     }
> 
>   }
> }
> ```
> 
> every project that we get with this payload would have a
> 
> ```
>   {
>                             "networkId": 100,
>                             "isRecipient": true,
>                             "address": "walletAddress..."
>                         }
> ```
> 
[...]

PR #673, opened 2022-10-02
Add filter for accept fund on gnosis

mohammadranjbarz:
related to #664

### DIFF
--- a/src/resolvers/projectResolver.test.ts
+++ b/src/resolvers/projectResolver.test.ts
@@ -555,55 +555,55 @@ function projectsTestCases() {
 
   // TODO below test cases doesnt pass but I checked https://github.com/Giveth/giveth-dapps-v2/blob/develop/src/apollo/gql/gqlProjects.ts
   // and it seems the frontend doesnt need addresses in projects() query, so I commented below test case
-  // it('should return projects, filter by accept donation on gnosis, return all addresses', async () => {
-  //   const savedProject = await saveProjectDirectlyToDb({
-  //     ...createProjectData(),
-  //     title: String(new Date().getTime()),
-  //     slug: String(new Date().getTime()),
-  //   });
-  //
-  //   const result = await axios.post(graphqlUrl, {
-  //     query: fetchAllProjectsQuery,
-  //     variables: {
-  //       filterBy: {
-  //         field: 'AcceptFundOnGnosis',
-  //         value: true,
-  //       },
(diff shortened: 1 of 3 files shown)


src/resolvers/projectResolver.test.ts
```diff
@@ -510,6 +512,137 @@ function projectsTestCases() {
...
+  //     ),
+  //   );
+  //   assert.isOk(
+  //     project.addresses.find(
+  //       address =>
+  //         address.isRecipient === true &&
+  //         address.networkId === NETWORK_IDS.MAIN_NET,
+  //     ),
+  //   );
+  // });
```
mohammadranjbarz: @CarlosQ96  I couldn't find a way to pass this test but as I explained in comment, the frontend doesn't use that, so it's safe , but if you can improve the query to pass this, I would appreciate it

src/server/bootstrap.ts
```diff
@@ -268,18 +265,6 @@ export async function bootstrap() {
...
-    ).getAuthUrl({
-      // trackId: generateRandomEtheriumAddress(),
-      trackId: 'STATE',
-    });
-    logger.info('twitter auth url', authUrl);
-    // const accessToken = await twitterAdapter.getUserInfoByOauth2Code({
-    //   oauth2Code:
-    //     'SDg1b1otX3lYaURDZjN3emQtTjVwVUMwMFNmeGFjQ0tLWlBNQnhobEszQ19hOjE2NTcxMTU5MjIzMjg6MTowOmFjOjE',
-    // });
-    // logger.info('twitter accessToken', accessToken);
```
mohammadranjbarz, line 282: There was an unnecassary and unused code

src/resolvers/projectResolver.ts
```diff
@@ -441,6 +443,12 @@ export class ProjectResolver {
       return query.andWhere(`project.${filter} ${isRequested} NULL`);
     }
 
+    if (filter === 'acceptFundOnGnosis' && filterValue) {
+      return query.andWhere(
+        `"isRecipient"=true AND "networkId"=${NETWORK_IDS.XDAI}`,
+      );
```
mohammadranjbarz: They will query on `addresses` entity ( the we joined it to project)

CarlosQ96 (approved):
LGTM from my side, all tests pass now. You can test the subquery, but everything else looks good!

PR #675, opened 2022-10-03
Fix changing categories migrations

mohammadranjbarz:


mohammadranjbarz:
@CarlosQ96 Some project has categories so when we set another category we get duplicate error, si I thought maybe it's better to use a for..loop and set them one by one see if one of them got error , we are still ok

Screen Shot 1401-07-11 at 19 31 08

CarlosQ96 (approved):
Thanks Mohammad! LGTM, honestly the solution to avoid loops is upsert but typeorm doesn't support it.

I found another workaround but its super wanky not worth the effort, tons untested methods.

CarlosQ96 (approved):

PR #678, opened 2022-10-05
Revoke badge revisions

CarlosQ96:
Related to the scenario @maryjaf reported in https://github.com/Giveth/giveth-dapps-v2/issues/1497

Also added a missed hook on the project, when a project update is modified it should be reflected on the project entity.

### DIFF
--- a/src/entities/project.ts
+++ b/src/entities/project.ts
@@ -495,6 +495,11 @@ class ProjectUpdate extends BaseEntity {
     await Project.update({ id: this.projectId }, { updatedAt: moment() });
   }
 
+  @AfterUpdate()
+  async updateProjectStampOnUpdate() {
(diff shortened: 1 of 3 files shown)


src/entities/project.ts
```diff
@@ -495,6 +495,11 @@ class ProjectUpdate extends BaseEntity {
     await Project.update({ id: this.projectId }, { updatedAt: moment() });
   }
 
+  @AfterUpdate()
+  async updateProjectStampOnUpdate() {
+    await Project.update({ id: this.projectId }, { updatedAt: moment() });
+  }
+
```
CarlosQ96, line 502: This is a missed hook when this was implemented way back.

I need it for the feedbacks of ashley and nikola.

mohammadranjbarz (approved):
Thank you @CarlosQ96  LGTM

PR #681, opened 2022-10-06
Revoke badge improvements

CarlosQ96:
Basically, if the project description is updated that is considered as an update, not just adding a new projectUpdate entity.

So this simplified the develop and I just consider the project's updatedAt Column.

UpdatedAt column is also automatically updated when a new projectUpdate is added, updated or removed.

### DIFF
--- a/src/repositories/projectRepository.ts
+++ b/src/repositories/projectRepository.ts
@@ -28,21 +28,9 @@ export const projectsWithoutUpdateAfterTimeFrame = async (date: Date) => {
       'project.projectVerificationForm',
       'projectVerificationForm',
     )
-    .innerJoinAndMapOne(
-      'project.projectUpdate',
-      ProjectUpdate,
-      'projectUpdate',
-      `project.id = projectUpdate.projectId AND projectUpdate.id = (
-        SELECT project_update.id
-        FROM project_update
-        WHERE project_update."projectId" = project.id
-        ORDER BY project_update.id DESC
-        LIMIT 1
-      )`,
-    )
     .where('project.isImported = false')
     .andWhere('project.verified = true')
-    .andWhere('projectUpdate.createdAt < :badgeRevokingDate', {
+    .andWhere('project.updatedAt < :badgeRevokingDate', {
       badgeRevokingDate: date,
     })
     .getMany();
--- a/src/repositories/projectVerificationRepository.ts
+++ b/src/repositories/projectVerificationRepository.ts
@@ -30,18 +30,14 @@ export const updateProjectVerificationFormStatusOnly = async (
   projectVerificationFormId: number,
   verificationStatus: PROJECT_VERIFICATION_STATUSES,
   reviewerId?: number,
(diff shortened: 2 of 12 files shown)


CarlosQ96:
@mohammadranjbarz the ropsten testnet seems to be disabled or down, I know mateo did a migration on the subgraph. So maybe we hit the same deprecation issue.
I removed some tests that I thought were not required as we had goerli test below. And others that I think it would be good to rewrite them as they are an interesting case.

src/server/adminBro.ts
```diff
@@ -2005,8 +2006,15 @@ export const verifyProjects = async (
   // prioritize revokeBadge
   const verificationStatus = revokeBadge ? false : verified;
   try {
+    const updateParams = { verified: verificationStatus };
+
+    if (verificationStatus) {
+      const key = 'verificationStatus';
+      updateParams[key] = null;
+    }
```
CarlosQ96, line 2014: If the project is verified, the verification status for revoking is reset to null.

src/repositories/projectVerificationRepository.ts
```diff
@@ -29,21 +29,37 @@ export const createProjectVerificationForm = async (params: {
...
   projectVerificationFormId: number,
   verificationStatus: PROJECT_VERIFICATION_STATUSES,
+  reviewerId?: number,
 ): Promise<UpdateResult> => {
+  const updateParams = { status: verificationStatus };
+
+  if (reviewerId) {
+    const key = 'reviewerId';
+    updateParams[key] = reviewerId;
+  }
```
CarlosQ96: I missed setting the verification AdminId

src/repositories/projectRepository.ts
```diff
@@ -28,21 +28,9 @@ export const projectsWithoutUpdateAfterTimeFrame = async (date: Date) => {
...
-        WHERE project_update."projectId" = project.id
-        ORDER BY project_update.id DESC
-        LIMIT 1
-      )`,
-    )
     .where('project.isImported = false')
     .andWhere('project.verified = true')
-    .andWhere('projectUpdate.createdAt < :badgeRevokingDate', {
+    .andWhere('project.updatedAt < :badgeRevokingDate', {
       badgeRevokingDate: date,
```
CarlosQ96, line 34: Simplified the query as any update on the project should relieve them from being revoked (description, title)

And also if they add any projectUpdate entity o modify the current one. It would automatically also update the updatedAt field on the project. So I think using updatedAt is good in this case.

mohammadranjbarz:
> @mohammadranjbarz the ropsten testnet seems to be disabled or down, I know mateo did a migration on the subgraph. So maybe we hit the same deprecation issue. I removed some tests that I thought were not required as we had goerli test below. And others that I think it would be good to rewrite them as they are an interesting case.

Thanks man, you did a good thing

mohammadranjbarz (approved):
Thanks @CarlosQ96 
Nice job, LGTM

PR #690, opened 2022-10-17
Add new segment api calls

CarlosQ96:
Uses a singleton instance to call the same npm package always, without reinitializing queues or redis. This is to save memory.

This pr does api calls directly on segment and it works similarly to previous implementation.

Known issue: I couldn't get the queues to process the events inside the npm package. (even tho it works on the npm package script I did)


CarlosQ96:
I was able to send many events directly and all arrived. I think direct https calls are more reliable than queues.
image

CarlosQ96:
Upon further testing with batches of 9, @divine-comedian @mohammadranjbarz events are still getting lost.

So in the end even with this rewrite, I feel segment is not reliable. (and I am hitting their API DIRECTLY, no inmemory queues like they had in the code)

I feel the best thing we can do is move from segment to something else like google analytics, at the very list I trust google 😄

Issue #699, opened 2022-10-25
Create notifications for donors of projects and who liked projects 


mohammadranjbarz:
Not for all of the notification types, just add donors and who liked for our current notifications

related to https://github.com/Giveth/notification-center/issues/6

PR #700, opened 2022-10-25
Send Notifications to donors and people who liked the projects

mohammadranjbarz:
related to #699


src/entities/reaction.ts
```diff
@@ -29,8 +30,12 @@ export class Reaction extends BaseEntity {
   @Column({ nullable: true })
   projectUpdateId: number;
 
+  @Field(type => User, { nullable: true })
+  @ManyToOne(type => User, { nullable: true })
+  user: User;
   @Field(type => ID)
   @Column()
+  @RelationId((reaction: Reaction) => reaction.user)
```
CarlosQ96: As much as I know this is the right way, this will most likely delete the data from the UserId column, by running a query on relationId to set the column.  (Because we using sync.)

Confirm this isn't the case and we can proceed.

CarlosQ96: Else we will have to do a different query.

src/repositories/donationRepository.ts
```diff
@@ -72,3 +72,23 @@ export const findDonationById = async (
...
+    .addSelect(['user.walletAddress', 'user.email'])
+    .distinctOn(['user.walletAddress'])
+    .where(`"projectId"=:projectId`, {
+      projectId,
+    })
+    .getMany();
+
+  return donations.map(donation => {
+    return {
+      walletAddress: donation.user.walletAddress?.toLowerCase() as string,
```
CarlosQ96: You can avoid this loop entirely by using ```.getRawMany()``` instead, this returns the selected data only as json and not entity mapped data.
And in the select you can do directly:
```
.select('LOWER(user.walletAddress) AS "walletAddress", user.email as email')
```

This should return:
```
[{ walletAddress: 'xxxxx', email: 'xxxx' }]
```

src/repositories/reactionRepository.ts
```diff
@@ -0,0 +1,20 @@
...
+    .where(`"projectId"=:projectId`, {
+      projectId,
+    })
+    .getMany();
+
+  return reactions.map(reaction => {
+    return {
+      walletAddress: reaction.user.walletAddress?.toLowerCase() as string,
+      email: reaction.user.email,
+    };
```
CarlosQ96: same as the donation repository, to avoid extra processing and loops.

src/repositories/donationRepository.ts
CarlosQ96: From what I see in the code we have 2 loops, 
Fetch query -> map data correctly -> map to enqueue.

If we delete the middle map this would greatly decrease code complexity.

src/entities/reaction.ts
CarlosQ96: Rest assured... leftJoins work without the need of relationId or ManyToOne, they are a bit less efficient because they don't have an index. But I think we can add it separately (haven't test this).
```
      .leftJoin(User, 'user', 'reaction.userId = user.id')
```

src/repositories/reactionRepository.ts
```diff
@@ -0,0 +1,20 @@
+import { Reaction } from '../entities/reaction';
+
+export const findUsersWhoLikedProject = async (
+  projectId: number,
+): Promise<{ walletAddress: string; email?: string }[]> => {
+  const reactions = await Reaction.createQueryBuilder('reaction')
+    .leftJoin('reaction.user', 'user')
```
CarlosQ96: If relationId deletes column, like in other comment. Remove it and add:
```
      .leftJoin(User, 'user', 'reaction.userId = user.id')
```

CarlosQ96 (changes requested):
Hey Mohammad thanks, I agree we gotta enqueue the events as they can be too many. So I think bull is awesome here.
However I have some suggestions to improve code:
1. With a different typeorm Query (getRawMany and LOWER). We can remove 1 extra loop for more efficiency. 
2. I think relationId deletes column data, so we have to do a leftJoin without relation.

CarlosQ96 (approved):
Awesome! You can merge 👍 thanks

src/entities/reaction.ts
mohammadranjbarz: Oh nice catch, thank you I rewrite that function considering all your feedback, it works now

src/repositories/donationRepository.ts
mohammadranjbarz: Done

src/repositories/reactionRepository.ts
mohammadranjbarz: Done

PR #706, opened 2022-11-03
Add givPower filter to allProjects, add verified orderBy to givPower sorting

mohammadranjbarz:
https://github.com/Giveth/giveth-dapps-v2/issues/1736


CarlosQ96 (approved):
Thanks Mohammad LGTM

PR #708, opened 2022-11-04
Just use our segment library

mohammadranjbarz:
@CarlosQ96 
Through moving segment events to notificaiton-center I noticed that somewhere in our codes we use segment directly ( instead of using the package that you wrote) so I changed to to use your library and after that will start migrating to notificaiton-center


CarlosQ96 (approved):
LGTM thanks Mohammad we can keep migrating!

PR #713, opened 2022-11-07
Get rid of segment and call notificaion-center instead

mohammadranjbarz:
related to  https://github.com/Giveth/giveth-dapps-v2/issues/709

 * Remove calling segment directly and instead calling notification-center
 * Remove all the things related to giveth trace


src/resolvers/projectVerificationFormResolver.ts
```diff
@@ -183,19 +184,10 @@ export class ProjectVerificationFormResolver {
...
-
       await sendMailConfirmationEmail(email, project, token);
-
-      SegmentAnalyticsSingleton.getInstance().track(
-        NOTIFICATIONS_EVENT_NAMES.SEND_EMAIL_CONFIRMATION,
-        `givethId-${userId}`,
-        emailConfirmationData,
-        null,
-      );
+      await getNotificationAdapter().projectSendEmailConfirmation({
```
CarlosQ96: I added this before but I think we can remove it. Segment doesn't send any email as we send it with:
```
      await sendMailConfirmationEmail(email, project, token);
```

Also tracking this data is not required, and thanks for removing the callbackUrl from the analytics, that info shouldn't be exposed.

I am all for deleting code as we can speed up requests.

CarlosQ96 (approved):
Thanks a lot Mohammad for deleting so much code that we don't need anymore. aNd also really clean implementation.

CarlosQ96:
I think lets merge and start testing with MaryJaf thats the best thing we can do in staging to see if anything goes wrong.

src/resolvers/projectVerificationFormResolver.ts
```diff
@@ -183,19 +184,10 @@ export class ProjectVerificationFormResolver {
       await projectVerificationForm.save();
 
       const callbackUrl = `https://${dappUrl}/verification/${project.slug}/${token}`;
```
CarlosQ96: if the callback url is not used, lets remove it here. I see you build it inside the sendMailConfirmationEmail method.

src/resolvers/projectVerificationFormResolver.ts
mohammadranjbarz: Nice, thanks for mentioning that, I remove it from both notificaiton-center and impact-graph

PR #714, opened 2022-11-08
Feature update milestones fields

CarlosQ96:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/1745

### DIFF
--- a/src/entities/projectVerificationForm.ts
+++ b/src/entities/projectVerificationForm.ts
@@ -74,13 +74,16 @@ export class ProjectContacts {
 export class Milestones {
   @Field(type => String, { nullable: true })
   foundationDate?: String;
-
   @Field({ nullable: true })
   mission?: string;
   @Field({ nullable: true })
   achievedMilestones?: string;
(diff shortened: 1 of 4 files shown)


src/entities/projectVerificationForm.ts
```diff
@@ -74,13 +74,16 @@ export class ProjectContacts {
...
-
   @Field({ nullable: true })
   mission?: string;
   @Field({ nullable: true })
   achievedMilestones?: string;
   @Field(type => [String], { nullable: true })
   achievedMilestonesProofs?: string[];
+  @Field(type => String, { nullable: true })
+  problem?: string;
+  @Field(type => String, { nullable: true })
```
CarlosQ96, line 85: I tested adding the new fields with existing data and running server
Then removing them with existing data that had this new fields and running server.

and the info was not deleted. This doesn't seem to delete data when server run db sync. So I think its safe

mohammadranjbarz (approved):
Thanks @CarlosQ96  LGTM

PR #715, opened 2022-11-08
Not include non-verified projects in power boosting snapshots

mohammadranjbarz:


src/repositories/dbCronRepository.test.ts
```diff
@@ -124,4 +135,64 @@ describe('db cron job test', () => {
 
     await unSchedulePowerBoostingSnapshot();
   });
+
+  // This test takes one minutes to become complete, just will run it in special cases manually
+  it.skip('should fill givpower boosting snapshot, not include non-verified project power boostings', async () => {
```
mohammadranjbarz, line 140: @CarlosQ96 
If you want to test you should remove the `.skip` from two test cases and run `npm run test:dbCronRepository`
You will see that tests passed, if you have a problem on local testing you can see this Action
https://github.com/Giveth/impact-graph/actions/runs/3421358467

That I uncommented those two test cases to see if they pass or not

CarlosQ96 (approved):
LGTM Mohammad thanks!

src/repositories/powerSnapshotRepository.test.ts
```diff
@@ -299,4 +254,50 @@ describe('test balance snapshot functions', () => {
...
+    assert.isOk(result);
+    assert.equal(result?.id, snapShot.id);
+    assert.equal(result?.blockNumber, snapShot.blockNumber);
+  });
+
+  it('should not find powerSnapshot by invalid id', async () => {
+    const result = await findPowerSnapshotById(100000000);
+    assert.isNotOk(result);
+  });
+}
```
mohammadranjbarz, line 303: It's nothing, I just wrapped some test cases to a function

migration/1663594895751-takePowerSnapshotProcedure.ts
```diff
@@ -1,7 +1,7 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
 
 // tslint:disable-next-line:class-name
-export class TakePowerBoostingSnapshotProcedure1663594895750
+export class TakePowerBoostingSnapshotProcedure1663594895751
```
mohammadranjbarz, line 4: I changed the name to make sure it will run again although it already has been executed

CarlosQ96: Got it!

CarlosQ96 (approved):

mohammadranjbarz:
@maryjaf  @MoeNick  @laurenluz 

Now when creating snapshots we exclude power boosting of non verified projects

maryjaf:
We tested this with @mohammadranjbarz 
Now, after the project is unverified, the amount of project boosts will be calculated as 0 in the next snapshots.
--In the first picture, my project has been verified and has a rank of 4, and we can see the last snapshot, which contains three boosts.
image
image

--After I unverified it, the snapshot is as below, as you can see in the picture, the boosts of the project are zero.

image

--when the project status was unverified, I removed the boost that my user had boosted to this project

--After a few minutes, when I re-verified, it will be displayed as the image below
 that the boost for the project is not displayed and it is true because it has been unverified in this time
 
image

--Finally after a round, the project's  current rank and list of boosting updated correctly (my user boost has been remved from the list correctly) 
image

In this case, there is only one problem, the expected rank is not displayed correctly in the next round and for projects that have no boost, the last available rank is displayed, which are also asked in the issue ##1736

laurenluz:
Hey waaait

@mohammadranjbarz @maryjaf 

If a project becomes unverified, it should not "lose" its GIVpower... it should still have the GIVpower on the project, only... it will not have a rank because it's GIVpower doesn't mean anything.... the GIVpower should basically be "hidden" or excluded from the calcuations of all the ranks.

The reason we keep the GIVpower on the project after it becomes unverified is because, in this case, it can become re-verified again. If it becomes reverified, it will have the lowest rank to start... and it's rank can update at the start of the next GIVbacks round (considering the amount of GIVpower that is still on it).

I hope that makes sense, please lmk if you have questions.

mohammadranjbarz:
> Hey waaait
> 
> @mohammadranjbarz @maryjaf
> 
> If a project becomes unverified, it should not "lose" its GIVpower... it should still have the GIVpower on the project, only... it will not have a rank because it's GIVpower doesn't mean anything.... the GIVpower should basically be "hidden" or excluded from the calcuations of all the ranks.
> 
> The reason we keep the GIVpower on the project after it becomes unverified is because, in this case, it can become re-verified again. If it becomes reverified, it will have the lowest rank to start... and it's rank can update at the start of the next GIVbacks round (considering the amount of GIVpower that is still on it).
> 
> I hope that makes sense, please lmk if you have questions.

@laurenluz  Now we do the same,

Imagine a project has 1000 givPower, we unverify it, so when getting snapshot for the round we don't count that project's givPowers, and once we make it verified we include those givPowers in new snapshots so since next round it would have some rank, I think it's similar to what you said, am I right?

maryjaf:
> Hey waaait
> 
> @mohammadranjbarz @maryjaf
> 
> If a project becomes unverified, it should not "lose" its GIVpower... it should still have the GIVpower on the project, only... it will not have a rank because it's GIVpower doesn't mean anything.... the GIVpower should basically be "hidden" or excluded from the calcuations of all the ranks.
> 
> The reason we keep the GIVpower on the project after it becomes unverified is because, in this case, it can become re-verified again. If it becomes reverified, it will have the lowest rank to start... and it's rank can update at the start of the next GIVbacks round (considering the amount of GIVpower that is still on it).
> 
> I hope that makes sense, please lmk if you have questions.

Yes, that's correct
Exactly the same situation happened for the above project that we tested
After being reverified and in the next round of boosts, boost and his current rank were displayed correctly.
I forgot to attach on above comment the last picture after reverification and the round is over, it is like this:
(For a better test,  when the project status was unverified, I removed the boost that my user had boosted to this project, and as you can see in the picture, the list of boosting will be displayed correctly after re-verification)

After reverify:
image

maryjaf:
@mohammadranjbarz , just one question, now on giveconomy environment, after the re-verification of the project, it took a round (15 min)for the boosting list and project rank to be updated correctly.
In the production environment, this time takes two weeks?
Or after several snapshots of the project with verified status, list of boosting & rank will be updated ?

laurenluz:
@maryjaf 

on GIVeconomy environment, rounds should be 20 min (if I'm not mistaken!) and snapshots happen every 5 min.
on production, rounds will be 2 weeks, and snapshots every 4 hrs.

The rank should only update at the beginning of a new round. So every 20 min on GIVeconomy, and it will be every 2 weeks on production.

Verified status can update at any time.

GIVpower on a particular project should update at any moment (as the users update their GIVpower) - but this is yet-to-be implemented (just discussed it today).

PR #716, opened 2022-11-09
Remove fromWalletAddress when event is donation recieved

CarlosQ96:
Related to https://github.com/Giveth/giveth-planning/issues/976

### DIFF
--- a/src/services/segment/DonationTracker.ts
+++ b/src/services/segment/DonationTracker.ts
@@ -8,6 +8,25 @@ import {
 
 const analytics = getAnalytics();
 
+interface DonationAttributes {
+  email?: String;
+  title?: String;
+  firstName?: String;
+  projectOwnerId?: String;
+  slug?: String;
+  amount?: Number;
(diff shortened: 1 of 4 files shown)


src/services/segment/DonationTracker.ts
```diff
@@ -56,7 +76,13 @@ class DonationTracker {
       donationValueEth: this.donation.valueEth,
       verified: Boolean(this.project.verified),
       transakStatus: this.donation.transakStatus,
-    };
+    } as Partial<DonationAttributes>;
+
+    if (this.eventName === NOTIFICATIONS_EVENT_NAMES.DONATION_RECEIVED) {
+      delete donationAttributes.fromWalletAddress;
+    }
```
CarlosQ96, line 83: If I didn't declare a partial I couldn't remove the key.

mohammadranjbarz (approved):
Thanks @CarlosQ96  for finding the reason, LGTM

PR #725, opened 2022-11-16
Add onRamper webhook

CarlosQ96:
Onramper webhook. This flow occurs after the donation is done on the widget of onramper in the frontend.

Related to https://github.com/giveth/impact-graph#workspaces/all-devs-61644589b92edf0016f6999c/board

Documentation: https://docs.onramper.com/API-Reference/#webhooks

Tested with real data too. But I reproduced the same environment with the test key. So it's same functionality.


mohammadranjbarz (approved):
Thanks @CarlosQ96  for your clean code and refactoring udateDonation price peace of code, LGTM

PS: I just changed some little things and comments

PR #726, opened 2022-11-16
Empty webhook for onramper apikeys

CarlosQ96:


src/services/onramper/webhookHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function onramperWebhookHandler(request, response) {
+  try {
+    // just log to test stuff
+    logger.info(
+      'onramperWebhookHandler header',
+      request.headers['X-Onramper-Webhook-Signature'],
+    );
+    logger.info('onramperWebhookHandler body', JSON.stringify(request.body));
+    // logic is on another branch this is temporary
+    response.status(200).send('Not implemented');
```
mohammadranjbarz, line 15: @CarlosQ96  I just replaced console.logs with logger.info and added a title to logs ( when we don't add a title to  logs and just log the data it makes it hard in staging/production to know what was the thing that we have logged)

mohammadranjbarz (approved):
Thanks @CarlosQ96  LGTM

PR #731, opened 2022-11-21
F 1837 handle cancel projects in power boosting

mohammadranjbarz:
related to https://github.com/Giveth/giveth-dapps-v2/issues/1837


src/repositories/powerBoostingRepository.ts
```diff
@@ -64,8 +64,8 @@ export const findUsersWhoBoostedProject = async (
 };
 
 export const findPowerBoostings = async (params: {
-  take: number;
-  skip: number;
+  take?: number;
+  skip?: number;
```
CarlosQ96, line 68: Are you fetching them all in one go? wouldn't this list be way too huge in the future? maybe you should set a default take.

CarlosQ96 (approved):
LGTM Mohammad! you can merge this one. 

I see our graphql query has default values for "Take Argument" so we don't worry about someone fetching too many powerboostings from the findPowerBoostings function.

PR #749, opened 2022-12-02
Modify get donors and who liked projects query to exclude somethings

mohammadranjbarz:
https://github.com/Giveth/notification-center/issues/17

### DIFF
--- a/src/repositories/donationRepository.ts
+++ b/src/repositories/donationRepository.ts
@@ -1,6 +1,8 @@
 import { Project } from '../entities/project';
 import { Donation } from '../entities/donation';
 import { ResourcesTotalPerMonthAndYear } from '../resolvers/donationResolver';
+import { User } from '../entities/user';
+import { Reaction } from '../entities/reaction';
 
 export const createDonation = async (data: {
   amount: number;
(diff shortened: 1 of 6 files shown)


src/repositories/donationRepository.ts
```diff
@@ -74,16 +76,23 @@ export const findDonationById = async (
...
+      'reaction',
+      'reaction.projectId = project.id AND user.id = reaction.userId',
+    )
     .distinctOn(['user.walletAddress'])
     .select('LOWER(user.walletAddress) AS "walletAddress", user.email as email')
-    .where(`"projectId"=:projectId`, {
+    .where(`donation."projectId"=:projectId`, {
       projectId,
     })
+    .andWhere(`reaction.id IS NULL`)
```
CarlosQ96, line 95: Hmm not sure about this condition or the test it self. Will work on double checking this.

CarlosQ96 (commented):
Excluding the owner looks good too me, ill work on the exclude the person that liked in that moment. I think the test and condition is not clear... or maybe is the naming or I am confusing myself haha

src/repositories/donationRepository.ts
CarlosQ96: Should it be instead?
```
    .andWhere(`reaction.userId != donation.userId`) 
```

CarlosQ96: Or do you mean users who did not like the project?

mohammadranjbarz:
@CarlosQ96 
Thanks for your review, we need to deploy it now, so I merge it, but we will fix it in the future

Issue #750, opened 2022-12-06
Archive old power boosting related data


aminlatifi:
Power Boosting Snapshot and Power Balance Snapshot tables can become too big after taking many snapshots. This impacts system performance and sometimes functionalities since the DB queries may take a long time.

For instance, today staging was not responding because the boosting-related queries used to take a very long time and blocked the DB!

As power boosting snapshot and power balance snapshot tables' data are indexed and queried a lot, one solution is to make them lighter. Technically, computing projects' rank and future rank needs only data related to 2 rounds ago. Therefore, we can move the rest to archived equivalent ones. In this way, we won't lose data for future analyses. 

The point is that only power boosting snapshot data is essential to be kept somewhere since the power balance snapshot data can be retrieved again by impact graph. BTW, we can store both so far till sometime we suffer from the huge amount of data in Db.

aminlatifi:
So far, I deleted old data with
```sql
delete FROM public.power_boosting_snapshot
where "powerSnapshotId" in (
	select "snapshot"."id" from public.power_snapshot as "snapshot",
	public.power_round as "powerRound"
	where "snapshot"."roundNumber" < "powerRound"."round" - 3
);

delete FROM public.power_balance_snapshot
where "powerSnapshotId" in (
	select "snapshot"."id" from public.power_snapshot as "snapshot",
	public.power_round as "powerRound"
	where "snapshot"."roundNumber" < "powerRound"."round" - 3
);
```

aminlatifi:
Maybe something like this may work
https://stackoverflow.com/questions/2974057/move-data-from-one-table-to-another-postgresql-edition/8095147#8095147

aminlatifi:
Maybe the best way is to create another pg_cron job for archiving that runs once per round and inside of that run that query. Obviously, we must create a procedure call with all archiving logic and set it as the query is run by the cron job. @CarlosQ96 @mohammadranjbarz I like you guys once create a pg_cron job and see how is the procedure.

CarlosQ96:
I think this might work after looking at the docs, so we can periodically run this query inside the cronjob and clean up the main table. 

On staging tho, we should add additional logic, because of testing queries I did. So we should delete stuff in the historic table eventually too in staging. 

Because Staging is the one that grows insanely fast compared to prod. (unless we are not gonna query this historic data, but if we are we should clean it in staging or we will crash the server again).

aminlatifi:
@CarlosQ96 Let's go forward, 

[...]

PR #754, opened 2022-12-07
Add givPower historic Snapshots tables

CarlosQ96:
- https://github.com/Giveth/impact-graph/issues/750


migration/1670429143091-createGivPowerHistoricTablesProcedure.ts
```diff
@@ -0,0 +1,73 @@
...
+                        FROM "power_snapshot" AS "snapshot", "power_round" AS "powerRound"
+                        WHERE "snapshot"."roundNumber" < "powerRound"."round" - 3
+                    )
+                    RETURNING ps."id", ps."time", ps."blockNumber", ps."roundNumber", ps."synced"
+                )
+                INSERT INTO "power_snapshot_history"
+                SELECT se."id", se."time", se."blockNumber", se."roundNumber", se."synced"
+                FROM snapshot_entity AS se
+            $BODY$;
+        `);
```
aminlatifi: I think it would be better to aggregate all these actions in a single procedure. I mean instead of three procedures, only define one which does the whole. We can do these three as a transaction also to be sure all is done or nothing will happen
https://www.postgresql.org/docs/current/tutorial-transactions.html

The main reason for this is that we will have to define separate pg_cron job and clutter it's table.

src/repositories/dbCronRepository.ts
```diff
@@ -33,6 +40,70 @@ export const schedulePowerBoostingSnapshot = async (
...
+      GRANT USAGE ON SCHEMA CRON TO POSTGRES;
+
+      SELECT CRON.SCHEDULE(
+        '${POWER_SNAPSHOT_HISTORY_TASK_NAME}',
+        '${cronJobExpression}',
+        $$CALL public."TAKE_POWER_BOOSTING_SNAPSHOT"()$$
+      );
+  `);
+};
+
```
aminlatifi: As mentioned in another comment, let's have only one procedure and one expression.

src/repositories/dbCronRepository.ts
```diff
@@ -1,9 +1,16 @@
...
 
 export const POWER_BOOSTING_SNAPSHOT_TASK_NAME =
   'take givpower boosting snapshot';
+export const POWER_BOOSTING_SNAPSHOT_HISTORY_TASK_NAME =
+  'take givpower boosting snapshot history';
+export const POWER_BALANCE_SNAPSHOT_HISTORY_TASK_NAME =
+  'take givpower balance snapshot history';
+export const POWER_SNAPSHOT_HISTORY_TASK_NAME =
+  'take givpower snapshot history';
+
```
aminlatifi: As another comment, let's have one.

aminlatifi (changes requested):
Thanks @CarlosQ96 , mostly good. Just a change and then we can go for creating cron jobs

migration/1670429143091-createGivPowerHistoricTablesProcedure.ts
CarlosQ96: Updated them into one! 
The only thing I couldn't do is put a BEGIN END keywords.
It kept saying the logs that You cannot put Begin inside a procedure.

src/repositories/dbCronRepository.ts
CarlosQ96: Done

src/repositories/dbCronRepository.ts
CarlosQ96: Done

migration/1670429143091-createGivPowerHistoricTablesProcedure.ts
CarlosQ96:
```
BEGIN is not allowed in an SQL function
```
Couldn't find a way to make it work.

docs/powerBoosting.md
```diff
@@ -229,6 +229,16 @@ limit 5
 |82       |155876.872369|4        |1779 |
 |223531   |129154.467198|5        |1779 |
 
+#### Power Boosting Snapshot, Power Balance Snapshot and Power Snapshot Historic tables
+
+Eventually our snapshot tables will be filled with a lot of information,
+this degrades the performance of the queries and the server. To solve this
+a cronjob will run periodically set by `ARCHIVE_POWER_BOOSTING_OLD_SNAPSHOT_DATA_CRONJOB_EXPRESSION` variable.
+
+This cronjob executes the procedure `ARCHIVE_POWER_BOOSTING_OLD_SNAPSHOT_DATA` that removes the data from `power_boosting_snapshot`, `power_balance_snapshot`, `power_snapshot` that is older than 3 rounds from the current powerRound.
```
aminlatifi:
```suggestion
This cronjob executes the procedure `ARCHIVE_POWER_BOOSTING_OLD_SNAPSHOT_DATA` that removes the data from `power_boosting_snapshot`, `power_balance_snapshot`, `power_snapshot` that is older than 2 rounds from the current powerRound.
```

aminlatifi (approved):

aminlatifi:
Thanks @CarlosQ96 , great job.

PR #758, opened 2022-12-12
Hotfix: Exclude failed donations from totals

CarlosQ96:
Related to: https://github.com/Giveth/giveth-dapps-v2/issues/1922#issuecomment-1345409267


src/resolvers/donationResolver.ts
```diff
@@ -622,13 +622,6 @@ export class DonationResolver {
         amount,
       );
 
-      // After updating, recalculate user total donated and owner total received
-      await updateUserTotalDonated(donorUser.id);
-
-      // After updating price we update totalDonations
-      await updateTotalDonationsOfProject(projectId);
-      await updateUserTotalReceived(Number(project.admin));
-
```
CarlosQ96, line 631: Since failed or pending donations are not ready. We shouldn't add them to the total until they are verified.

CarlosQ96:
I don't think we should write a migration to fix all projects, I don't think there are many failed donations, rare case in my opinion. We can fix them manually by donating to them. Like the project Griff mentioned.
https://giveth.io/project/relocating-web3-iranians

But from here on out after this PR, this issue will not happen again.

CarlosQ96:
@mohammadranjbarz Let me know if you agree.

CarlosQ96:
@MoeNick also optimized the queries from 6 to 3. Which is more efficient.

mohammadranjbarz:
> @mohammadranjbarz Let me know if you agree.

Yeah Carlos, I agree with you

mohammadranjbarz (approved):
Thanks @CarlosQ96 
LGTM

Issue #729, opened 2022-11-17
Capture status change events in edit project page in admin bro


mohammadranjbarz:
When user changes **verified**, **listed** or changing status of projects to one **activated** , **deactivated** and **cancelled** we should send notification (and email if it's needed) 

<img width="1439" alt="Screen Shot 1401-08-26 at 12 05 03" src="https://user-images.githubusercontent.com/9850545/202456723-40929cdd-1ca1-41b3-8fbd-ca1ad832632d.png">

mohammadranjbarz:
@maryjaf  @MoeNick 
It's on staging now

maryjaf:
> @maryjaf @MoeNick It's on staging now

I think this problem is not completely solved @mohammadranjbarz 
Because when I change the project from the edit page to the listed status, I do not receive a notification
But when I make this change from the list, I receive this notification: Nice! Your [testfilterr](https://staging.giveth.io/project/testfilterr) project is listed.


When I unlist the project from the edit page, I get a notification as a who that like this project not project owner (i liked my project): The [testfilterr](https://staging.giveth.io/project/testfilterr) that you liked is no longer visible. 
and when unlist project from list i receive two notification as who liked project and project owner :
The [testfilterr](https://staging.giveth.io/project/testfilterr) that you liked is no longer visible.
6 seconds ago
Your [testfilterr](https://staging.giveth.io/project/testfilterr) project is no longer visible.
6 seconds ago

----
And in general, now we don't get notified based on the status change, if I just click, I get notified
For example, the status of my project is listed, if the admin user hits the list again, even though the status has not changed, a notification will be sent.
is that ok? @mohammadranjbarz @MoeNick

mohammadranjbarz:
> > @maryjaf @MoeNick It's on staging now
> 
> I think this problem is not completely solved @mohammadranjbarz Because when I change the project from the edit page to the listed status, I do not receive a notification But when I make this change from the list, I receive this notification: Nice! Your [testfilterr](https://staging.giveth.io/project/testfilterr) project is listed.
> 
[...]

PR #776, opened 2022-12-18
Fix notification events not being sent

CarlosQ96:
Related to: https://github.com/Giveth/impact-graph/issues/729

Seems it was a class type error, I wasn't able to fix, so I took the methods OUT of the class.


mohammadranjbarz (approved):
Thanks @CarlosQ96  Seems good to me, we had not used functions outside the module so I removed `export` from those functions declaration

PR #782, opened 2022-12-21
Hotfix cache queries for performance

CarlosQ96:
Used Threads Library to off load hashing of cache keys.
Added cache for count and results of projects filters queries.
Moved main filters query to projectRepository (this is because queries share state, so instantiating different query builders avoid wrong results when using cache).


CarlosQ96:
Still work in progress, I am rewriting a little the queries, but the cache broke the tests.

CarlosQ96:
To debug you can add ```DEBUG=threads:*``` to any command like npm start or npm run test

src/repositories/projectRepository.ts
```diff
@@ -28,6 +38,85 @@ export const findProjectById = (
     .getOne();
 };
 
+// return query without execution
```
CarlosQ96, line 38: Moved Filter query to the repository so I could reuse it for count query and results query.
When using the same querybuilder object, the cache was throwing errors.

src/server/bootstrap.ts
```diff
@@ -135,6 +148,12 @@ export async function bootstrap() {
       }
     }
 
+    // instantiate pool once and pass as context
```
CarlosQ96, line 124: If you separate it on another module ```export const pool``` throws tons of errors.
If you put it at the root of the project resolver, thread instantiation gets stuck.

This is the only way I found it worked. Instantiate it in the express server and pass it as context to the graphql resolvers.

src/utils/utils.ts
```diff
@@ -1,9 +1,32 @@
...
+  mainCategory?: string;
+  filters?: FilterField[];
+  sortingBy?: SortingField;
+  connectedWalletUserId?: number;
+  userId?: number;
+  suffix?: string;
+}) => {
+  const orderedArgs = stringify(args);
+
+  return await hashMD5(orderedArgs, { algorithm: 'md5' });
```
aminlatifi: `object-hash` has the parameter
```
unorderedObjects {true|false} Sort objects before hashing, i.e. make hash({ x: 1, y: 2 }) === hash({ y: 2, x: 1 }). default: true
```
As mentioned, its default value is true. As we order and stringify by `json-stable-stringify`, it's not efficient to order the object twice. We can either set that parameter as false, or skip `json-stable-stringify`

src/server/bootstrap.ts
```diff
@@ -66,6 +67,18 @@ const cors = require('cors');
 
 Resource.validate = validate;
 
+const options = {
+  concurrency: Number(
+    process.env.PROJECT_FILTERS_THREADS_POOL_CONCURRENCY || 1,
+  ),
+  maxQueuedJobs: Number(
+    process.env.PROJECT_FILTERS_THREADS_POOL_MAX_QUEUED || 4,
```
aminlatifi: Why do we override its default unlimited value? If we get many queries simultaneously it will cause issue.

src/resolvers/projectResolver.ts
```diff
@@ -567,221 +573,88 @@ export class ProjectResolver {
...
-        'projectPower.totalPower',
-        'projectPower.powerRank',
-        'projectPower.round',
-      ])
-      .where(
-        `project.statusId = ${ProjStatus.active} AND project.listed = true`,
-      );
+      user,
+    );
+    const filtersResultHashTask = projectsFiltersThreadPool.queue(hasher =>
```
aminlatifi: Why we run two queries, one for count and one for fetching the projects?

src/resolvers/projectResolver.ts
```diff
@@ -567,221 +573,88 @@ export class ProjectResolver {
...
+    const projectsQuery = filterProjectsQuery(
+      limit,
+      skip,
+      searchTerm,
+      category,
+      mainCategory,
+      filters,
+      sortingBy,
+      connectedWalletUserId,
+      user,
```
aminlatifi: I think we must skip the user from the key. I mean fetch the projects independent of the user, then in a separate query load the user reactions and fill projects! Of course, db does the joining better than us, but 
- For every user cache value is useless
- User may update the reaction (like/unlike a project)! and the cache will not be invalidated immediately

aminlatifi (changes requested):

src/utils/utils.ts
CarlosQ96: Ok awesome, I missed that detail, I think in this case we should remove the json-stringify, less packages in our project the better.

src/server/bootstrap.ts
CarlosQ96: I was probably thinking unlimited stuff is bad hahaha

but I will remove the limit as we don't want now to break the query when we have high loads of people, we want performance not new issues.

src/resolvers/projectResolver.ts
CarlosQ96: I will test GetManyAndCount() query with 1 cache and let you know. I think I avoided it because I was getting errors with the cache. Will report back on this comment.

src/resolvers/projectResolver.ts
CarlosQ96: As we decided by DM, I will remove the reaction of the logged in user portion from the cache and the userId from the cache key. 
Query the reaction separately and then in a Worker merge the data.

aminlatifi:
Made[ this improvement PR](https://github.com/Giveth/impact-graph/pull/810)

We have a build issue with typescript v3.
https://github.com/andywer/threads.js/issues/359#issuecomment-831966607
So we are blocked by other dependency upgrade PRs #808 and #630

aminlatifi:
Merged https://github.com/Giveth/impact-graph/pull/810 as tests passed.
The major idea behind improvements:
1. Merging reaction complexity improved, from O(n^2) to O(n)
2. Relay on json-stable-stringify, since it support nested object sorting. object-hash does only first level key sorting

aminlatifi:
@CarlosQ96 I merged staging into it, and everything seems okay.
Please resolve the above comments, and then we can merge this. Do you think we need to configure the staging env? or the default values are fine?

CarlosQ96:
Hey Amin for staging I think the default values are enough.

Issue #775, opened 2022-12-15
Create notification for when someone boosts a project you own with GIVpower


laurenluz:
This notification is for verified project owners. When someone boosts their project, they should be notified. Copy:

Someone boosted your project {projectname}!

@CarlosQ96 @mohammadranjbarz

mohammadranjbarz:
@maryjaf 
It's on staging you can test it now

mohammadranjbarz:
> @maryjaf It's on staging you can test it now

@maryjaf  We send notifications one per project-booster
it means when someone boost a project we send notificaiton to project owner, but if user remove the boosted and boost that project again we would not send another notification

maryjaf:
> > @maryjaf It's on staging you can test it now
> 
> @maryjaf We send notifications one per project-booster it means when someone boost a project we send notificaiton to project owner, but if user remove the boosted and boost that project again we would not send another notification

I tested this scenario but i didn't receive notification

mohammadranjbarz:
> > > @maryjaf It's on staging you can test it now
> > 
> > 
> > @maryjaf We send notifications one per project-booster it means when someone boost a project we send notificaiton to project owner, but if user remove the boosted and boost that project again we would not send another notification
> 
> I tested this scenario but i didn't receive notification

I had forgotten to merge notificaiton-center PR, now I did it, can you test again?

maryjaf:
Thanks @mohammadranjbarz Now when some one boost my project i receive notification like below pic

<img width="774" alt="image" src="https://user-images.githubusercontent.com/111529185/210377497-e11935bf-fadc-4910-b577-5fddf0f63dea.png">

PR #785, opened 2022-12-26
Send notification to project owner when someone boosts a project

mohammadranjbarz:
related to https://github.com/Giveth/impact-graph/issues/775

### DIFF
--- a/src/adapters/notifications/NotificationCenterAdapter.ts
+++ b/src/adapters/notifications/NotificationCenterAdapter.ts
@@ -311,31 +311,22 @@ export class NotificationCenterAdapter implements NotificationAdapterInterface {
   }
 
   async projectBoosted(params: {
-    project: Project;
+    projectId: number;
     userId: number;
   }): Promise<void> {
-    const { project, userId } = params;
-    const projectOwner = project.adminUser as User;
-
-    return sendProjectRelatedNotification({
-      project,
+    const { projectId, userId } = params;
+    const project = (await findProjectById(projectId)) as Project;
+    sendProjectRelatedNotificationsQueue.add({
+      project: project as Project,
+      eventName: NOTIFICATIONS_EVENT_NAMES.PROJECT_BOOSTED,
 
       // With adding trackId to notification, notification-center would not create new notification
       // If there is already a notification with this trackId in DB
       trackId: generateTrackId({
         userId,
-        projectId: project.id,
+        projectId: project?.id as number,
         action: 'boostProject',
       }),
-      eventName: NOTIFICATIONS_EVENT_NAMES.PROJECT_BOOSTED,
-      sendEmail: true,
-      segment: {
-        analyticsUserId: projectOwner.segmentUserId(),
-        anonymousId: projectOwner.segmentUserId(),
-        payload: getSegmentProjectAttributes({
-          project,
-        }),
-      },
     });
   }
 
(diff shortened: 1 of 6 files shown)


src/resolvers/reactionResolver.ts
```diff
@@ -224,6 +224,7 @@ export class ReactionResolver {
       const userWhoLiked = (await findUserById(user.userId)) as User;
```
aminlatifi: We are doing an unnecessary job here! load user data from id, and only use the id value

src/adapters/notifications/NotificationCenterAdapter.ts
```diff
@@ -292,6 +310,59 @@ export class NotificationCenterAdapter implements NotificationAdapterInterface {
     });
   }
 
+  async projectBoosted(params: {
+    project: Project;
```
aminlatifi: let's pass projectId here and allow this method itself fetch the project info

src/adapters/notifications/NotificationCenterAdapter.ts
```diff
@@ -292,6 +310,59 @@ export class NotificationCenterAdapter implements NotificationAdapterInterface {
...
+
+        // With adding trackId to notification, notification-center would not create new notification
+        // If there is already a notification with this trackId in DB
+        trackId: generateTrackId({
+          userId,
+          projectId: project?.id as number,
+          action: 'boostProject',
+        }),
+      });
+    });
```
aminlatifi: let's instead of duplicating code here, reuse `projectBoosted` method with changes I mentioned above (passing projectId instead of project)

src/resolvers/powerBoostingResolver.ts
```diff
@@ -132,13 +141,21 @@ export class PowerBoostingResolver {
...
     }
 
-    return setSingleBoosting({ userId: user.userId, projectId, percentage });
+    const result = await setSingleBoosting({
+      userId,
+      projectId,
+      percentage,
+    });
+    const project = (await findProjectById(projectId)) as Project;
+    await getNotificationAdapter().projectBoosted({ project, userId });
```
aminlatifi: I guess it's not very cool to fetch project information here, we can rely on projectBoosted method to do it.
Also, and importantly, I think it's not cool to block API for sending a notification. We can do it in parallel, but if any error happens we must be sure will see it (collecting by sentry or anything else).
Technically, remove that `await` and handle it in `.catch` for example.

aminlatifi (changes requested):

src/resolvers/reactionResolver.ts
mohammadranjbarz: Haaahaaa !!

you are goddamn right

src/adapters/notifications/NotificationCenterAdapter.ts
mohammadranjbarz: Done

src/adapters/notifications/NotificationCenterAdapter.ts
mohammadranjbarz: Yeah  thanks for your suggestion it's cleaner, Done.

src/resolvers/powerBoostingResolver.ts
mohammadranjbarz: Done, I didn't know what to do to not get a warning from IDE without disabling eslint, nice idea

mohammadranjbarz:
@aminlatifi  I did all your requested changes, please review it again

PR #788, opened 2022-12-29
F 1974 return all power boostings in response

mohammadranjbarz:
related to  https://github.com/Giveth/giveth-dapps-v2/issues/1974


mohammadranjbarz:
@aminlatifi 
After merging to staging and testing it we can create a hotfix PR from `f_1974_return_all_power_boostings_in_response` to `master`

aminlatifi:
@mohammadranjbarz I don't think it's a good idea to remove pagination from backend

Issue #780, opened 2022-12-19
Broadcast from Adminbro notifications


CarlosQ96:
### Write a service in adminbro and notificationcenter that allows broadcasting the notifications to all users. 

This service should receive an action name (broadcastType): 'broadcast', 'broadcast to group' and send the notification according to the received content.
The loop AND SQL query for creating it the notification should be done in the notification center.

### STEPS

**_1. Notification Center:  /broadcast_**
Body: content, userIds?, broadcastType, title

Executes:
-   SQL QUERY that creates the notification in the db for all users or selected users in 1 query ONLY. (no loop here)
- LOOP for enqueuing segment events in a bullQueue and send them.

**_2. Adminbro Custom Page and entity: BroadcastNotification_**
This entity will allow creating broadcast messages for reusing anytime. The model should have:
_broadcastType, content, userIds (optional), title, createdAt, updatedAt._ 

And a custom button called **Send Broadcast** that will call the /broadcast endpoint in the notification center, sending the params of the model.

**_3. Broadcast history entity_**
To keep track of the broadcasts we sent, we should fill this table everytime the button **Send Broadcast** is pressed.

CarlosQ96:
Caveats or issues: 
We can do it without filtering by users, this can be a future issue. Just make it so it broadcasts to everyone. 
I say this because making good UX user filterer in a custom form in adminbro is quite time consuming because of limitation.

@MoeNick @mohammadranjbarz @aminlatifi  let me know what you think.

CarlosQ96:
Since we have the adminbro in the impact graph and the admin it's entity dependent, It's inevitable to create the Broadcast notification entity in the impactgraph.

It's somewhat redundant. but if you have a better idea, let me know.

mohammadranjbarz:
Nice, thanks for clarifying this issue, I agree with what you have written, just it should be good that we implement this feature in admin-bro for specific roles, for instance `Operator` role should not have access to send broadcast to all users

mohammadranjbarz:
@maryjaf 
Now this feature is on staging , you can send broadcast notification from 
[https:](https://serve.giveth.io/admin and see the notifications in 
https://notification-center-13.vercel.app/notifications

**PS** You need to know basic things about Markdown for sending notifications, if you needed any help, you can ask Me or @MohammadPCh

maryjaf:
When I  created new broadcast notification I see the notification like below pic for a second and after it has been fixed and the correct time is shown
[...]

PR #793, opened 2023-01-02
Implement adminBro page for broadcast notifications

mohammadranjbarz:
related to #780


mohammadranjbarz:
It uses this notificationType

https://github.com/Giveth/notification-center/blob/e040e6baf5f19b2df0b7fe5d5b4710d02aa30780/migrations/1660716115917-seedNotificationType.ts#L100-L132

mohammadranjbarz:
You can create broadcast notifications from this page

Screen Shot 1401-10-13 at 17 03 27

src/adapters/notifications/MockNotificationAdapter.ts
```diff
@@ -201,4 +204,12 @@ export class MockNotificationAdapter implements NotificationAdapterInterface {
     });
     return Promise.resolve(undefined);
   }
+
+  broadcastNotification(
+    params: BroadCastNotificationInputParams,
+  ): Promise<void> {
+    // throw new Error('test adminBro')
```
CarlosQ96: Remove comment

src/adapters/notifications/NotificationCenterAdapter.ts
```diff
@@ -819,4 +692,259 @@ export class NotificationCenterAdapter implements NotificationAdapterInterface {
...
+    while (!allUserFetched) {
+      const { users } = await findAllUsers({ take, skip });
+      if (users.length === 0) {
+        allUserFetched = true;
+        break;
+      }
+      skip += users.length;
+      const queueData: SendBatchNotificationBody = [];
+      for (const user of users) {
+        queueData.push({
```
CarlosQ96, line 713: I don't think this is necessary here @mohammadranjbarz. We can offload all the work to the notification center, because this would spam the notification center with many requests, it might hit a rate limit (not sure if we have with nginx).

Remember the notification center also have all users registered with their wallet address. 

Simply send a request with the type: "Broadcast" and the data, and handle the segment loop in the notification center in multiple queues. (I think we would need to implement the queue system in notification center that we don't currently have).

For the notifications to insert in the DB, I think we can insert all the values in one single query in the notification center.
https://stackoverflow.com/questions/24344783/insert-into-for-all-users-who-are-in-another-table

So to summarize:
1. Send 1 request with the type of notification **broadcast** and the data. (this enqueues the segment events enqueuer looper).
2. Run the query to insert all values to all users in a single query.  (I can help with this)
3. The segment events enqueuer, loops and enqueues each individual user notification to be sent with segment (in its own queue).

Its a lot more work, but its more efficient.

CarlosQ96: BUT!!!!!!!!!! ***************************

If we are short in time, we can use this to start, but the loop do it inside a Queue not in the main thread. Because We had issues before where doing heavy lifting in an endpoint (like not paginating) and the backend slows down.

(this is my limited knowledge in js, for instance since it's single threaded does bull spawn new threads? or uses the same one?) (if its the same thread then doing the loop in a queue won't do anything performance wise, and you can ignore this last suggestion and merge).

CarlosQ96 (changes requested):
Hey Mohammad I think the code is good and works!!!

 I have some important suggestions for performance in the loop for sending requests to the notification center. Let me know what you think and we can decide the best move forward.

src/adapters/notifications/NotificationCenterAdapter.ts
mohammadranjbarz: Thanks @CarlosQ96  for your comments, I remember you had mentioned on the specs of issue as well

During the holidays I and Amin talked about that and we came to this conclusion:
1. We don't have all of the users in notification-center, we just have the user's info that we already sent them at least 1 notification
2. We don't have emails of users there
3. We implemented https://staging.notification.giveth.io/docs/#/Notification/SendBulkNotification s we can send 100 notifications in one request, so if all of our users are 2,000 we can send notification for them with only 20 requests  and also we queue the requests so it would not make lots of load on notification-center
4. probably don't use this feature too much, maybe just a few times a month (maybe less) so I think a minimal implementation can be enough for that and it would not make lots of load on impact-graph

So I think yeah current implementation might be good at the beginning, do you agree with me?

src/adapters/notifications/MockNotificationAdapter.ts
mohammadranjbarz: Done

mohammadranjbarz:
@CarlosQ96 
I talked with @MoeNick  and we have to do something before merging so I made this Pr draft:
1. Add email of admin that sent notification
2. instead of sending multiple fields just send one markdown
3. preview the markdown in admin bro if it's possible

I would work on these 3 items

mohammadranjbarz:
Screen Shot 1401-10-18 at 16 27 38
Now we changed it to get markdown as input and we show the preview of it in admin-bro and convert it to HTML and send it to notificaiton-center

@CarlosQ96

mohammadranjbarz:
@CarlosQ96  It's ready for review now

migration/1672836674875-createTestAdminUser.ts
```diff
@@ -0,0 +1,25 @@
...
+                INSERT INTO public.user (email, "walletAddress", role,"loginType", name,"encryptedPassword") 
+                VALUES('test-admin@giveth.io', '0x66F014F90cdE78130e143b00F8Bd3F77c73bb0Bc', 'admin','wallet', 'test', '${hash}');
+              `);
+  }
+
+  async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+               DELETE FROM public.user where email='test-admin@giveth.io'`);
+  }
+}
```
mohammadranjbarz, line 25: I added this migration file to create a test admin user when someone run migrations on local machine and wants to contribute

CarlosQ96: Perfect

src/server/adminBro.ts
```diff
@@ -162,23 +169,12 @@ export const getAdminBroRouter = async () => {
...
-          }
-          return false;
-        } catch (e) {
-          logger.error({ e });
+      authenticate: async (email, password): Promise<User | boolean> => {
+        const isPasswordCorrect = await checkAdminPassword({ email, password });
+        if (!isPasswordCorrect) {
           return false;
         }
+        return (await findAdminUserByEmail(email)) as User;
```
CarlosQ96: While I agree with the refactor, we are doing the same query twice. We should only query the user once.
Maybe the checkAdminPassword can return the user and we return it here.

CarlosQ96 (changes requested):
Great work Mohammad! I think we can improve the auth handler a bit. Ill push this change and Approve

CarlosQ96 (approved):
Approved. There seems to be a network issue tho. Once that is fixed we can merge. improved auth handler

Issue #790, opened 2023-01-01
Remove smart contract condition in WalletAdressIsValid


MoeNick:
> This will allow us to avoid the issues around contracts... we can let contracts receive funds. The big question is if we allow them to only accept some currencies or not.

@CarlosQ96 we have to remove the "smart contract" condition in WalletAddressIsValid. 
![image](https://user-images.githubusercontent.com/80677512/210176605-6cfa0900-2638-4435-bf99-83e945ffccc1.png)

_Originally posted by @MoeNick in https://github.com/Giveth/giveth-dapps-v2/issues/106#issuecomment-1368471405_

mohammadranjbarz:
@maryjaf  @MoeNick 
It's on staging and ready for test

MoeNick:
Test Scenario:
user will be allowed to add a smart contract as a recipient address and won't see the previous error msg.

maryjaf:
Thanks @MoeNick @mohammadranjbarz
I tested it and create new project with receiving fund address : Eth address 0xc011a73ee8576fb46f5e1c5751ca3b9fe0af2a6f is a smart contract, I did not get the previous error and it was recorded and I made a successful donation to this address

PR #794, opened 2023-01-03
Comment out isSmartContract logic

CarlosQ96:
Related to: https://github.com/Giveth/impact-graph/issues/790

Mohammad I didnt delete maybe this is added later who knows (As this is a request for a specific client I believe).  Tests could be usefull.


mohammadranjbarz (changes requested):
Thanks @CarlosQ96  I think code is good, but I have a suggestion , instead of commenting test cases let's change the assertion , for instance instead of expecting get error when sending smart contract address now we expect it to be successful, in that case this issue can be covered with our automatic tests, what do you think?

CarlosQ96:
Thanks Mohammad Ill do it.

PR #818, opened 2023-01-23
Feat/recent donations + cache homepage queries

aminlatifi:
https://github.com/Giveth/giveth-dapps-v2/issues/1992

### DIFF
--- a/src/repositories/donationRepository.ts
+++ b/src/repositories/donationRepository.ts
@@ -112,6 +112,11 @@ export const donationsTotalAmountPerDateRange = async (
   }
   const donationsUsdAmount = await query.getRawOne();
 
+  query.cache(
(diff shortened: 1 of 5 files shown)


CarlosQ96 (approved):
Thanks Amin Loooks Good to me!

I would just ask if we should instead the cache configurable by an env variables instead of hardcoding, what do you think?

aminlatifi:
> Thanks Amin Loooks Good to me!
> 
> I would just ask if we should instead the cache configurable by an env variables instead of hardcoding, what do you think?

Thanks @CarlosQ96. You are generally right, however, these APIs will be called to fetch historic data.  It's not needed to be so accurate and updated.
BTW, our config file is a flat `.env` type file which is already cluttered and big. I think we must go later for a better approach supporting optional, default value, grouping and hierarchy.

PR #826, opened 2023-01-25
Upgrade Apollo Server

aminlatifi:


src/types/ApolloContext.ts
```diff
@@ -0,0 +1,14 @@
+import { Response } from 'express';
+import { Pool, ModuleThread } from 'threads';
+import { ProjectResolverWorker } from '../workers/projectsResolverWorker';
+import { JwtVerifiedUser } from '../services/authorizationServices';
+
+export interface ContextRequest {
+  user: JwtVerifiedUser;
+  auth: any;
```
CarlosQ96: Should we avoid using any? I think we can specify it like this.
```
export interface AuthContextRequest {
   token?: string;
   error?: any;
}
```
The error is kind of hard to define..... as it depends 100% on which graphql function is called with what message. Maybe something General

src/server/bootstrap.ts
```diff
@@ -122,48 +126,30 @@ export async function bootstrap() {
...
-        try {
-          if (!req) {
-            return null;
-          }
+    const apolloServerPlugins = [
+      process.env.DISABLE_APOLLO_PLAYGROUND !== 'true'
+        ? ApolloServerPluginLandingPageGraphQLPlayground({
+            endpoint: '/graphql',
+          })
+        : ApolloServerPluginLandingPageDisabled(),
```
CarlosQ96, line 140: This plugin sets the following page:
image

CarlosQ96: Compared to simply disabling it like we had it before:
image

CarlosQ96: Should we leave it like that or simple return some form of 404 not found like before? 

I don't think it's that important but I think a 404 is more appropriated. A disabled route shouldn't reveal any data.

CarlosQ96 (approved):
Hey Amin I tested:

Some queries manually in the graph playground everything seems (ok), all tests seem (ok), I saw you fixed a bug with the user context (ok),  adminbro seems unnafected (ok), thread pools initialized correctly when debugging them (ok) 

I think we can proceeed to test in staging.

I just left some low priority concerns in comments.

src/server/bootstrap.ts
aminlatifi: Good point, I don't know how to solve it in a short time. Let's do it in another issue.

aminlatifi:
Thanks @CarlosQ96

mohammadranjbarz (approved):
Thanks @aminlatifi 
I just make the dependencies'versions fixed

Issue #806, opened 2023-01-15
Campaigns backend support


MoeNick:
As a comms team member, I want to define a custom banner (or video) on HompePage plus a filter of related projects, so the users can easily find the Highlighted or Featured projects and have a clue when they land on or homepage following the campaigns lead on digital channels like Twitter. 

**Suggested task list:**

1. Analysis and data schema design
2. Implementing the services for creating campaigns, editing, viewing a list of active campaigns and relating campaigns to projects, and default sorting in Admin Bro
3. Implement AdminBro changes
4. Services for filtering [ Featured Projects](https://github.com/Giveth/giveth-dapps-v2/issues/1994) 
5. Services for [custom banners](https://github.com/Giveth/giveth-dapps-v2/issues/1995) block
6. list of Default filters for not showing all campaigns in the projects filter dropdown box

**Details:**
 _#1 data schema design_
We need an entity called Campaigns. admin can view edit and define new campaigns in admin bro and relate it to different projects, with default sorting, these campaigns have different data fields like:

- isActive
- Name: simple string for custom naming
- Title: Shown on the homepage
- Subtitle: shown on the home page
- Hashtags: array of hashtags which first one is the key hashtag of this campaign
- mainMedia: IPFS link to a movie or a picture (guide on size and dimension should get clarified by the front end and get validated on admin bro
- related projects: it could be an array or all projects (PS: this can be handled by a tag and have tags array in the projects entity instead of this entity
- DefauttOrder: current sorting or other sorts
- isAmongDefaultFilters: this shows whether this filter is available on the front-end filter box or no
- homepage Order: we show up to 3 active campaign on the HomePage, this shows which come first
- landingLink

_#2 adminBro services:_

- View list f campaigns, default view: name, creator, isActive, createDate - 
- Sort and filter by about fields
- Create campaigns with the above fields
- edit
- (I'm not sure about deleting it could be making to inactive

_#3 Implement AdminBro_

_#4 Services for filtering_ - please see the details in  [ Featured Projects](https://github.com/Giveth/giveth-dapps-v2/issues/1994) - up to 9 project with default sorting will appear here

_#5  Services for [custom banners](https://github.com/Giveth/giveth-dapps-v2/issues/1995) block_ ,please see the link for more detail  - up to 3 active campaigns will show up here. 

_#6 list of Default filters for not showing all campaigns in the projects filter dropdown box_
[...]

PR #840, opened 2023-02-01
Add campaign entity

mohammadranjbarz:
related to #806


src/repositories/projectRepository.ts
```diff
@@ -289,3 +305,20 @@ export const makeProjectListed = async (id: number): Promise<void> => {
...
 };
+
+export const findProjectsBySlugArray = async (
+  slugArray: string[],
+): Promise<Project[]> => {
+  // TODO should refactor it and convert it to single query
+  const projects: Project[] = [];
+  const result = await Promise.all(
+    slugArray.map(slug => findProjectBySlug(slug)),
+  );
```
CarlosQ96, line 329: For this we can us the ```IN```operator in sql.
project.slug IN (:...slugs)

But will need to investigate how to do it with the slughistory part.

CarlosQ96 (approved):
I think we can proceed to testing in staging. We could improve some loops with complex joins but at this point I am not sure if typeorm is efficient enough with those as we are having problems with the projects query.

So LGTM lets merge. And maybe we can investigate the joins issue later.

Issue #842, opened 2023-02-01
Dont try to verify pending donations that has been made long time ago 


mohammadranjbarz:
We should set some value in config to don't check donations before that time, I think just checking pending donations of last 48 hours can be good, and we leave donations beore that as `pending` ( don't change it to `failed`) maybe later we we want to do something about them so it's better to have them as they are

maryjaf:
I think it's better to talk about how to test this

MoeNick:
@maryjaf donate with a very low gas fee that you sure it never get verified, check the donation before the _configuredDate_ and after that, before it should show pending and after that it should be null. Am I right @mohammadranjbarz ?

maryjaf:
Thanks @MoeNick 
Are the changes deployed on the staging environment? @mohammadranjbarz

mohammadranjbarz:
@maryjaf  @MoeNick  It would be seen in the donor's donations page, but not in project's page
It was the same before this change, all the thing that we do it's not trying to verify them and calling lots of requests to chain, so this issue is more technical, Mary can't test them ( it was part of technical debts)

@maryjaf  You can just make some donations and make sure they got verified, so if our success scenario is ok you don't need to test this issue

PR #843, opened 2023-02-01
Set expiration for donations to just verify them in that period of time

mohammadranjbarz:
related to #842 

We need to add `DONATION_VERIFICAITON_EXPIRATION_HOURS` to .env after merging that


aminlatifi (approved):

aminlatifi:
Thanks @mohammadranjbarz

PR #857, opened 2023-02-06
Removed unused dependencies 

aminlatifi:


CarlosQ96 (approved):
LGTM we don't use node-fetch at all or object-path

aminlatifi:
@CarlosQ96 I removed some others

CarlosQ96 (approved):
Tested adminbro, server and tests, everything seems in order and postgresql working too.

PR #861, opened 2023-02-13
Just count verified donations in user's donations count

mohammadranjbarz:
related to Giveth/giveth-dapps-v2#1902


CarlosQ96 (approved):
LGTM thanks!

PR #864, opened 2023-02-14
Used redis for typeorm caching

aminlatifi:


CarlosQ96 (approved):
Was testing queries in the graphql playground, seems good no issues with the queries.

Issue #816, opened 2023-01-20
Decrease get all projects request size


MohammadPCh:
each time we fetch the 15 projects to show on our projects page the response is about 2MB:
![image](https://user-images.githubusercontent.com/17097576/213704437-d96c2012-0b20-4996-b66d-e7a1571897dc.png)

the root cause of this huge size is the project's description, but we want to show this information in the project card:
![image](https://user-images.githubusercontent.com/17097576/213705012-ca124da4-04d2-4a11-bb56-3d2768e46f97.png)
as you can see we only need plain text (not HTML) and a maximum of 30 words.
if we remove the description from the response you will see the size decreases a lot:
![image](https://user-images.githubusercontent.com/17097576/213705813-4ef38bfb-cc0b-4ad6-ad76-3149cd416c94.png)
I suggest having a new attribute for our projects called summary. we can easily produce it in the backend with this package:
https://www.npmjs.com/package/html-to-text
@aminlatifi @MoeNick

aminlatifi:
I like this typeorm listener https://typeorm.io/listeners-and-subscribers#beforeupdate
@mohammadranjbarz @CarlosQ96 Do you think we can rely on it?

MohammadPCh:
@aminlatifi I think it's better to have a summarized plain text version for all our HTML entities since we also need it for project updates.

aminlatifi:
> @aminlatifi I think it's better to have a summarized plain text version for all our HTML entities since we also need it for project updates.

@MohammadPCh Do you have any spec in your mind for the summary? the length I mean

aminlatifi:
> @aminlatifi I think it's better to have a summarized plain text version for all our HTML entities since we also need it for project updates.

@MohammadPCh Do you have any spec in your mind for the summary? the length I mean

MohammadPCh:
@aminlatifi In my opinion, about 40~50 words or 300 chars is suitable.

aminlatifi:
`descriptionSummary` for projects and `contentSummary` for projectUpdates are ready, deployed to the staging env

aminlatifi:
@MohammadPCh I am concerned about displaying summary contents in the UI!
As an example summary
```
[...]

PR #867, opened 2023-02-15
Html field summary

aminlatifi:
Added descriptionSummary to Project entity
Added contentSummary to ProjectUpdate entity
Added getHtmlTextSummary method to summarize html using html-to-text Added listeners to Project and ProjectUpdate to fill summary fields on insert and update

Refs giveth/impact-graph#816


migration/1676472760533-FillProjectDescriptionSummaryProjectUpdateContentSummary.ts
```diff
@@ -0,0 +1,59 @@
...
+      );
+
+      skip += projects.length;
+      if (skip >= count) break;
+    }
+
+    skip = 0;
+    while (true) {
+      const [projectUpdates, count] =
+        await queryRunner.manager.findAndCount<ProjectUpdate>(ProjectUpdate, {
```
CarlosQ96, line 38: I think you should specify a ```take``` amount, this will fetch all the records in the db might be an expensive operation and block the migration.

CarlosQ96 (changes requested):
Looks good to me only one suggestion to use take and not fetch all from db.

migration/1676472760533-FillProjectDescriptionSummaryProjectUpdateContentSummary.ts
CarlosQ96: The docs don't specify a ```take``` by default in the resulting query so I assume it fetches all.

CarlosQ96:
After that can merge.

aminlatifi:
> After that can merge.

Thanks @CarlosQ96

mohammadranjbarz (approved):
Thanks @aminlatifi 
I didnt review all files, but the last feedback of @CarlosQ96  has been resolved

LGTM

aminlatifi:
Thanks @mohammadranjbarz. I merged it, after running the migrations all the projects and projectUpdates must have summary in their descriptionSummary and contentSummary respectively

PR #879, opened 2023-02-22
Project review status field

aminlatifi:
Added project ReviewStatus field in parallel to the listed field


CarlosQ96 (approved):
Thanks Amin, looks good to me the Parallel column so we keep it backward compatible for frontend!

I pushed a change to update listed column when reviewStatus changes in the UI

PR #883, opened 2023-02-23
Release 1.10.2

aminlatifi:
- Fixed project set review status migration issue


CarlosQ96 (approved):
Thanks! Pure SQL migration is better!

Issue #890, opened 2023-03-01
Add polygon support to monoswap


aminlatifi:
https://github.com/Giveth/impact-graph/issues/887

aminlatifi:
I assigned it to @mateodaza as he is the only one who has access to the monoswap repo. I think it would be better to migrate that repo to Giveth organization.

aminlatifi:
@mateodaza I forked the branch to https://github.com/Giveth/monoswap, and published to https://www.npmjs.com/package/@giveth/monoswap
Let's make any further modifications to that.
If you don't have access to publish a new version let me know

MoeNick:
@mateodaza I changed the assignee to @aminlatifi fyi, seems he forked it to Giveth Org so we may not have any blocker for now.

aminlatifi:
Monoswap is developed based on uniswap sdk v2, which doesn't support polygon network. Investigating migrating to uniswap SDK v3.

aminlatifi:
Published monoswap v1 which supports uniswap v3 protocol, consequently prices on the polygon network. Will test impact graph with this new version

aminlatifi:
https://github.com/Giveth/monoswap/pull/1

PR #898, opened 2023-03-08
monoswap upgrade + polygon price support

aminlatifi:
- Upgraded moswap
- Added polygon price support
- #890


src/services/donationService.test.ts
```diff
@@ -585,4 +587,42 @@ function fillOldStableCoinDonationsPriceTestCases() {
...
+      token,
+      ['USDT', 'MATIC'],
+      CHAIN_ID.POLYGON,
+      amount,
+    );
+
+    donation = (await findDonationById(donation.id))!;
+
+    expect(donation.valueUsd).to.gt(0);
+    assert.equal(
```
mohammadranjbarz, line 622: Do `MACIT` and `ETH` has same price?

mohammadranjbarz (approved):
Thanks @aminlatifi  I tested locally and it works fine, LGTM

PR #900, opened 2023-03-12
Feature total donations analytics

CarlosQ96:
New query for analytics dashboard.


CarlosQ96:
New analytics query for the dashboard

Issue #860, opened 2023-02-09
Add origin header to jrpc queries


aminlatifi:
https://github.com/Giveth/devops/issues/93

aminlatifi:
@MoeNick someone from devops team can test this feature.

MoeNick:
@mhmdksh can you QA this issue?

mhmdksh:
Hey @MoeNick yes sure.
I tested this before and it is acting as it should be.
We have metrics monitored on our Endpoints by header. Showing the queries origin. I think we can close this

<img width="1195" alt="image" src="https://user-images.githubusercontent.com/52987806/230899067-fe7e7ec6-8d1f-43c4-9df7-c83c2bfc1c04.png">

PR #906, opened 2023-03-14
Add Origin header to node rpc queries + remove web3js dependency

aminlatifi:
Removed web3 dependency
Added origin headers to node rpc queries
Added SERVICE_NAME key to the config

https://github.com/Giveth/impact-graph/issues/860


mohammadranjbarz (approved):
Thanks @aminlatifi  

LGTM

PR #925, opened 2023-03-27
User redis cluster

aminlatifi:
Removed redis docker from docker-compose-production.yml
Added Redis username field to the config


mohammadranjbarz (approved):
Thanks @aminlatifi LGTM

PR #936, opened 2023-03-30
Use zero address for celo native tokens

aminlatifi:


mohammadranjbarz (approved):
Thanks @aminlatifi 
LGTM

PR #941, opened 2023-04-04
Hotfix add some donations to db with migration

mohammadranjbarz:
Griff wanted to add some donations in DB , but don't pay them givbacks, Griff's message:

* This is 6 donations in xDAI and GIV:
https://gnosisscan.io/tx/0x7c56dce8d49799d11de5f2c8b9637564c9b12fe6638de9226ff6bd698beed8d9/
* This is 15 donations on mainnet: https://etherscan.io/tx/0x7d62181f89d5abde19d4159a59a5e2bfee17860207b0e5fb026e93d49cc98fdd
* The matic one has 6 donations: 
https://polygonscan.com/tx/0x7e34422db03a384def1d7eb69d95e063b84fb025a39d6cce42cefd54a903e35f 

@CarlosQ96  I think for this kind of support tasks we can write migration, in this way we would know what we have done so far and someone else can review our things, instead of adding them directly to DB

### DIFF
--- a/migration/1680507337701-add_some_donations_to_db.ts
+++ b/migration/1680507337701-add_some_donations_to_db.ts
@@ -323,7 +323,7 @@ export class addSomeDonationsToDb1680507337701 implements MigrationInterface {
     }
     await queryRunner.query(`
                 INSERT INTO public.user ( "walletAddress", role,"loginType", name) 
-                VALUES('${fromWalletAddress}', 'restricted','wallet', 'test');
+                VALUES('${fromWalletAddress}', 'restricted','wallet', 'A Gnosis safe wallet');
               `);
     const user = (
       await queryRunner.query(`SELECT * FROM public.user
--- a/migration/1680507337701-add_some_donations_to_db.ts
+++ b/migration/1680507337701-add_some_donations_to_db.ts
@@ -1,0 +1,364 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import config from '../src/config';
+import { Donation } from '../src/entities/donation';
+import { NETWORK_IDS } from '../src/provider';
+import moment from 'moment';
+
+// For seeing donations detail you can see this message ( if you have access to channel)
+// https://discord.com/channels/679428761438912522/928813033600475207/1089868809302724618
+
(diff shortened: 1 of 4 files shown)


migration/1680507337701-add_some_donations_to_db.ts
```diff
@@ -0,0 +1,364 @@
...
+    transactionNetworkId: NETWORK_IDS.POLYGON,
+    createdAt: new Date(1679915779000),
+  },
+];
+
+export class addSomeDonationsToDb1680507337701 implements MigrationInterface {
+  async up(queryRunner: QueryRunner): Promise<void> {
+    const environment = config.get('ENVIRONMENT') as string;
+
+    if (environment !== 'local') {
```
CarlosQ96: this wont let the donation be created in production.
Staging and production has Environment as "staging" and "production" respectively.

CarlosQ96 (changes requested):
Hey Mohammad Code looks good to me but this:
if (environment !== 'local') wont let the donations be created on prod or staging.
Only when env === 'local' it will let them.

Staging and prod have 'staging' and 'production' strings in the env.

CarlosQ96:
Also testss are failing for me locally too like in Amin's Branch I think we have to update env keys. Some key is failling.

migration/1680507337701-add_some_donations_to_db.ts
mohammadranjbarz: Thanks for mentioning, I fixed it

mohammadranjbarz:
@WhyldWanderer  @GriffGreen the donations are in our produciton now

https://giveth.io/project/graceaid-earthquake-relief
https://giveth.io/project/anka-relief
https://giveth.io/project/banklessdao-turkey-disaster-relief-fund

mohammadranjbarz:
@CarlosQ96  I merged it and after that I remembered something, when we add donations with migration we can updated total donations of donor and project in that migration file as well, I just mentioned it for future migrations

mohammadranjbarz:
https://github.com/Giveth/impact-graph/pull/1004 is more complete approach

Issue #937, opened 2023-04-03
Set size limit for string fields


mohammadranjbarz:
We dont have any validator for setting limit for string input size of project/projectUpdates and it makes problem, for instance  https://giveth.io/project/sustainable-food-security-and-ecosystem-restoration project had a 29MB description, so we couldn't fetch this project, related issue
https://github.com/Giveth/giveth-dapps-v2/issues/2441

aminlatifi:
The main problem is images are save in base64 format! we have projects and project updates with huge sizes. Some project descriptions have length around 3MB and the situation is worse for project updates, with lengths like ~= 11 MB, 7MB ....
I can create a PR to limit those sizes, but before that we must find a way to clean those contents

MoeNick:
@aminlatifi @mohammadranjbarz does it need QA ? is it merged ?

aminlatifi:
Unfortunately, I reverted all I did because it made issues on the backend. We must implement validators on resolvers

laurenluz:
Amin from dApp call:
need to put limitation and validation on APIs
issue when this happened: someone created a project w/ 38MB
We need to protect our APIs by saying limits

aminlatifi:
https://github.com/Giveth/impact-graph/pull/958

PR #942, opened 2023-04-04
Feat/limit text fields length

aminlatifi:
#937


aminlatifi:
I checked with production db to not set max limit lower than already filled values.

CarlosQ96 (commented):
THe code looks good to me ! I ran it locally to my db and data persist fine. 

Issue is 6 test fails locally and in the actions I think its related to env keys.

PR #945, opened 2023-04-05
Allow admins to change the admin of a project

CarlosQ96:
Allow admins to change the adminuserId.
I did it this way because if I use the default behaviour of the adminJs, it will load 7000+ record array (thiss breaks the server or putss it slow). Its better to just allow changing the string and we will change it by code the relation.

This is a highly requested feature and part of the every week activities.


mohammadranjbarz (approved):
Thanks @CarlosQ96  LGTM

PR #947, opened 2023-04-05
Hotfix: Improve chainvine validations to avoid sending not required data

CarlosQ96:
Feedback from Maryjaf


mohammadranjbarz (approved):
Nice catch @CarlosQ96 
LGTM

PR #954, opened 2023-04-10
Refactor admin bro

mohammadranjbarz:


mohammadranjbarz:
@CarlosQ96  The things that I did on this PR:
1. Merge `master` to to your branch
2. Refactor role permissions function to delete some duplicated codes
3. Deleted `false` permissions to make the functions shorter ( based on our codes undefined and false is equal for permissions, so we do not set them it makes codes simpler)
4. Split `adminBro.ts` to multiple files ( 1 file per tab)
5. Write a migration to create admins with different roles, so we can run the migration locally and easily test all roles

mohammadranjbarz:
@CarlosQ96  I merged this PR to `hotfix_admin_campaign_role` by mistake haahaa
Please just see it to stay in the loop, then we can merge your PR

PR #991, opened 2023-05-16
Implement new chainvine api and entities

CarlosQ96:
Related to: https://github.com/Giveth/giveth-dapps-v2/issues/2616

Second try as first release broke staging


CarlosQ96:
@mohammadranjbarz @aminlatifi Could you double check the migration please? see why it wont run on develop or staging? 

I had to revert staging it was down because of this Pr when I merged.

mohammadranjbarz:
> @mohammadranjbarz @aminlatifi Could you double check the migration please? see why it wont run on develop or staging?
> 
> I had to revert staging it was down because of this Pr when I merged.

@CarlosQ96 
I was getting error while running migrations locally, I asked chatGPT, it has this recommendation and it fixed it for me, I guess, it would resolve it on server as well, although  if we still getting error on staging server we can pass the error message to chatGPT for fixing that, Anyway I pushed the commit on this PR


## ChatGPT response
The error you're seeing is coming from this part of your migration:

```
{
  name: 'id',
  type: 'serial',
  isPrimary: true,
  isGenerated: true,
}
```
In the CREATE TABLE query, it tries to create a field 'id' NOT NULL which is incorrect SQL syntax. This is because the type 'serial' in Postgres is a not-null integer column and is auto-incremented.

The TypeORM definition for the 'serial' type should correctly infer that the column is not null and auto-incremented. However, for some reason it appears that TypeORM isn't translating this correctly into SQL.

To fix this, you can use 'int' as the type and separately specify the auto-increment property as generationStrategy: 'increment'. So your column definition would look like:

```
{
  name: 'id',
  type: 'int',
  isPrimary: true,
  isGenerated: true,
  generationStrategy: 'increment',
}
```
This should correctly translate to the desired SQL for a serial column. Please try this and see if it fixes the error.

mohammadranjbarz (approved):
@CarlosQ96  Thanks for this huge implementation, LGTM

Issue #1003, opened 2023-05-19
Adding some donations manually to DB


mohammadranjbarz:
@WhyldWanderer asked for adding 
https://etherscan.io/tx/0x639f0170a9d56a3ea007be67afac8d5afa945421c517dca4c6d54e7941a2e827 TXs as donation in DB

mohammadranjbarz:
I should do it like I did before in https://github.com/Giveth/impact-graph/pull/941

mohammadranjbarz:
@WhyldWanderer 
I added the donations to production DB, so you can see them in givback calculation response or in project's page like below screenshot


https://giveth.io/project/emprende-dao?tab=donations
https://giveth.io/project/Shenanigan-0?tab=donations
https://giveth.io/project/1hive-gardens-0?tab=donations
https://giveth.io/project/follow-the-black-hare?tab=donations
https://giveth.io/project/lunco-robotics-engineering?tab=donations
https://giveth.io/project/moloch-cloudship-0?tab=donations
https://giveth.io/project/africa-hyper-volunteers-centre?tab=donations
https://giveth.io/project/metagame-0?tab=donations
https://giveth.io/project/akiyadao?tab=donations
https://giveth.io/project/free-the-food-0?tab=donations
https://giveth.io/project/NFTP-Rainbow-Rolls-0?tab=donations
https://giveth.io/project/giveth-names?tab=donations
https://giveth.io/project/cambiatus?tab=donations
https://giveth.io/project/art-and-culture-center?tab=donations
https://giveth.io/project/empowerment-incubator?tab=donations
https://giveth.io/project/the-village-project?tab=donations

<img width="815" alt="Screen Shot 1402-03-08 at 12 05 00" src="https://github.com/Giveth/impact-graph/assets/9850545/dce0987c-89c7-49e8-ba1d-34cdfcc617d0">

WhyldWanderer:
Looks good Mohammad! I can see the donations reflected on all project pages and in the Swagger API when I run the eligible donations script for GIVbacks. Thanks so much!

PR #1004, opened 2023-05-21
F 1003 write migrations to add donations

mohammadranjbarz:
related to #1003


mohammadranjbarz:
Initing AppDataSource is needed that is missing in this PR see
https://github.com/Giveth/impact-graph/commit/da12b0695345756a767c6084760baa519ee36284

Issue #988, opened 2023-05-08
Backend for Estimate Matching


laurenluz:
Fork what Gitcoin did:

https://github.com/gitcoinco/web/tree/master/app/grants

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/30273b35-d5c4-4861-9f49-6d3663bcb711)

MoeNick:
This is the correct link:
https://github.com/gitcoinco/web/tree/master/app/grants

@mohammadranjbarz @aminlatifi do you need more information to implement this?

mohammadranjbarz:
This link return `404`
<img width="1825" alt="Screen Shot 1402-02-24 at 16 43 49" src="https://github.com/Giveth/giveth-dapps-v2/assets/9850545/75339a84-5a31-43e4-abb6-c3da37586554">

Anyway , @laurenluz  Do we have to use reverse engineering, don't we have any specific algorithm or formula for calculating an estimated match?
I think if someone provides that formula it's not only easier to implement but also make it easier for @maryjaf  to test it later

MoeNick:
I edited the link by adding "s" at the end of url

laurenluz:
double check that link that moenick sent @mohammadranjbarz

we shouldn't have to do much reverse engineering, we should be able just to fork the code that gitcon already created.

aminlatifi:
I could find this js implementation of QF
https://github.com/Anish-Agnihotri/quadratic-funding/blob/3e173091a5058faf6d44bebdb346262b62596dbb/pages/index.js#L84-L115

It explains the logic in js, in order to have this data per project in the backend we must provision some APIs and processes

aminlatifi:
> I could find this js implementation of QF https://github.com/Anish-Agnihotri/quadratic-funding/blob/3e173091a5058faf6d44bebdb346262b62596dbb/pages/index.js#L84-L115
> 
> It explains the logic in js, in order to have this data per project in the backend we must provision some APIs and processes
 
The link above is the source code of https://wtfisqf.com/?grant=&grant=&grant=&grant=&match=1000

The logic is in this part
<img width="810" alt="image" src="https://github.com/Giveth/impact-graph/assets/5684607/02c58ffc-ef1c-4447-b15c-d28c5607b8af">


In PR #1009 my plan was to create an API which response
1. Current active qf round `allocatedFound`
2. The sum of square root of donations to the project. In case of above code, is `sumAmount` before be squared (its value till line 100)
[...]

PR #1009, opened 2023-05-25
estimate matching

aminlatifi:
#988

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "test:campaignRepository": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/repositories/campaignRepository.test.ts",
     "test:socialProfileRepository": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/repositories/socialProfileRepository.test.ts",
     "test:projectVerificationRepository": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/repositories/projectVerificationRepository.test.ts",
+    "test:qfRoundRepository": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/repositories/qfRoundRepository.test.ts",
     "test:project": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/entities/project.test.ts",
     "test:notifyDonationsWithSegment": "NODE_ENV=test mocha ./test/pre-test-scripts.ts  ./src/services/cronJobs/notifyDonationsWithSegment.test.ts",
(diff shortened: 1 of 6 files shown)


CarlosQ96:
Will Merge it to upgrade the main branch of Mohammad this way we can unblock frontend in dev server
I think the calculationss are correct I added new queries and helpers. Will add tests in main branch. @mohammadranjbarz

CarlosQ96:
HOWEVER, following AMin's idea with the PR a query was made with the different parts of the formula. I think to calculate this frontend should do it, can be expensive for backend:
image

CarlosQ96:
Or we can make a method receiving multiples values and return them all at once.

aminlatifi:
@CarlosQ96 Thanks for finishing this PR, but there is a point. My idea was to use `getProjectDonationsSqrtRootSum` instead of `getQfRoundTotalProjectsDonationsSumExcludingProjectById` which you has created. 
First, that query is cheaper to run, and the same result can be achieved in the frontend. What do you think?
Besides, you have added to methods which would clutter the code.

CarlosQ96:
Yeah lets optimize now, I think the first step was to unblock cherik and ramin since we had them there waiting. I think it's the correct idea to do the calculations in the frontend. I thought that would be your idea. 

However!!!!

The reason I made ```getQfRoundTotalProjectsDonationsSumExcludingProjectById``` its because if they are going to simulate the added estimated matching based on 9 dai, 10 dai , 20 dai like in the following picture:
image


They have to recalculate the total, this total, can't include the current square-root of the project since it changed. They have to add the new squareroot to the sum and to the total. 
Then subtract NewEstimateMatching - CurrentEstimatedMatching = added estimated value per donation.

That's the way I saw to calculate the image above. Or is there an easier way?

CarlosQ96:
What do you think @aminlatifi ?

aminlatifi:
> e way I saw to calculate the image above. Or is there an easier wa

I think there is an easier way, but I didn't explain it here, sorry for that. 
The front-end can easily calculate the square root of the project and reduce it from the total value! Then to obtain the new value do the calculation again with computing the modified square root of the project donations and and apply the different to the total sum too. @CarlosQ96 Does it make sense?

```js
const projectSquareRootSquared = Math.Power(projectSquareRoot,2);
const totalExceptProject = TotalSum -projectSquareRootSquared ;
const newProjectSquareRoot = projectSquareRoot + Math.sqrt(newDonationValue);
const newProjectSquareRootSquared = Math.power(newProjectSquareRoot, 2)
const newTotal  = totalExceptProject + newProjectSquareRootSquared
....

```

CarlosQ96:
Yeah thanks, I overlooked that way. I agree.

Issue #980, opened 2023-05-10
Backend changes for support QF


mohammadranjbarz:
We should add new entity for qfRound

# Entity changes

## qfRound entity

| name          | type    | required | description                                                                                           |
|---------------|---------|----------|-------------------------------------------------------------------------------------------------------|
| name          | string  | yes      | can have name or something like round1, round2, ...                                                   |
| beginDate     | Date    | yes      |                                                                                                       |
| endDate       | Date    | yes      |                                                                                                       |
| active        | boolean | yes      |                                                                                                       |
| allocatedFund | integer | no       | With knowing how much allocated to this round we can estimate project's share for showing in frontend |
| createdAt     | Date    | yes      |                                                                                                       |

## project entity changes
We just need to add a boolean named `qfEligible` and if it's true, we will know this project is assigned to active qfRound and it's enough for us

## donation entity change
We should add
* `qfRound`: We fill this with active qfRound, if project that we donate on that is assigned to current active qfRound otherwise it will be empty
* `donorPassportScore`: We fill this with donor's passportScore if `qfRound` is filled and also we might update this field in the end of the qfRound

# Endpoints
* We need an endpoint to return projects related to current active qfRound
* Get donation details and estimated fund share for the project

# Others
* We need to have a function, that get projectId, then return the allocated qfRound fund to this project based on donations and donorScores to this project
* will need to have a function that considers the amount donated to a particular project & the number of donors... and compares it to the amounts donated & number of donors on every other project... and then returns some values for "estimated matching". These numbers should update every x period of time... so that they stay up-to-date as more people donate.
* We need a function that get a qfRound number as input, and updates all donorScores of the donations that are related to this qfRound
* We need to add a button in admin panel on project page to add/remove that project to current active qfRound
* We should can edit qfRound in admin panel

MoeNick:
I have these in my mind that may have some changes on your ideas:

- List of donations by QF and the impact they made. 
- Authentication of createDonation validation by Scorer (Gitcoin0

our regular donations is sending some token to another address and their validation is easy, let's review the QF donations. where they go? don't they go to a smart contact if so, the verification may change.

aminlatifi:
Thanks @mohammadranjbarz for preparing this,

In my view, we must reconsider project<>qfRound relation. As I know, a project will be participated in qfRound or not. So it can be only a true/false flag instead of a many-to-many relation with qfRound entity.
[...]

PR #1014, opened 2023-06-06
Move estimated match calculations to materialized views

mohammadranjbarz:
related to  #980

@CarlosQ96  implementing below comment
https://github.com/Giveth/impact-graph/issues/988#issuecomment-1574827339

@aminlatifi  you can take a look if you want to


migration/1685958638251-ProjectEstimatedMatchingView.ts
```diff
@@ -0,0 +1,51 @@
...
+          "donation"."projectId",
+          "donation"."qfRoundId",
+          SUM("donation"."valueUsd") as "valueUsd"
+        FROM
+          "donation"
+          INNER JOIN "user" ON "user"."id" = "donation"."userId"
+          INNER JOIN "qf_round" ON "qf_round"."id" = "donation"."qfRoundId"
+        WHERE
+          "donation"."status" = 'verified'
+          AND "user"."passportScore" >= "qf_round"."minimumPassportScore"
```
aminlatifi: It's decided to skip users' passport score in calculating the estimate matching, so its condition must be removed.
Also, we must fill the view for donations of an active qfRound, not all of them.

aminlatifi (changes requested):

mohammadranjbarz:
Another feedback from @aminlatifi 

> Did you keep getQfRoundTotalProjectsDonationsSum method? Non of our views has that information
You must create a materialized view, refreshed on qf round information change or new donation. Then change the function implementation to read from that

mohammadranjbarz:
@laurenluz  @MoeNick 

I and @aminlatifi  discussed about something , but couldn't conclude so I ask it here to decide about it.

before jumping to topic, I should explain a little about materialized view, sometimes we need complex queries that put too much loads on database, for executing those queries each time, we create some materialized view that store result of those queries somewhere something like table so we can easily feth data from that view instead of executing heavy queries, but we have to refresh those views manually, for instance in give power views, when someone boost s project we refresh related view to make sure we have updated data there

Now in estimated matching materialized view we have two ways:
1. in query of materialized view we can consider donor's passport score, so don't count donations that donors doesn't have minimum score, but we have to sometimes refresh materialized view because some user's scores have changed( or we can do not refresh , because after every donation we refresh the views and it can do that thing, or even we can just refresh it once in a day to user new updated passport scores)

2. Not consider donor's passport score in estimated matching, so the estimated amount might have lots of differents because in this way when estimating we consider all donations and donors are eligible, imaging some non-eligible donor donate about 1000$ to a project, or in a project 20 non eligible in passport donate to a project, those project will have lots of estimated amount but when we want to distribute fund they will receive much less and it can be misleading

@aminlatifi  correct me if I'm wrong or I missed something

@laurenluz  @MoeNick  the implementation for both way it's almost similar and we don't need much time for implementing, so based on this info please tell me what should I do

aminlatifi:
Thanks @mohammadranjbarz for explaining the issue.
Mohammad is correct, we can do it in both ways, what's important is our refreshing policy. Logically, we must refresh the views whenever its source data is updated or on a regular basis. There is a trade-off basically between preciseness and performance. 
I am thinking about upgrading our db to a cluster with a main node for write and some read replicas to solve this performance issue in long term, not started that yet.

aminlatifi:
I talked with @mhmdksh, we found some [doc](https://docs.digitalocean.com/products/databases/postgresql/how-to/add-read-only-nodes/) that shows having db read replicas would be possible in digitalocan, our provider.

mhmdksh:
@aminlatifi @mohammadranjbarz I have created a read-replica for the DB.

Please DM me for connection credentials when you need them

laurenluz:
Hey @mohammadranjbarz @aminlatifi 

Tbh, I think the best option is to just ignore user's passport score in the calculation of estimated matching. The purpose of the estimated matching amount is to encourage users to donate... it's to give them the felt experience of "wow, my donation is make 2x / 5x/ 10x the impact!"

In the early gitcoin rounds, the estimated matching was very inaccurate... but it didn't matter in the end.

We want to allow users to verify their gitcoin passport and improve their score at any point before the end of the round... and don't want to scare them away from donating/create more friction for donating by saying "you will have an estimated matching of 0".

does that make sense? wdyt?

I think it also sounds like that will make it easier on the backend since we won't have to refresh so much

mohammadranjbarz:
> Hey @mohammadranjbarz @aminlatifi
> 
> Tbh, I think the best option is to just ignore user's passport score in the calculation of estimated matching. The purpose of the estimated matching amount is to encourage users to donate... it's to give them the felt experience of "wow, my donation is make 2x / 5x/ 10x the impact!"
> 
> In the early gitcoin rounds, the estimated matching was very inaccurate... but it didn't matter in the end.
> 
> We want to allow users to verify their gitcoin passport and improve their score at any point before the end of the round... and don't want to scare them away from donating/create more friction for donating by saying "you will have an estimated matching of 0".
> 
> does that make sense? wdyt?
> 
> I think it also sounds like that will make it easier on the backend since we won't have to refresh so much

Yeah sure, @aminlatifi  was telling that as well, so I will remove `passportScore` condition for estimated matching

mohammadranjbarz:
@CarlosQ96  I moved all queries to materialized views and also deleted the functions/endpoints that @aminlatifi  said we don't need  them 

now frontend guys can get estimated matching data like this in graphql query

```
        estimatedMatching {
           projectDonationsSqrtRootSum
           allProjectsSum
           matchingPool
         }
```

And the can calculate estimate matching amount with this function:
```
export const calculateEstimatedMatchingWithParams = async (params: {
  matchingPool: number;
  projectDonationsSqrtRootSum: number;
  allProjectsSum: number;
}): Promise<number> => {
  const { projectDonationsSqrtRootSum, allProjectsSum, matchingPool } = params;

  return (
    ((projectDonationsSqrtRootSum * projectDonationsSqrtRootSum) /
      allProjectsSum) *
    matchingPool
  );
};
```

CarlosQ96 (approved):
Calculations look good to me. Ill merge to continue the single project page calculations.

Issue #1018, opened 2023-06-11
Write cronjob or something to periodically check qfRound.endDate and make them inactive if time passed


mohammadranjbarz:

mohammadranjbarz:
@maryjaf 
Now you can test it, if it passes more than 10 minutes from and active qfRound out job will deactivate that qfRound

PR #1024, opened 2023-06-13
Update active status of qfRounds if endTime has passed

mohammadranjbarz:
related to  #1018


CarlosQ96 (approved):
Thanks Mohammad PR looks good to me. So we refresh views when we detect expired rounds to leave the Actual Matching

PR #1035, opened 2023-06-20
Prevent project address from throwing errors silently

CarlosQ96:
Critical bug found by Ashley when debugging two projects that somehow have no addresses. 

We where using the insert method that threw no errors. Modified this to throw errors, halting execution and cleaning up.

### DIFF
--- a/src/repositories/projectAddressRepository.ts
+++ b/src/repositories/projectAddressRepository.ts
@@ -2,6 +2,7 @@ import { ProjectAddress } from '../entities/projectAddress';
 import { Project } from '../entities/project';
 import { User } from '../entities/user';
 import { BaseEntity } from 'typeorm';
+import { logger } from '../utils/logger';
 
 export const getPurpleListAddresses = async (): Promise<
   { projectAddress: string }[]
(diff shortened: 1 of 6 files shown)


CarlosQ96:
@mohammadranjbarz do you think this is a viable solution? Halting the execution and throwing the error to the user to inform them in the modal. Let me know what you think

CarlosQ96:
Following the project closely everything was created EXCEPT the addresses, even the projectUpdate (that occurs after project.save() and bulk insert of addresses is saved). 
Confirmed the previous insert does not throw error. So the users is assuming everything went well.

mohammadranjbarz:
@CarlosQ96  I think it might cause some problems, for instance we see project has been created but we couldn't add addresses, I think best solution is to see why it failed ( I mean find possible errors for bulk insert) then write a validator function for those addresses, so when creating/updating projects before doing anything we first validate the addresses in this case we make sure everything goes successfully or we do not insert any thing
WDYT?

CarlosQ96:
actually lets just leave the logs to detect the error. and make a branch of production as I think we should leave this live.  ill close this pr since its based of staging

CarlosQ96:
Recreating pr

Issue #1073, opened 2023-07-24
Editting project from project view page in admin panel causes removing recipient addresses!!!!


mohammadranjbarz:

mohammadranjbarz:
@maryjaf It's deployed on Staging, if you tested and it was ok, tell me to merge the hotfix PR on production
https://github.com/Giveth/impact-graph/pull/1075

maryjaf:
Thanks @mohammadranjbarz,  I'v checked on stg and this problem has been resolved

mohammadranjbarz:
@maryjaf  Now It's on the production can you test it?

@laurenluz  @WhyldWanderer 
I also ran below query to see what projects don't have  addresses to see maybe their projects got affected by this bug

```
SELECT p.slug, p.title, p.id
FROM project p
LEFT JOIN project_address pa ON p.id = pa."projectId"
WHERE pa."projectId" IS NULL;
```

So these are the result of the query

|slug                                            |title                                             |id  |
|------------------------------------------------|--------------------------------------------------|----|
|All-for-Climate-DAO-0                           |All for Climate DAO                               |1241|
|Music-Hive-Community-0                          |Music Hive Community                              |1417|
|internal-testing                                |Internal Testing                                  |1648|
|testfilter                                      |testfilter                                        |2132|
|testemail                                       |testemail                                         |2222|
|maryam                                          |maryam                                            |2238|
|testfinal001                                    |testfinal001                                      |2289|
|pay-the-pizza-forward-at-ethdenver              |Pay the Pizza Forward at ETHDenver!               |2345|
|the-river-giver-for-creators                    |The River Giver for Creators                      |2361|
|testy                                           |testy                                             |2366|
|rrr                                             |RRR                                               |2370|
|her-dao-latam                                   |HER DAO LATAM                                     |2508|
|needs-map                                       |NeedsMap                                          |2518|
|needs-map-0                                     |Needs Map                                         |2519|
|cvf                                             |CVF                                               |2526|
|community-volunteers-foundation                 |Community Volunteers Foundation                   |2527|
|support-to-life                                 |Support To Life                                   |2539|
|afad-disaster-and-emergency-management-authority|AFAD - Disaster and Emergency Management Authority|2543|
|tider-basic-needs-association                   |TIDER - Basic Needs Association                   |2544|
|turkish-red-crescent                            |Turkish Red Crescent                              |2546|
|akut-search-and-rescue                          |AKUT Search and Rescue                            |2548|
[...]

PR #1074, opened 2023-07-24
Hotfix 1073 fix removing recipient addresses while editting in admin panel

mohammadranjbarz:
related to #1073


CarlosQ96 (approved):
LGTM!

Issue #1002, opened 2023-05-19
Data Collection for Sybil Analysis


laurenluz:
After each round of QF, we will need to collect the donation data, make a table/spreadsheet/dashboard for it, and then analyze it to remove possible sybil attacks.

I hope @mohammadranjbarz can help me to set everything up in the backend for QF so that this data is easy to grab & organize.

This, for example, is the data set that QF round organizers get from gitcoin in the beta round:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/c26c5bec-fd4f-4e70-a5c4-8076053b40aa)

Info we need:
- donor address
- "to" project id
- amount sent
- timestamp
- which chain
- passport score
- passport score threshold 

We might need more information for this, but I'll leave a comment in this issue once I figure it out. Also open to suggestions on things I missed. cc @moenick @griffgreen

@mohammadranjbarz - can you keep this in mind and keep an eye on this issue?

FYI @aminlatifi @MohammadPCh

MoeNick:
Do you prefer to have this feature in AdminJs? there's a csv export feature there. 
OR
we may add an endpoint like GIVbacks and set kinda OPS for each round. 
@laurenluz

laurenluz:
ahh man @MoeNick I don't understand the options you're proposing really. maybe you can explain it to me?

From my last meeting with Trustalabs "with whom, we are thinking about doing the sybil analysis work", they said... all they will need is a simple list of address that donated during the QF round.

I still need to talk with them more to get clarification and finalize that that is true.

MoeNick:
yeah if we know the manual operation process you and they may want to do, we may get more clear, I'd better not propose things actually. If you only need the data, it's fine. we can have it whenever you want, but if you need a feature, we have to make it clear. 

So this issue tmk is just a query.

laurenluz:
Hey @MoeNick @mohammadranjbarz @CarlosQ96 

yesterday I spoke w/ the crew at Trustalabs, and this is the information they will need from us to do the analysis... For all donations in our QF round to QF-eligible projects, we'll need:
- Donor Wallet address
- USD value of donation
- project they donated to - (I think link + wallet address + chain)
- Transaction hash

Will we need to set up anything special to get this info? or can we just get it from querying our database?

CarlosQ96:
[...]

PR #1076, opened 2023-07-26
Remake exporting to csv filters fetch

CarlosQ96:
Related to: https://github.com/Giveth/impact-graph/issues/1002
But also a bug fix, after updating adminjs, all routes changed and express middlewares were no longer accesible.
Removed Redis and instead used the rawHeaders attributes directly to fetch filter params, could not find another way with adminbro objects.


CarlosQ96:
Export example with fake data, including the userscore and qfroundId, fromwalletaddress,projectId, toWalletaddress, network id and transaction hash are all included like before.
image

mohammadranjbarz (approved):
Thanks @CarlosQ96 
LGTM

Issue #1067, opened 2023-07-17
Improve performance of query - getPowerBoostingSnapshotWithoutBalance


aminlatifi:
This query takes more than one minutes to run.
```sql
select "userId", "powerSnapshotId", "blockNumber","walletAddress"
        from (select DISTINCT "powerSnapshotId", "userId" from power_boosting_snapshot) as boosting
        inner join public."user" as "user" on  "userId"= "user".id
        inner join power_snapshot as "snapshot" on boosting."powerSnapshotId" = snapshot.id
        where snapshot."blockNumber" is not NULL
        and not exists (
          select
          from power_balance_snapshot as "balance"
          where balance."powerSnapshotId" = boosting."powerSnapshotId" and
          balance."userId" = boosting."userId"
        )
        order by "blockNumber" ASC
        LIMIT $1
        OFFSET $2
```

divine-comedian:
Need to change how we have designed back-end, putting a heavy load on the back-end, we probably cannot optimize this since it is a complex query. 

Will need a few weeks to work on technical debts like this

maryjaf:
How can I test this issue?
should we check all scenarios related to Givpower and ranking calculation  to sure every thing works as before on STG or sth more should be tested?
@aminlatifi

aminlatifi:
> How can I test this issue? should we check all scenarios related to Givpower and ranking calculation to sure every thing works as before on STG or sth more should be tested? @aminlatifi

@maryjaf Yes, GIVpower sorting (not instant boosting) must be tested. We must be sure the balance snapshotting and calculating the balance based on average of periodic snapshots is working as before

maryjaf:
On stg:
Start Date :Sun Aug 6 2023 8:00 PM
End Date: Sun Aug 20 2023 8:00 PM
Can you decrease the round time on stg ?

maryjaf:
Some times by modifying the percentage of Givpower, my Givpower is changed to 0 and after that by refreshing has been resolved

<img width="1397" alt="image" src="https://github.com/Giveth/impact-graph/assets/111529185/fbef99b4-cdcc-4de0-9c1a-d7823873b134">

<img width="1125" alt="image" src="https://github.com/Giveth/impact-graph/assets/111529185/70d07bd2-2129-46f5-a6ad-ca8b5fe52df0">

@aminlatifi

aminlatifi:
[...]

PR #1079, opened 2023-07-31
Changed givpower boosting related queries to improve performance

aminlatifi:
Changed power_balance_snapshot table balance column to accept null values Change take TAKE_POWER_BOOSTING_SNAPSHOT procedure to make instances of power_balance_snapshot for users has power_boosting_snapshot for each snapshot Updated tests to match these changes
Ref #1067


aminlatifi:
@CarlosQ96 @MohammadPCh We must monitor database load before and after this merge

aminlatifi:
I noticed some of gnosis tests pass by setting xdai rpc to https://xdai-rpc.gateway.pokt.network/

CarlosQ96:
@aminlatifi once I change the XDAI node url like you said the error is fixed.

If you look at the errors its a UNIQUENESS ERROR. Which doesn't make sense. But I encountered this before.

 I found the root cause, I think it's a bug in our test library.
When any http/blockchain/rpc call fails (in this case the xdai url), the tests retry themselves multiple times.
Since we don't clean the data on each test start it hit the uniqueness error.

We should update the test environment to this new url and rerun tests.

It's not the PRs issue but some network, xdai rpc url issue.

CarlosQ96:
when changing Xdai url most tests worked but seems 2 are failing. Couldn't find the cause, if its still the xdai url or something else.

aminlatifi:
@CarlosQ96 Let's raise the gnosis provider with the Devops team.
Did you review the PR? can we proceed? do you think we must wait for all tests be passed?

CarlosQ96:
@aminlatifi I think all test need to pass or else it wont deploy. But I think the code is fine. 

Its a url/rpc issue not code.

CarlosQ96 (approved):
LGTM but we have to fix the URLS/RPC for tests to pass or wont deploy.

CarlosQ96:
Locally all tests pass for me, its an rpc provider issue.

CarlosQ96:
@geleeroyale @mhmdksh We are having some issue with the gnosis provider as it's not being reliable and failing when requesting data, throwing errors so our tests are failing.
We might need to find a new url or test won't pass.

geleeroyale:
Hey Carlos! Which provider do we have set? If it is a gateway RPC I recommend to switch back to our quicknode instance

PR #1088, opened 2023-08-06
Optimize latest project update query

CarlosQ96:
Related to:https: //github.com/Giveth/impact-graph/issues/1086

### DIFF
--- a/src/resolvers/projectResolver.ts
+++ b/src/resolvers/projectResolver.ts
@@ -1798,10 +1798,10 @@ export class ProjectResolver {
     @Ctx() { req: { user } }: ApolloContext,
   ): Promise<ProjectUpdatesResponse> {
     const latestProjectUpdates = await ProjectUpdate.query(`
-      SELECT pu.id
+      SELECT pu.id, pu."projectId"
       FROM public.project_update AS pu
       WHERE pu."isMain" = false
-      GROUP BY pu.id
+      GROUP BY pu."projectId", pu.id
       ORDER BY MAX(pu."createdAt") DESC
       LIMIT ${take}
       OFFSET ${skip};
@@ -1819,7 +1819,7 @@ export class ProjectResolver {
       .where('projectUpdate.id IN (:...ids)', {
         ids: latestProjectUpdates.map(p => p.id),
       })
-      .orderBy('projectUpdate.id', 'DESC');
+      .orderBy('projectUpdate.createdAt', 'DESC');
 
     if (user && user?.userId)
       query = ProjectResolver.addReactionToProjectsUpdateQuery(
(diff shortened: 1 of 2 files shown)


src/resolvers/projectResolver.ts
```diff
@@ -1799,14 +1799,14 @@ export class ProjectResolver {
...
-        LIMIT 1
-      )
+      FROM public.project_update AS pu
+      JOIN (
+          SELECT "projectId", MAX("createdAt") AS max_createdAt
+          FROM public.project_update
+          WHERE "isMain" = false
+          GROUP BY "projectId"
+      ) AS puu_max
+      ON pu."projectId" = puu_max."projectId" AND pu."createdAt" = puu_max.max_createdAt
```
aminlatifi: As I see the Join is not necessary! No further information is added outside and both tables are same `project_update`.
@CarlosQ96 Do we have tests for this? can you add if there isn't and then test without join statement?
Something like this
```sql
 SELECT "projectId", MAX("createdAt") AS max_createdAt
          FROM public.project_update
          WHERE "isMain" = false
          GROUP BY "projectId"
```

aminlatifi (changes requested):

CarlosQ96:
Yes Amin there are tests for this. However, I think one can be improved a lot so ill finish the test and optimize more the query.

MoeNick:
Hey @CarlosQ96 
Amin already requested some change to this. 
Did  you review it?

CarlosQ96:
Further optimization from what I did in this PR, breaks a bunch of tests. I am somewhat stuck on that fixing tests.

src/resolvers/projectResolver.ts
CarlosQ96: @aminlatifi I implemented your suggestion, modified it a little and it works fine.

I also tested my join version (previous) and it worked fine.
The tests are in the projectResolver.test.ts (those worked with the original implementation too)

Can review again.

CarlosQ96:
@MoeNick @aminlatifi can review again

src/resolvers/projectResolver.ts
```diff
@@ -1799,17 +1799,12 @@ export class ProjectResolver {
...
-        SELECT puu.id
-        FROM project_update as puu
-        WHERE puu."isMain" = false AND pu."projectId" = puu."projectId"
-        ORDER BY puu."createdAt" DESC
-        LIMIT 1
-      )
-      ORDER BY pu."createdAt" DESC
+      FROM public.project_update AS pu
+      WHERE pu."isMain" = false
+      GROUP BY pu.id, pu."projectId"
```
aminlatifi: Are you sure it must be grouped by both `id` and `projectId`?! maybe just `projectId`

aminlatifi (changes requested):

src/resolvers/projectResolver.ts
CarlosQ96: You are right Amin I removed the double grouping. I only need the projectUpdate.projectId

aminlatifi (approved):
LGTM

PR #1101, opened 2023-08-22
Fix isWalletAddressInPurpleList webservice to just return addresses of verified active projects, and not case-sensitive

mohammadranjbarz:
related to https://github.com/Giveth/giveth-dapps-v2/issues/2861#issuecomment-1682321196


CarlosQ96 (approved):
LGTM!

Issue #1103, opened 2023-08-22
Cache qf related queries result


aminlatifi:
getQfRoundTotalProjectsDonationsSum and probably getProjectDonationsSqrtRootSum query result can be cached, later shorter. I noticed they are not cached and are executed on each API call.

maryjaf:
Thanks @mohammadranjbarz I checked and there hasn't been any problem related to matching fund, estimated matching and ...

PR #1122, opened 2023-09-06
Add cache to qfRound estimated matching queries

mohammadranjbarz:
related to #1103

@CarlosQ96  We needed to to change our query from raw SQL query to typeorm one to can use cache feature


CarlosQ96 (approved):
Thanks Mohammed LGTM

Just removed unused code, can merge

mohammadranjbarz:
Thanks @CarlosQ96 
I need to explain this to @maryjaf  and she should create some qfRounds before merging this, so after we prepared for testing that, I'll merge it

PR #1124, opened 2023-09-10
Feature: Add eligible donations to qfround entity

CarlosQ96:
Related to the meeting we had. but I cant find the issue. Or if @laurenluz created it.

Can add array of networks to qfround and in the donation creation decides if marking or not

### DIFF
--- a/migration/1694295208252-AddEligibleNetworksToQfRoundEntity.ts
+++ b/migration/1694295208252-AddEligibleNetworksToQfRoundEntity.ts
@@ -1,0 +1,16 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class AddEligibleNetworksToQfRoundEntity1694295208252
+  implements MigrationInterface
+{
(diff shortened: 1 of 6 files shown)


mohammadranjbarz (approved):
Thanks @CarlosQ96  it's a great simple solution, just wondering if default value for eligileDonations should be empty array or an array of all supported networks, I'm worried about previous rounds

CarlosQ96:
I thought about the empty state to reduce the workload for admins, and also instead of deleting each, just add those they want.

mohammadranjbarz:
> I thought about the empty state to reduce the workload for admins, and also instead of deleting each, just add those they want.

You are right, so if you just modify the migration file to fill previous qfRounds eligibleNetworks to make sure we would not have any problem on them would be great

CarlosQ96:
Ok ill add the migration tomorrow morning.

CarlosQ96:
@mohammadranjbarz ready for final review and merging.

Also for frontend they should check the attribute
```eligibleNetworks``` in the graphql results in any query they fetch this info.

Issue #1051, opened 2023-07-09
Return list of campaigns of a project in projectBySlug webservice


mohammadranjbarz:
https://github.com/Giveth/giveth-dapps-v2/issues/2544

MoeNick:
If you'r ok I can plan for this one this week.

mohammadranjbarz:
> If you'r ok I can plan for this one this week.

Yeah it's ok, we can do it

MoeNick:
@CarlosQ96 we planned this one from last week, is there any progress?

mohammadranjbarz:
@alireza-sharifpour  @MohammadPCh  @MoeNick 

Now we return list of active campaigns of a project in `projectBySlug` web service.

sample:

‍‍
‍‍‍```

export const fetchProjectsBySlugQuery = `
  query (
    $slug: String!
  ) {
    projectBySlug(
      slug: $slug
    ) {
      id
      title
      balance
      image
      slug
      creationDate
      updatedAt
      admin
      description
      walletAddress
      impactLocation
      qualityScore
      verified
      traceCampaignId
      listed
      reviewStatus
      givingBlocksId
      campaigns {
        id
        title
        description
        type
        photo
        video
        videoPreview
        slug
        isActive
        order
        landingLink
        filterFields
        sortingField
        createdAt
        updatedAt
      
      }
      ......
    }
  }
`;

```

maryjaf:
Are frontend changes required to see the campaign name as a tag in project page? @alireza-sharifpour

alireza-sharifpour:
> Are frontend changes required to see the campaign name as a tag in project page? @alireza-sharifpour

Yes we need to update the frontend. @maryjaf

maryjaf:
> > Are frontend changes required to see the campaign name as a tag in project page? @alireza-sharifpour
> 
> Yes we need to update the frontend. @maryjaf

Plz let me know when it's ready to test

MoeNick:
We have another issue for front end this one was only for backed. can you test by API @maryjaf ?

maryjaf:
Sample response:

{
    "data": {
        "projectBySlug": {
            "id": "223639",
            "campaigns": [
                **{
                    "id": "13",
                    "title": "testi"
                }**
            ],
            "slug": "testpfp010",
            "givingBlocksId": null,
            "changeId": null,
            "slugHistory": [],
            "updatedAt": "2023-07-17T06:21:33.227Z",
            "listed": true,
[...]

PR #1136, opened 2023-09-27
Add Other types of campaigns to projectBySlug webservice

mohammadranjbarz:
https://github.com/Giveth/impact-graph/issues/1051#issuecomment-1733713799


src/services/campaignService.ts
```diff
@@ -36,9 +35,55 @@ export const fillCampaignProjects = async (params: {
...
   }
+  projectCampaignCache = {};
+  const activeCampaigns = await findAllActiveCampaigns();
+  for (const campaign of activeCampaigns) {
+    const projectsQueryParams = createFetchCampaignProjectsQuery(campaign);
+    if (!projectsQueryParams) {
+      break;
+    }
+    const projectsQuery = filterProjectsQuery(projectsQueryParams);
+    const projects = await projectsQuery.getMany();
```
CarlosQ96, line 63: I think this is a dangerous query, we could have multiple active campaigns and a campaign might contain many many projects (specially filter type campaigns). This will load a lot of information PER REQUEST in memory, we had issues before where loading too much can block the request.

And its also inside a double loop. 

I suggest we reduce the complexity by making a similar query to this one but passing the ID of the project we need specifically (the minimal project). 

Since projectBySlug query returns only 1 item. I don't think its necessary to load everything into memory every time. Wdty?

src/resolvers/projectResolver.test.ts
```diff
@@ -5505,6 +5510,133 @@ function projectBySlugTestCases() {
...
+      type: CampaignType.SortField,
+      sortingField: CampaignSortingField.Newest,
+      slug: generateRandomString(),
+      title: 'title1',
+      description: 'description1',
+      photo: 'https://google.com',
+      order: 1,
+    }).save();
+    const result = await axios.post(graphqlUrl, {
+      query: fetchProjectsBySlugQuery,
```
CarlosQ96: can you rename the fetchProjectsbySlugQuery as fetchProjectBySlugQuery without the s. 

That was my bad but its a bit confusing, I thought it returned multiple projects but its just 1.

src/services/campaignService.ts
CarlosQ96: Even with the cache, every 10 minutes it could be a big memory spike.

CarlosQ96 (changes requested):
Mohammad ```getAllProjectsRelatedToActiveCampaigns``` I think we could improve the inner query and loops a bit. I think we could potentially load too much in memory. Wdyt?

Aside from that other tests/code looks good.

src/services/campaignService.ts
mohammadranjbarz: I dont know how we can solve it with a simple query, if there is something in your mind definitely it would be better, because for campaigns that uses filter or sort, the included projects may change during the time so if want to know a project belongs to which campaign we should check all campaign's projects and it would decrease performance especially **projectBySlug** is one of webservices that frontend call it almost many times.

And campaigns that uses filter and sort just have 10 projects so the cached object would not be a large object so I thought maybe it's better to store it in memory instead of redis ( because it's already complicated this part, I though it  may make it more complicated

src/resolvers/projectResolver.test.ts
mohammadranjbarz: haahaa you are right, done.

CarlosQ96 (approved):
I think we can move forward so QA can keep testing. Tests show it all looks good!!
Thanks!

My only suggestion is if there is any weird results or issues we use the methods inside redis.ts file, and cache using redis -> Which is less prone to data corruption/errors. We can do it in another Pr if its required.

WDYT? @mohammadranjbarz 
Can merge meanwhile.

mohammadranjbarz:
> I think we can move forward so QA can keep testing. Tests show it all looks good!! Thanks!
> 
> My only suggestion is if there is any weird results or issues we use the methods inside redis.ts file, and cache using redis -> Which is less prone to data corruption/errors. We can do it in another Pr if its required.
> 
> WDYT? @mohammadranjbarz Can merge meanwhile.

Thank @CarlosQ96 , you did a really good job on reviewing this PR, I appreciate it

PR #1137, opened 2023-09-29
Add qfFilter with estimatedMatchingView

CarlosQ96:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/2942


CarlosQ96:
@mohammadranjbarz ready for review

Its a filtered sort, per talking with @laurenluz

mohammadranjbarz (approved):
Thanks @CarlosQ96 LGTM
but I dont know with changing the **ProjectEstimatedMAtchingView.ts** file don't you need to write a migration to change the view in DB?
if we need it please add migration otherwise feel free to merge it

CarlosQ96:
@mohammadranjbarz they were mostly annotations that were missing, it should be safe to merge as we don't synchronize the entities.
Thanks!

PR #1190, opened 2023-12-05
Hotfix add matching funds as donations

mohammadranjbarz:
Related to #1186

### DIFF
--- a/migration/1701689359018-add_fields_to_qf_round_history.ts
+++ b/migration/1701689359018-add_fields_to_qf_round_history.ts
@@ -1,0 +1,69 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class addFieldsToQfRoundHistory1701689359018
+  implements MigrationInterface
+{
+  async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            ALTER TABLE "qf_round_history"
+            ADD COLUMN "matchingFundAmount" real ,
+            ADD COLUMN "matchingFundPriceUsd" real ,
+            ADD COLUMN "matchingFundCurrency" text ;
+        `);
+
+    await queryRunner.query(`
+            ALTER TABLE "donation"
+            ADD COLUMN "distributedFundQfRoundId" integer;
+
+            -- If you have a foreign key constraint to enforce the relationship
+            ALTER TABLE "donation"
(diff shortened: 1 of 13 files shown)


mohammadranjbarz:
I just added a button in admin panel to sync donations with distributed matching funds, but we should add some cronjobs for that to call the function daily, or something like that

**Relate Donations with Distributed Funds** button
Screen Shot 1402-09-14 at 16 02 46

migration/1701689359018-add_fields_to_qf_round_history.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // Alpha round in production
+    await queryRunner.query(`
+        UPDATE qf_round_history 
+        SET 
+            "matchingFundAmount" = "matchingFund", 
+            "matchingFundPriceUsd" = 1, 
+            "matchingFundCurrency" = 'WXDAI'
+        WHERE 
+            id = 2 AND "matchingFund" IS NOT NULL;
```
jainkrati, line 35: why are we hard coding the id to 2?

src/services/donationService.ts
```diff
@@ -436,3 +444,91 @@ export const sendSegmentEventForDonation = async (params: {
...
+    );
+    return;
+  }
+  logger.debug(
+    `insertDonationsFromQfRoundHistory Filling ${qfRoundHistories.length} qfRoundHistory info ...`,
+  );
+
+  const matchingFundFromAddress =
+    (process.env.MATCHING_FUND_DONATIONS_FROM_ADDRESS as string) ||
+    '0x6e8873085530406995170Da467010565968C7C62'; // Address behind donation.eth ENS address;
```
jainkrati: lets take this as a constant value

jainkrati (changes requested):
Left a comment. Once resolved we are good to merge

migration/1701689359018-add_fields_to_qf_round_history.ts
mohammadranjbarz: We have some qf rounds already in our DB and they have qfRoundHistory items in DB, they dont have new fields, so I had to put their ids hard coded here

mohammadranjbarz: Screen Shot 1402-09-14 at 17 10 15

@jainkrati

jainkrati (approved):

PR #1195, opened 2023-12-12
add validations for multisig donations

CarlosQ96:


mateodaza (approved):
Thanks, looking great

Issue #1187, opened 2023-12-04
Add solana address to projects


aminlatifi:
We need backend support for the project's Solana address:
- Support Solana address as a project address in database schemas. In the current state, projects can have different ethereum addresses per chain, but they don't support Solana.
- Add/Edit API for adding/editing Solana address of a project

MoeNick:
can we have it more generic, like non-evm addresses table? @aminlatifi 
we may have some of them for upcoming grants

CarlosQ96:
We have a addresses table for projects we can re-use it. I think this might take just a few **3-4 hours.** 
And modify the api where we edit addresses. So ill add another 1-2 hours.

Total: **6 hours.** (Max 8 hours -> a day rightfully tested)

This issue is part of a bigger one: https://github.com/Giveth/impact-graph/issues/1188 which is making a donation to a project address.

PR #1196, opened 2023-12-12
Project Solana Address Support

aminlatifi:
#1187


src/resolvers/projectResolver.ts
```diff
@@ -475,140 +482,84 @@ export class ProjectResolver {
...
+          return;
+        case FilterField.AcceptFundOnSolana:
+          acceptFundOnSolanaSeen = true;
+          return;
+        default:
+          return query.andWhere(`project.${filter} = true`);
+      }
+    });
+
+    if (networkIds.length > 0 || acceptFundOnSolanaSeen) {
```
CarlosQ96, line 555: We will have to limit it in the frontend. Like you said both are not posible

CarlosQ96 (approved):
Looks great functionality wise, I will add the tokens in my feature branch for the donation.

src/resolvers/projectResolver.ts
aminlatifi: No, I believe we don't need to limit the front end. Before this, we had an `AND` conjunction between accepting funds on chains, so choosing multiple ones had no meaning. After this, we will have an `OR` conjunction between them.

src/resolvers/projectResolver.test.ts
```diff
@@ -1262,6 +1277,145 @@ function allProjectsTestCases() {
...
+      chainType: ChainType.SOLANA,
+      networkId: 0,
+      isRecipient: true,
+    });
+    await solanaAddress.save();
+
+    const result = await axios.post(graphqlUrl, {
+      query: fetchMultiFilterAllProjectsQuery,
+      variables: {
+        filters: ['AcceptFundOnMainnet', 'AcceptFundOnSolana'],
```
aminlatifi, line 1499: @CarlosQ96 Look at this test, all other test scenarios we have had before only test filters with single critrion.

mohammadranjbarz (approved):
Thanks @aminlatifi  for implementing Solana and refactoring some old codes.
LGTM, but I think it's better to add `chainType` to admin bro then admins can filter donations/tokens based on chainType

https://github.com/Giveth/impact-graph/blob/248356696da9a949101d3dd4e911b33159c97154/src/server/adminJs/tabs/tokenTab.ts#L174

https://github.com/Giveth/impact-graph/blob/248356696da9a949101d3dd4e911b33159c97154/src/server/adminJs/tabs/donationTab.ts#L372

aminlatifi:
Closed and made #1200

aminlatifi:
> Thanks @aminlatifi for implementing Solana and refactoring some old codes. LGTM, but I think it's better to add `chainType` to admin bro then admins can filter donations/tokens based on chainType
> 
> https://github.com/Giveth/impact-graph/blob/248356696da9a949101d3dd4e911b33159c97154/src/server/adminJs/tabs/tokenTab.ts#L174
> 
> https://github.com/Giveth/impact-graph/blob/248356696da9a949101d3dd4e911b33159c97154/src/server/adminJs/tabs/donationTab.ts#L372

Let's solve it as a separate issue, because it's not clear for me how it would be!

Issue #1171, opened 2023-10-25
Fetch status of Solana donations/transactions 


divine-comedian:
We will need to be able to track & verify successful donations made on the Solana network via the Giveth dapp - similar to how we currently handle Ethereum donations. 

@jainkrati needs to check in with @CarlosQ96 or @mohammadranjbarz to see how we handle this on the dapp with Ethereum and see if it presents any challenges with doing this on Solana

jainkrati:
Most likely this "get transaction details" api will help us confirm exact amount transfered, token type transferred https://docs.solana.com/api/http#gettransaction

jainkrati:
@aminlatifi please use the above API and let me know if you need to discuss on this. I am assuming that you are very comfortable with Impact graph; else we can get pointers from Mohammad or Carlos

aminlatifi:
@jainkrati @divine-comedian I played with the JSON RPC node and could have progressed to some extent. But, it's not still straightforward to me how to deduct transferred value! I made test transfers and got API output. It's a simple 1 native token transfer!

```json
{
    "jsonrpc": "2.0",
    "result": {
        "blockTime": 1701350064,
        "meta": {
            "computeUnitsConsumed": 150,
            "err": null,
            "fee": 5000,
            "innerInstructions": [],
            "loadedAddresses": {
                "readonly": [],
                "writable": []
            },
            "logMessages": [
                "Program 11111111111111111111111111111111 invoke [1]",
                "Program 11111111111111111111111111111111 success"
            ],
            "postBalances": [
                30589135362421630,
                1000000000,
                1
            ],
            "postTokenBalances": [],
            "preBalances": [
                30589136362426630,
                0,
                1
            ],
            "preTokenBalances": [],
            "rewards": [],
            "status": {
                "Ok": null
            }
        },
        "slot": 262018390,
        "transaction": {
            "message": {
                "accountKeys": [
                    "9B5XszUGdMaxCZ7uSQhPzdks5ZQSmWxrmzCSvtJ6Ns6g",
                    "GEhUKKZeENY1TmaavqvLJ5GbbQs9GkzECFSE2bpjzz3k",
                    "11111111111111111111111111111111"
                ],
                "header": {
                    "numReadonlySignedAccounts": 0,
                    "numReadonlyUnsignedAccounts": 1,
                    "numRequiredSignatures": 1
                },
                "instructions": [
                    {
                        "accounts": [
                            0,
                            1
                        ],
[...]

PR #1209, opened 2023-12-24
Verify Solana transactions

mohammadranjbarz:
Now it just supports **SOL** transfer on Solana chain, for token transfer I will continue in another PR

related to #1171


src/services/chains/solana/transactionService.ts
```diff
@@ -0,0 +1,77 @@
...
+import { normalizeAmount } from '../../../utils/utils';
+import {
+  i18n,
+  translationErrorMessagesKeys,
+} from '../../../utils/errorMessages';
+
+const SOLANA_NODE_RPC_URL = process.env.SOLANA_NODE_RPC_URL as string;
+const solana = new SolanaWeb3.Connection(SOLANA_NODE_RPC_URL);
+
+export const getSolanaTransactionDetailForNormalTransfer = async (params: {
```
aminlatifi:
```suggestion
export const getSolanaTransactionDetailForSolanaTransfer = async (params: {
```

src/services/chains/solana/transactionService.ts
```diff
@@ -0,0 +1,77 @@
...
+  validateTransactionWithInputData,
+} from '../index';
+import { normalizeAmount } from '../../../utils/utils';
+import {
+  i18n,
+  translationErrorMessagesKeys,
+} from '../../../utils/errorMessages';
+
+const SOLANA_NODE_RPC_URL = process.env.SOLANA_NODE_RPC_URL as string;
+const solana = new SolanaWeb3.Connection(SOLANA_NODE_RPC_URL);
```
aminlatifi: Let's not put it here, this will be executed each time the module is imported. It means the whole app will crash if anything happens to it! 
Please move it inside a function. It's not a big deal if we instantiate the connection each time we do the validation.

src/services/chains/solana/transactionService.ts
```diff
@@ -0,0 +1,77 @@
...
+  } catch (e) {
+    logger.error('getSolanaTransactionDetailForNormalTransfer error', e);
+    return null;
+  }
+};
+
+function getTransactionDetailForTokenTransfer(
+  input: TransactionDetailInput,
+): Promise<NetworkTransactionInfo> {
+  throw new Error();
```
aminlatifi:
```suggestion
  throw new Error('Not implemented!');
```

src/services/chains/solana/transactionService.ts
```diff
@@ -0,0 +1,77 @@
...
+}): Promise<NetworkTransactionInfo | null> => {
+  const TRANSFER_INSTRUCTION_TYPE = 'transfer';
+  const SOL_CURRENCY_TYPE = 'SOL';
+
+  try {
+    const result = await solana.getParsedTransaction(params.txHash);
+    const data = result?.transaction?.message?.instructions?.find(
+      instruction =>
+        (instruction as ParsedInstruction)?.parsed?.type ===
+        TRANSFER_INSTRUCTION_TYPE,
```
aminlatifi: Let's filter based on source and destination as well. I am not sure single transfer exists in a Solana transaction

src/services/chains/index.test.ts
```diff
@@ -957,4 +871,47 @@ function getTransactionDetailTestCases() {
...
+        toAddress: 'DvWdrYYkwyM9mnTetpr3HBHUBKZ22QdbFEXQ8oquE7Zb',
+        timestamp: 1702931400 + ONE_DAY,
+        amount,
+      });
+    };
+    await assertThrowsAsync(
+      badFunc,
+      errorMessages.TRANSACTION_CANT_BE_OLDER_THAN_DONATION,
+    );
+  });
```
aminlatifi, line 935: Please add more tests if you can to cover other cases. You can use other transactions, like
3nzHwgxAu7mKw1dhGTVmqzY8Yet3kGWWqP5kr5D2fw1HzqPjqDGDe6xT5PguKXk8nAJcK4GpBEKWw7EzoLykKkCx

aminlatifi (changes requested):

src/services/chains/solana/transactionService.ts
mohammadranjbarz: Done

src/services/chains/solana/transactionService.ts
mohammadranjbarz: Done

src/services/chains/solana/transactionService.ts
mohammadranjbarz: Done

src/services/chains/solana/transactionService.ts
mohammadranjbarz: Good idea, done

src/services/chains/index.test.ts
mohammadranjbarz: @aminlatifi 
I just added another successful test case with your tx, for adding more test cases to cover all scenarios I need more time, does it needed for now or I can do that later?

aminlatifi (approved):

aminlatifi:
Thanks @mohammadranjbarz

CarlosQ96 (approved):
Thanks Mohammad, all changes and feedback LGTM. 
We should test asap, maybe in staging to see if we don't have any edge case.

PR #1224, opened 2023-12-30
Add test and regex for solana token addresses

CarlosQ96:
Add hotfix for solana program address being sent on the params.

### DIFF
--- a/src/utils/validators/graphqlQueryValidators.ts
+++ b/src/utils/validators/graphqlQueryValidators.ts
@@ -20,7 +20,9 @@ const resourcePerDateRegex = new RegExp(
 );
 
 const ethereumWalletAddressRegex = /^0x[a-fA-F0-9]{40}$/;
-const solanaWalletAddressRegex = /^(?:[A-Za-z0-9]{32}|[A-Za-z0-9]{43,44})$/;
+const solanaWalletAddressRegex = /^[A-Za-z0-9]{43,44}$/;
+const solanaProgramIdRegex =
+  /^(11111111111111111111111111111111|[1-9A-HJ-NP-Za-km-z]{43,44})$/;
 const txHashRegex = /^0x[a-fA-F0-9]{64}$/;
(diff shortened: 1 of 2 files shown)


aminlatifi (approved):

src/utils/validators/graphqlQueryValidators.ts
```diff
@@ -20,7 +20,7 @@ const resourcePerDateRegex = new RegExp(
 );
 
 const ethereumWalletAddressRegex = /^0x[a-fA-F0-9]{40}$/;
-const solanaWalletAddressRegex = /^[A-Za-z0-9]{43,44}$/;
+const solanaWalletAddressRegex = /^(?:[A-Za-z0-9]{32}|[A-Za-z0-9]{43,44})$/;
```
aminlatifi: Solana ProgramIds are 43 to 44 character long. There is only one exception, system program id, which is 11111111111111111111111111111111 
So this must work:
const programIdRegex = /^(11111111111111111111111111111111|[1-9A-HJ-NP-Za-km-z]{43,44})$/;

aminlatifi (changes requested):

src/utils/validators/graphqlQueryValidators.ts
```diff
@@ -95,7 +97,11 @@ export const createDonationQueryValidator = Joi.object({
   transactionNetworkId: Joi.string()
     .required()
     .valid(...Object.values(NETWORK_IDS)),
-  tokenAddress: Joi.string().pattern(ethereumWalletAddressRegex),
+  tokenAddress: Joi.when('transactionNetworkId', {
+    is: 0, // if its solana network
+    then: Joi.string().pattern(solanaProgramIdRegex),
+    otherwise: Joi.string().pattern(ethereumWalletAddressRegex),
+  }),
```
mohammadranjbarz, line 104: Nice feature, I didnt know that

mohammadranjbarz (approved):
Thanks @CarlosQ96,
I think we should cover more test cases ( for instance get error when sending solana contract address but networkID is no **0**) but codes look good to me

aminlatifi (approved):

Issue #1172, opened 2023-10-25
[solana-v2]Hash SPL addresses for projects with Solana addresses & add to project information


divine-comedian:
In Solana token transfers work differently than in EVM 

Each wallet has separate "token accounts" for every token 

When a token is transferred from wallet A -> wallet B it is transferred from the corresponding token account of wallet A to the token account associated with wallet B

A wallet owner needs to manually create their token accounts to view & manage tokens they've received

However we are able to know what their given token account address is/will be by hashing together their main account address with the address of the token contract.. 

@jainkrati - can also clarify if I missed or misunderstood any points

jainkrati:
Here is the reference link : https://spl.solana.com/token#example-transferring-tokens-to-another-user

divine-comedian:
I think we also need this:
https://spl.solana.com/associated-token-account

divine-comedian:
- SOL (native token)
- mSOL `mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So`
- USDC `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
- USDT `Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB`
- RAY `4k3Dyjzvzp8eMZWUXbBCjEvwSkkk59S5iCNLY3QrkX6R`
- bSOL `bSo13r4TkiE4KumL71LsHTPpL2euBYLFx6h9HP3piy1`
- AUDIO `9LzCMqDgTKYz9Drzqnpgee3SGa89up3a247ypMj2xrqM`
- MANGO `MangoCzJ36AjZyKwVj3VnYU4GTonjfVEnJmvvWaxLac`
- Coin98 `0xaec945e04baf28b135fa7c640f624f8d90f1c3a6`

jainkrati:
For now we are only supporting donation with SOL token in v1. This is what we agreed @divine-comedian 

In v2 when more tokens are involved, we can use JS implementation given here https://spl.solana.com/token#example-transferring-tokens-to-another-user to calculate the Associated Token Accounts before transfer.

divine-comedian:
@jainkrati Yes that is correct! We can focus on just making SOL donations happen and work on SPL tokens in a later iteration of this feature

MoeNick:
According to our timeline, we have to start v1, I would appreciate assigning it to the backend guys or starting it yourself. @aminlatifi 
[...]

Issue #1175, opened 2023-10-25
Fetch prices of SOL & Solana tokens 


divine-comedian:
Tokens will be 

- SOL (native token)
- mSOL `mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So`
- USDC `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`
- USDT `Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB`
- RAY `4k3Dyjzvzp8eMZWUXbBCjEvwSkkk59S5iCNLY3QrkX6R`
- bSOL `bSo13r4TkiE4KumL71LsHTPpL2euBYLFx6h9HP3piy1`
- AUDIO `9LzCMqDgTKYz9Drzqnpgee3SGa89up3a247ypMj2xrqM`
- MANGO `MangoCzJ36AjZyKwVj3VnYU4GTonjfVEnJmvvWaxLac`
- Coin98 `0xaec945e04baf28b135fa7c640f624f8d90f1c3a6`

WDYT @jainkrati ?

We should figure out if we are getting these from APIs (which could be expensive) or if we want to use a solution similar to what we have with the monoswap repo pricing @mateodaza and @mohammadranjbarz handles most of this.

Here's the repo
https://github.com/Giveth/monoswap

jainkrati:
Fetching latest token prices https://www.quicknode.com/guides/solana-development/3rd-party-integrations/pyth-price-feeds

divine-comedian:
> Fetching latest token prices https://www.quicknode.com/guides/solana-development/3rd-party-integrations/pyth-price-feeds

I see this protocol requires some development in Rust.. Is this the only option or is there easier options for our team that are available in JS?

jainkrati:
Thanks @divine-comedian . Some options in Typescript APIs are https://coinmarketcap.com/api/documentation/v1/#operation/getV2CryptocurrencyQuotesLatest, https://docs.moralis.io/web3-data-api/solana/reference/get-sol-token-price?network=mainnet&address=SRMuApVNdxXokk5GT7XD5cUUgXMBCoAz2LHeuAoKWRt, https://docs.birdeye.so/reference/get_public-price,

mohammadranjbarz:
@jainkrati  @aminlatifi  @MoeNick  @divine-comedian 
[...]

PR #1232, opened 2024-01-04
Verify transfer spl-token on solana, supporting multi chain solana

mohammadranjbarz:
related to #1175

@CarlosQ96  @aminlatifi  I changed something about networks and chains to support dev, test and mainnet in Solana


CarlosQ96:
Pushed graphql validators fix and network error. 
Will Push migration for other tokens

CarlosQ96:
I checked testnet and devnet but I couldn't find some tokens. 

Some tokens have multiple different test ones, like USDc dev and USDC coin dev, usdc saber testnet.

Should we keep the mainnet ones only?

CarlosQ96 (approved):
Thanks Mohammad, your code look good to me. Can review mine too now jajaja

src/utils/validators/graphqlQueryValidators.ts
```diff
@@ -99,9 +99,17 @@ export const createDonationQueryValidator = Joi.object({
     .required()
     .valid(...Object.values(NETWORK_IDS)),
   tokenAddress: Joi.when('transactionNetworkId', {
-    is: 0, // if its solana network
+    is: NETWORK_IDS.SOLANA_MAINNET,
```
mohammadranjbarz: @CarlosQ96  IT doesnt need to add this exception for SOLANA DEVNET and TESTNET as well?

CarlosQ96: Its in the otherwise clauses.

src/provider.ts
```diff
@@ -18,7 +18,11 @@ export const NETWORK_IDS = {
   CELO_ALFAJORES: 44787,
   ETC: 61,
   MORDOR_ETC_TESTNET: 63,
-  SOLANA: 0,
```
aminlatifi, line 21: The SOL token was by migration `migration/1703044586989-addSolanaToken.ts` before this change, will have networkId=0. We must address that

src/utils/validators/graphqlQueryValidators.ts
mohammadranjbarz: Oh you are right

mohammadranjbarz:
Thanks @CarlosQ96  
Your codes look good to me.

src/utils/validators/graphqlQueryValidators.ts
aminlatifi: First, I think there would be a better solution than nesting, like 
```suggestion
    is: Joi.valid(NETWORK_IDS.SOLANA_MAINNET, NETWORK_IDS.SOLANA_DEVNET, NETWORK_IDS.SOLANA_TESTNET),
```

Beyond that, I guess it's better to decide the valid regext to match based on chainType rather than networkId

aminlatifi (changes requested):

src/provider.ts
CarlosQ96: I think we can address it manually, the migration uses the seedtoken file which was updated, so deploying to develop or production will have the correct values.

Ill address it manually in staging.

src/utils/validators/graphqlQueryValidators.ts
CarlosQ96: Done

src/utils/validators/graphqlQueryValidators.ts
```diff
@@ -98,18 +99,10 @@ export const createDonationQueryValidator = Joi.object({
   transactionNetworkId: Joi.string()
     .required()
     .valid(...Object.values(NETWORK_IDS)),
-  tokenAddress: Joi.when('transactionNetworkId', {
-    is: NETWORK_IDS.SOLANA_MAINNET,
+  tokenAddress: Joi.when('chainType', {
+    is: Joi.string().pattern(chainTypeRegex),
```
aminlatifi: It must match only with solana

aminlatifi (changes requested):

src/utils/validators/graphqlQueryValidators.ts
CarlosQ96: Whoops I pushed wrong regex during testing. Fixed.

aminlatifi: So, why regex? :)
It only would match with "SOLANA" value which can be simple equal condition (then remove `chainTypeRegex` variable). Also, it's better to use ChainType.SOLANA instead of repeating "SOLANA" here.

aminlatifi (changes requested):

aminlatifi:
@mohammadranjbarz does this resolve https://github.com/Giveth/impact-graph/issues/1172 as well?

mohammadranjbarz:
> @mohammadranjbarz does this resolve #1172 as well?

Yeah we check the token account for each SOLANA wallet, so yes I think it's handled

src/services/chains/solana/transactionService.ts
```diff
@@ -10,16 +10,33 @@ import {
...
+  } else if (chainId === NETWORK_IDS.SOLANA_TESTNET) {
+    solanaProviders[chainId] = new SolanaWeb3.Connection(
+      process.env.SOLANA_TEST_NODE_RPC_URL as string,
+    );
+  } else {
+    // DEVNET
+    solanaProviders[chainId] = new SolanaWeb3.Connection(
+      process.env.SOLANA_DEVNET_NODE_RPC_URL as string,
+    );
+  }
```
aminlatifi: let's benefit from switch-case for more readable code.

src/services/chains/solana/transactionService.ts
```diff
@@ -62,10 +79,55 @@ export const getSolanaTransactionDetailForSolanaTransfer = async (
...
+    const txInfo = parsedData.parsed.info;
+    if (!txInfo) {
+      return null;
+    }
+    return {
+      from: txInfo.authority,
+
+      // we already check toAddressBalance.owner === params.toAddress
+      to: toAddressBalance?.owner,
+      amount: txInfo.tokenAmount?.uiAmount,
```
aminlatifi: The example I have doesn't hold any value in parsed.info.tokenAmount.uiAmount!
The safer approach would be get receivers balance in preTokenBalances and postTokenBalances, then use their diff here.

aminlatifi (changes requested):

src/services/chains/solana/transactionService.ts
aminlatifi: I got my example by running this
```ts
 const connection = new Connection(clusterApiUrl("devnet"), "confirmed");
  const txInfo = await connection.getParsedTransaction(
    "3m6f1g2YK6jtbfVfuYsfDbhVzNAqozF8JJyjp1VuFDduecojqeCVK4htKnLTSk3qBwSqYUvgLpBTVpeLJRvNmeTg"
  );
  console.log(JSON.stringify(txInfo, null, 2));
  ```

Comment (mohammadranjbarz) [reply src/services/chains/solana/transactionService.ts]: let me write a test case for your example, to see what happens

Comment (aminlatifi) [reply src/services/chains/solana/transactionService.ts]: Some info about the transfer
From wallet: 26Aks2rN6mfqxdYRXKZbn8CS4GBv6fCMGFYfGWvfFfcx
To wallet: 7TJgw4hDHh5wdKep3EsBkGMSvtf9LsxdXf89LA48uHoq
Mint address: "8LDBhHJB7oMAjkJaetXa4njjetUVWDRTqvzkmhFQjgeK"
From Token Account: 9VsnTMDpnSoMhYx2FQpXAs5f4a1UNh9WxUBLL6LpiYZo
To Token Account: HhYuUyc9T7V758kfR3f8PG8pKz38MsDvhkQtEEtbVBrV
Mint tx: 5NDEXP9F8UncZqZsMcssSRjX4YMHEz2iZyDgfCcPdSjzcAP1u8Kg2Js7BCKz4taiLSsK88DqdiFnCEZZLfyo1fpg
Transfer tx: 3m6f1g2YK6jtbfVfuYsfDbhVzNAqozF8JJyjp1VuFDduecojqeCVK4htKnLTSk3qBwSqYUvgLpBTVpeLJRvNmeTg

Comment (mohammadranjbarz) [reply src/services/chains/solana/transactionService.ts]: Thanks I'm working on it

File: src/services/chains/solana/transactionService.ts (hunk 4 above)
Comment (mohammadranjbarz) [reply src/services/chains/solana/transactionService.ts]: @aminlatifi  Yes you are right, Done

File: src/utils/validators/graphqlQueryValidators.ts (hunk 3 above)
Comment (mohammadranjbarz) [reply src/utils/validators/graphqlQueryValidators.ts]: I changed it, done.

File: src/utils/locales/en.json [hunk 6]
```diff
@@ -1,105 +1,107 @@
...
+	"YOU_ARE_NOT_THE_OWNER_OF_THIS_SOCIAL_PROFILE": "You are not the owner of social profile",
+	"ERROR_IN_GETTING_ACCESS_TOKEN_BY_AUTHORIZATION_CODE": "Error in getting accessToken by authorization code",
+	"ERROR_GIVPOWER_BOOSTING_FIRST_PROJECT_100_PERCENT": "First project boosting value must be 100%",
+	"ERROR_GIVPOWER_BOOSTING_INVALID_DATA": "Invalid data",
+	"ERROR_GIV_POWER_BOOSTING_SUM_IS_GREATER_THAN_MAXIMUM": "Giv power boosting summation is greater than 100",
+	"ERROR_GIVPOWER_BOOSTING_MAX_PROJECT_LIMIT": "Number of boosted projects exceeds limit",
+	"REGISTERED_NON_PROFITS_CATEGORY_DOESNT_EXIST": "There is not any category with name registered-non-profits, probably you forgot to run migrations",
+	"PROJECT_UPDATE_CONTENT_LENGTH_SIZE_EXCEEDED": "Content length exceeded",
+	"INVALID_TOKEN_ADDRESS": "INVALID_TOKEN_ADDRESS",
+	"Invalid tokenAddress": "Invalid tokenAddress"
```
aminlatifi:
```suggestion
	"INVALID_TOKEN_ADDRESS": "Invalid tokenAddress"
```

src/utils/errorMessages.ts
```diff
@@ -231,6 +232,7 @@ export const translationErrorMessagesKeys = {
     'CATEGORIES_MUST_BE_FROM_THE_FRONTEND_SUBSELECTION',
   INVALID_TX_HASH: 'INVALID_TX_HASH',
   INVALID_TRANSACTION_ID: 'INVALID_TRANSACTION_ID',
+  INVALID_TOKEN_ADDRESS: 'Invalid tokenAddress',
```
aminlatifi:
```suggestion
  INVALID_TOKEN_ADDRESS: 'INVALID_TOKEN_ADDRESS',
```

aminlatifi (changes requested):

aminlatifi (approved):

PR #1236, opened 2024-01-10
Import lost donations into DB cronjob 

CarlosQ96:
Related to: https://github.com/Giveth/giveth-dapps-v2/issues/3526


CarlosQ96:
Will not import multisig donations, those were not part of the round and we just released that. There should not be any or any communication about it. Not sure how they were done.

src/services/cronJobs/importLostDonationsJob.ts
```diff
@@ -0,0 +1,267 @@
...
+    transaction.blockNumber as number,
+  );
+
+  const transactionTo = transaction.to;
+  const amount = ethers.utils.formatEther(transaction.value);
+
+  return {
+    from: transaction.from,
+    timestamp: block.timestamp as number,
+    to: transactionTo as string,
```
mateodaza, line 237: Awesome job man. @CarlosQ96 
Just make sure you're checking the from and to values on the Transfer event from the logs instead. Then you're making sure you're grabbing the right values each time. Check the multisig code for it, we did something similar

mateodaza (approved):

src/services/cronJobs/importLostDonationsJob.ts
CarlosQ96: Thanks man

CarlosQ96:
@aminlatifi @mohammadranjbarz can review please so we import the donations. Ill handle the optimism ones first then the rest. This should work for all.

mohammadranjbarz:
> @aminlatifi @mohammadranjbarz can review please so we import the donations. Ill handle the optimism ones first then the rest. This should work for all.
@CarlosQ96 
Name PR should not be 
**Import lost donations into DB cronjob** ?

src/services/cronJobs/importLostDonationsJob.ts
```diff
@@ -0,0 +1,279 @@
...
+        givbackFactor,
+        projectRank,
+        powerRound,
+        bottomRankInRound,
+        status: 'verified',
+        anonymous: false,
+        segmentNotified: true,
+        isTokenEligibleForGivback: tokenInDB?.isGivbackEligible,
+        isProjectVerified: project.verified,
+        qfRoundId: qfRound?.id,
```
mohammadranjbarz, line 197: @CarlosQ96  Should not add qfRound to all missed donations, we should check if project is connected to qfRound, then add `qfRoundId` to donation ( and its nice to have a test case for that scenario as well)

mohammadranjbarz (changes requested):

src/services/cronJobs/importLostDonationsJob.ts
mohammadranjbarz: I did it myself just please take a look yourself too

mohammadranjbarz (approved):
Thanks @CarlosQ96 
I skipped the changes for qfRound as it's just for our tx hashes

Issue #1247, opened 2024-01-16
P0 - Save failed donations reported by back-up service


MoeNick:
As we now have a backup service that saves the lost donations in MongoDB, we need a flow to save them in the impact-graph again and mark them as saved. 

Questions:
- When do we save? what will be the flow?
- Do we have the missed items in data like price? How we could fetch it?
- How do we mark them as save and what information we will add for tracking? 
- 

I invite @aminlatifi @jainkrati @mhmdksh @mohammadranjbarz @CarlosQ96 and @MohammadPCh @geleeroyale to discuss here and make a decision ASAP as it's a p0 story.

MoeNick:
FYI: @laurenluz @divine-comedian

laurenluz:
If it's possible, I would ideally like to add the saved donations at least once/day so that we can have up to date information about round stats every day.

Just my 2 cents :) looking forward to hearing from the masterminds

jainkrati:
- When do we save? what will be the flow?  : We can save every 8 hours. That would be ideal for most time zones.Flow could be: Backup service dumps transactions to be backed up in redis from where we add to DB every 8 hours and then delete. @mhmdksh what are your thoughts here? 
- Do we have the missed items in data like price? How we could fetch it? : There are historical token prices which we can fetch from some [chainlink historical price feed api](https://docs.chain.link/quickstarts/historical-price-feeds-api) or with an API like https://docs.moralis.io/web3-data-api/evm/how-to-get-historical-erc20-token-price; 
- How do we mark them as save and what information we will add for tracking? : We could mark all mongoDB entries as 'Pending' by default. This is a one time operation. Then we can mark them 'Backed' as we create the backup. cc @aminlatifi @mohammadranjbarz

aminlatifi:
> * When do we save? what will be the flow?  : We can save every 8 hours. That would be ideal for most time zones.Flow could be: Backup service dumps transactions to be backed up in redis from where we add to DB every 8 hours and then delete. @mhmdksh what are your thoughts here?
> * Do we have the missed items in data like price? How we could fetch it? : There are historical token prices which we can fetch from some [chainlink historical price feed api](https://docs.chain.link/quickstarts/historical-price-feeds-api) or with an API like https://docs.moralis.io/web3-data-api/evm/how-to-get-historical-erc20-token-price;
[...]

PR #1253, opened 2024-01-18
Import from Donation Backup Service

aminlatifi:
https://github.com/Giveth/impact-graph/issues/1247


src/services/cronJobs/backupDonationImportJob.ts
```diff
@@ -0,0 +1,334 @@
...
+  while (donations.length > 0) {
+    for (const donation of donations) {
+      try {
+        await createBackupDonation(donation);
+        await getDonationSaveBackupAdapter().markDonationAsImported(
+          donation._id,
+        );
+      } catch (e) {
+        logger.error(`donation error with id ${donation._id}: `, e);
+        logger.error('donation error with params: ', donation);
```
aminlatifi, line 46: There is a point here, if importing an item gets an error, it will be fetched in the next round of importing again. So, it will endlessly be processed in each iteration.

src/services/cronJobs/backupDonationImportJob.ts
```diff
@@ -0,0 +1,334 @@
...
+      }
+    }
+    skip += limit;
+    donations =
+      await getDonationSaveBackupAdapter().getNotImportedDonationsFromBackup({
+        limit,
+        skip,
+      });
```
aminlatifi: There is a serious issue here, the items we import above, will not match the query filter any more (their imported field will have true value). So, we shouldn't increase the skip value here naively, just increase by number of items which we got error in processing above.

src/services/cronJobs/backupDonationImportJob.ts
mohammadranjbarz: @aminlatifi 
How about adding another field(like isImported) for this kind of donations and set it true in this case, and not get them again

src/services/cronJobs/backupDonationImportJob.ts
mohammadranjbarz: Good point

src/services/cronJobs/backupDonationImportJob.ts
aminlatifi: The name can be `importError` I think to hold the error message as well. Before going for that, we must be sure this importing is a deterministic process. I mean, if it fails here, it will fail every next time we try. If it holds, that would be ok to mark the items got error in importing and ignore them, but if not we must allow them to be processed in next iteration.

mohammadranjbarz: Cool , I will do it

CarlosQ96: I see the skip was removed. Could simply add a counter that increases when it enters the catch() statement. And on each iteration reset it to 0. And skip based on that number.

But at the same time we could simply increase the limit and let the process run every hour or so. It shouldn't be a heavy process.
I was checking failed donations, and even on the previous round where we had big errors, the amount is never a huge number.

CarlosQ96 (approved):
Thanks Mohammad, those are the changes I was wanted. Better error handling, interfaces and mocks. 
I think we should probably merge, check there isn't any issue in staging.

PR #1254, opened 2024-01-21
Add more spl-token transfer test cases

mohammadranjbarz:


mohammadranjbarz:
I closed them as we don't need them right now

PR #1255, opened 2024-01-21
Donation save backup import revision

mohammadranjbarz:
I applied these changes on @CarlosQ96  PR

* Add adapter and interface for saveFailedDonationAdapter
* Get createdAt time of donation from mongo _id
* Remove saveFailedDonationAdapter test cases, because we should use mock adapter for testing and there is no point in adding test cases for some mock functions
* Put DB credential in config


aminlatifi (approved):

CarlosQ96 (approved):

CarlosQ96:
can merge to the branch thanks!

PR #1258, opened 2024-01-22
Fix-verify-donation-job-queue-add

aminlatifi:
Ref https://github.com/Giveth/giveth-dapps-v2/issues/3637


CarlosQ96 (approved):
Looks, ok. I merged master to staging to its weird it still showing it as changes instead of just your PR commit. Can merge.

mohammadranjbarz (approved):
@aminlatifi  Code seems good, but I don't know how it solves the problem

aminlatifi:
> @aminlatifi Code seems good, but I don't know how it solves the problem


When an issue happens in verification, the donation is added multiple times!
So, the verify process is called plenty of times in this case

PR #1269, opened 2024-01-28
Disable jobId on verifyDonationsQueue adding process

aminlatifi:


mohammadranjbarz (approved):
Thanks @aminlatifi 
It makes sense

PR #1273, opened 2024-01-28
Add jobId back to verifyDonationsQueue

aminlatifi:
Added back jobId to verifyDonationsQueue, along with removeOnComplete and removeOnFail


mohammadranjbarz (approved):
Nice, I wasn't aware of these options
LGTM

Issue #1142, opened 2023-10-04
[3rd April] Save & Retrieve recurring donation data for projects


divine-comedian:
We'll need to save & fetch certain information to show user's & project owners data regarding a given donation stream. In order to query the information we need to check the stream between the donor & recipient address (or anchor address (allo protocol))

When a superfluid stream is created through the Giveth UI we should save the following information to our back-end:

- Sender 
- Receiver
- Super Token
- ProjectId
- txHash (for batch call or createFlow tx) - e.g. https://mumbai.polygonscan.com/tx/0x7845ac65b07ece9a4d2da0c94d1fe3032e62a0ce840f77e4b4c97f863ddbd056

We identify the stream on the subgraph or other superfluid service by the sender, receiver & token and we can alias it to Giveth by the saved projectId in our DB. 

When we need to fetch the data in the case of showing recurring donations for donors & project owners (public & private view) this will allow us to know exactly what to query. 

### Optimism
https://console.superfluid.finance/subgraph?_network=optimism-mainnet

### OP Goerli
https://console.superfluid.finance/subgraph?_network=optimism-goerli

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOG9HQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--20a570828b8c57a3b45df9e3ea8412a6c4425743/image.png)

related to https://github.com/Giveth/giveth-dapps-v2/issues/3093 , https://github.com/Giveth/giveth-dapps-v2/issues/3092

divine-comedian:
@jainkrati   - I'm second guessing myself on this one - Do we need back-end services on this one or can we handle it all from the front-end with the subgraph to get recurring donation (stream) data?

mohammadranjbarz:
@divine-comedian  Good question, if frontend can communicate with subgraph directly, in the backend side we just need to fetch donation data fro superfluid and insert new record in our DB?

divine-comedian:
> @divine-comedian Good question, if frontend can communicate with subgraph directly, in the backend side we just need to fetch donation data fro superfluid and insert new record in our DB?

[...]

PR #1283, opened 2024-01-31
Implement recurringDonationsByProjectId webservice

mohammadranjbarz:
Related to #1142

@MohammadPCh  needs this web service and some other web services that I should implement them after this one


src/entities/recurringDonation.ts
```diff
@@ -31,14 +31,33 @@ export class RecurringDonation extends BaseEntity {
...
   @Field()
-  @Column()
+  @Column({ nullable: false })
   networkId: number;
 
+  @Field()
+  @Column({ nullable: false })
+  amount: number;
+
+  // daily, weekly, monthly, yearly
```
CarlosQ96, line 41: We should add an ENUM for this

CarlosQ96 (approved):
LGTM I would make an ENUM for the values in the interval field so we have fixed constants. Feel free to make this change and merge

PR #1295, opened 2024-02-05
add mpeth prices fetch and token

CarlosQ96:
Related: https://github.com/Giveth/giveth-dapps-v2/issues/3639


CarlosQ96:
I did the implementation with the subgraph Mateo and me found  from velodrome (only api that worked): https://thegraph.com/explorer/subgraphs/A4Y1A82YhSLTn998BVVELC8eWzhi992k4ZitByvssxqA?view=Playground&chain=arbitrum-one

Coingecko doesnt have mpeth listed, cryptocompare api doesn't either and monoswap works with uniswap apis which don't have the token either @mohammadranjbarz @jainkrati 

@laurenluz Said she contacted them to see if we can get them listed in coingecko.

 In the meantime this implementation works, if we wanna move forward, please review and merge. Else we can wait for tomorrow to see coingecko response.

src/services/mpEthPriceService.ts
```diff
@@ -0,0 +1,37 @@
...
+        'Content-Type': 'application/json',
+      },
+      timeout: axiosTimeout,
+    });
+    return Number(result.data.data.tokens[0].lastPriceUSD);
+  } catch (e) {
+    logger.error('fetching Giv Price fetchGivPrice() err', e);
+    throw e;
+  }
+};
```
mohammadranjbarz, line 37: It would be good if you can write a test case for this function
(Just to make sure that we get price and not throw error, and also the price is greater than zero, it would be enough)

mateodaza (approved):
Just as discussed, thanks. Looking good

PR #1314, opened 2024-02-13
Make master and staging pipeline look like each other

aminlatifi:


mhmdksh (commented):
This will not work for 2 reasons:

1. Although these two are now similar, the docker-compose versions on our staging and production environments are different and so is the `docker-compose` or `docker compose` command
To explain more what I mean, check the screenshots below:
Staging:
image

Production:
image

3. This will not fix the pipeline itself since the failure is in the tests and migrations. This will not guarantee a working pipeline.

PR #1325, opened 2024-02-14
Remove chainType from older seed token values

aminlatifi:
Removed chainType from older seed token values to fix migration issues.
As older schema doesn't have chainType and evm is default value of chainType in new schema, it won't cause issue


mohammadranjbarz (approved):
@aminlatifi Makes sense, thanks
LGTM.

jainkrati (approved):

PR #1326, opened 2024-02-14
Draft donation patch

aminlatifi:
Added guard to only support draft donation on evm chain
Don't match draft donations have been made in last one minute
Return old draft donation on repetitive save


jainkrati (approved):

src/resolvers/donationResolver.ts
```diff
@@ -65,6 +65,12 @@ import {
   getDefaultSolanaChainId,
 } from '../services/chains';
 import { markDraftDonationStatusMatched } from '../repositories/draftDonationRepository';
+import {
+  DRAFT_DONATION_STATUS,
+  DraftDonation,
+} from '../entities/draftDonation';
+
+const draftDonationEnabled = process.env.ENABLE_DRAFT_DONATION === 'true';
```
mohammadranjbarz, line 73: @aminlatifi  Please add ENABLE_DRAFT_DONATION to `.env.example`

mohammadranjbarz (changes requested):
Seems good to me, just please add `ENABLE_DRAFT_DONATION` to `.env.example`

src/resolvers/donationResolver.ts
aminlatifi: It's already there!

aminlatifi: It has been added before.

PR #1329, opened 2024-02-16
F: add project fraud entity

CarlosQ96:
Related to : https://github.com/Giveth/giveth-dapps-v2/issues/3752

### DIFF
--- a/migration/1708096213854-addProjectFraudEntity.ts
+++ b/migration/1708096213854-addProjectFraudEntity.ts
@@ -1,0 +1,78 @@
+import {
+  MigrationInterface,
+  QueryRunner,
+  Table,
+  TableForeignKey,
+} from 'typeorm';
+
+export class addProjectFraudEntity1708096213854 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.createTable(
+      new Table({
+        name: 'project_fraud',
+        columns: [
+          {
+            name: 'id',
+            type: 'int',
+            isPrimary: true,
+            isGenerated: true,
+            generationStrategy: 'increment',
+          },
+          {
+            name: 'confirmedFraud',
+            type: 'boolean',
+            isNullable: false,
+            default: false,
+          },
+          {
+            name: 'projectId',
+            type: 'int',
+          },
+          {
+            name: 'qfRoundId',
+            type: 'int',
+          },
+        ],
+      }),
+      true,
+    );
+
+    await queryRunner.createForeignKey(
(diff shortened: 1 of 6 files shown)


migration/1708096213854-addProjectFraudEntity.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  MigrationInterface,
+  QueryRunner,
+  Table,
+  TableForeignKey,
+} from 'typeorm';
+
+export class addProjectFraudEntity1708096213854 implements MigrationInterface {
```
jainkrati, line 8: why the specific number in the class name?

src/server/adminJs/tabs/projectFraudTab.ts
```diff
@@ -0,0 +1,165 @@
...
+    const { projectId, qfRoundId, csvData, confirmedFraud } = request.payload;
+    if (csvData) {
+      // Parse the CSV data
+      const jsonArray = await csv().fromString(csvData);
+
+      // Validate and extract all unique walletAddresses
+      const slugs: string[] = [];
+      for (const obj of jsonArray) {
+        if (!obj.walletAddress || !obj.qfRoundId) {
+          throw new Error('Missing slug for project');
```
jainkrati: is it possible to log projectID in this error log

jainkrati (changes requested):
overall looks good. Once comments are resolved, pls dm me.

src/server/adminJs/tabs/projectFraudTab.ts
CarlosQ96: Thanks I made a mistake here ill log it correctly.

migration/1708096213854-addProjectFraudEntity.ts
CarlosQ96: Those are autogenerated by the framework typeorm when creating new migration, we use this command:
```
typeorm migration:create --dir ./migration --name addProjectFraudEntity
```
Its a number based on the date of creation of the migration.

src/server/adminJs/tabs/projectFraudTab.ts
CarlosQ96: Since the value is missing I cant log it, we are exporting a csv with (qfroundId, slug) of a project.
I will log instead the ROW of the csv that is missing. So the user knows which is wrong.

CarlosQ96: I used slugs because comms team use it regularly

jainkrati (approved):

PR #1332, opened 2024-02-18
add project weight to qfround export spreadsheet

CarlosQ96:
Related: https://github.com/Giveth/giveth-dapps-v2/issues/3687


aminlatifi (commented):
@CarlosQ96 should not this flag be qf round specific? or it's will be a general flag true/false on all qfRounds.

jainkrati (approved):

PR #1346, opened 2024-02-20
fix Ortto totalDonated decimal points

RamRamez:
#1290


mohammadranjbarz (approved):
Thanks @RamRamez  LGTM

PR #1348, opened 2024-02-21
Remove all things about user_passport_score

mohammadranjbarz:
related to https://github.com/Giveth/giveth-dapps-v2/issues/3719#issuecomment-1956310894


aminlatifi (changes requested):
We should not modify or delete already executed migrations.

mohammadranjbarz:
> We should not modify or delete already executed migrations.

You are right, I brought it back

aminlatifi (approved):

PR #1350, opened 2024-02-21
Hotfix/image upload

aminlatifi:
Fixed pinata image upload issue

- https://github.com/Giveth/giveth-dapps-v2/issues/3779


aminlatifi:
Forked from master, so we can be pushed there as a standalone hotfix

src/middleware/pinataUtils.ts
```diff
@@ -2,65 +2,44 @@
...
-  return Axios.post('https://api.pinata.cloud/pinning/pinFileToIPFS', data, {
-    maxContentLength: Infinity, // this is needed to prevent Axios from throw error with large files
-    headers: {
-      'Content-Type': `multipart/form-data; boundary=${data._boundary}`,
-      pinata_api_key: config.get('PINATA_API_KEY') as string,
-      pinata_secret_api_key: config.get('PINATA_SECRET_API_KEY') as string,
-    },
+export const pinFile = (
+  file: ReadableStream,
+  filename: String = 'untitled',
```
jainkrati, line 24: suggestion: lets name it something like temp_giveth_project_image so that we can debug later if needed in some edge case

jainkrati (approved):
Looks good. What was the exact API error?

src/middleware/pinataUtils.ts
aminlatifi: That can be set on the caller side.... it's a general function used on any kind of upload

mohammadranjbarz (approved):
Thanks @aminlatifi  for troubleshooting and fixing this bug

Issue #1291, opened 2024-01-31
Add/remove tag to user in Ortto when they are added/removed to QF round


divine-comedian:
When an admin is using the admin panel it is possible to add one or many projects into an active QF round, as well as remove them from an active qf round
![image](https://github.com/Giveth/impact-graph/assets/67759413/44ec55f6-3530-43d2-85c8-d84c17901f0b)

in the current situation it isn't easy to track the project owners and email them and involves a ton of manual work in spreadsheets. 

## Adding a project

when an admin calls add project to qf round we should call this API (same as #1290)  :
https://help.ortto.com/a-257-create-or-update-one-or-more-people-merge#person-field-id-format

get the following information about the user: 
- userId (of project owner)

and add one tag to the person
- QF Project Owner `<roundName>`

if more than one project is added at once you can handle it by having an array of "people" objects - this is explained in the documentation link above.

## Removing a project 
In this case we need to reset or erase any information added in the previous call
![image](https://github.com/Giveth/impact-graph/assets/67759413/2e27cac6-c8c7-436d-99d3-ced5ba8eaa51)

get the following information:
- email

and remove this tag from the person:
- QF Project Owner `<roundName>`

divine-comedian:
get rid of qfProjectTitle and qfProjectLink to keep the email generic

RamRamez:
@maryjaf to test and verify

maryjaf:
By adding project to a qf round a new tag like below pics will be added to project owner 
and by removing project from QF round this tag will be removed 

<img width="1523" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdytSQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--21ed58c848209348945d7690fbea4d67b0e2cd72/image.png" alt="image.png" />

<img width="1116" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdzZSQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0a15a68bf6633e54892748794d34aacd19018697/image.png" alt="image.png" />

PR #1356, opened 2024-02-22
Add/remove tag to user in Ortto when they are added/removed to QF round

RamRamez:
#1291


mohammadranjbarz (approved):
Thanks @RamRamez LGTM

PR #1373, opened 2024-02-27
fix showing native token symbol on donation received

RamRamez:
[#3800](https://github.com/Giveth/giveth-dapps-v2/issues/3800)


CarlosQ96 (approved):
Approved.

PR #1383, opened 2024-03-03
Remove amount, interval and add flowRate

mohammadranjbarz:


MohammadPCh (approved):
@mohammadranjbarz LGTM, Thanks

Issue #1384, opened 2024-03-03
Exclude donations from purple list addresses in calculating matching fund


mohammadranjbarz:
As @laurenluz  said in the discord
> we should exclude from matching donations make from purple list addresses... but only if the purple list address corresponds to a project that is listed on the dapp (no draft/unlisted projects

divine-comedian:
@mohammadranjbarz is this completed now? I can see the PR is merged.

laurenluz:
yessir

PR #1385, opened 2024-03-03
Exclude purple list from actual matching fund

mohammadranjbarz:
related to #1384

### DIFF
--- a/migration/1709468854359-project_actual_matching_v6.ts
+++ b/migration/1709468854359-project_actual_matching_v6.ts
@@ -1,0 +1,102 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ProjectActualMatchingV61709468854359
+  implements MigrationInterface
+{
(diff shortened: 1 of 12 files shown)


migration/1709468854359-project_actual_matching_v6.ts
```diff
@@ -0,0 +1,102 @@
...
+                        AND p3."listed" = true
+                    )
+                    AND NOT EXISTS (
+                        SELECT 1
+                        FROM public.project_address pa
+                        JOIN public.project p4 ON p4.id = pa."projectId"
+                        WHERE LOWER(d2."fromWalletAddress") = LOWER(pa.address)
+                        AND p4."verified" = true
+                        AND p4."listed" = true
+                    )
```
CarlosQ96, line 61: Both queries do the same, as they exclude fromWalletAddresses that are in the project_address table. The Not Exists is more efficient, so I think we can remove the not in.

CarlosQ96 (changes requested):

Issue #1303, opened 2024-01-31
Create a new table for managing sybil addresses in the backend


laurenluz:
Right now, we only have one tag "known sybil"... but each round will; have its own sybil address set, and they will not always be the same and we should not always punish an address for multiple rounds if they were sybil in one...

This is the list of sybil addresses we have so far: https://docs.google.com/spreadsheets/d/1eUHSmPJ1QaWWWFeVcxYQJBtfnPIzxiWicylxBaZ2d3M/edit#gid=1145721990

We should create some system for seeing "all sybils" OR "sybils from JUST giving season"

jainkrati:
Thanks for assigning me @MoeNick .  Could you move this to Impact Graph

MoeNick:
I moved it to impact-graph @jainkrati

CarlosQ96:
I can take this, ill have to add a middle table between qfround and user and the inner join it.

laurenluz:
is this something @maryjaf can test @CarlosQ96 @jainkrati ?

soon we'll need to make a list of sybil addresses for polygon & get them in the db. I also want to make sure that we are making the appropriate adjustments to the "export QF round data" button in admin bro, so that it is eliminating "known sybils" from that QF round only.

how do we manage these lists? is there something in admin bro now?

CarlosQ96:
@maryjaf in the adminJs there is a new entity called Sybil you can add items to the list and filter them. Can be tested that way. When generating the qfround data from the qfround tab => button, it will exclude them.

The csv import/export is another issue. As that takes more time.

maryjaf:
I've sent a request for access to this google sheet  to complete the test
@laurenluz 

https://docs.google.com/spreadsheets/d/1eUHSmPJ1QaWWWFeVcxYQJBtfnPIzxiWicylxBaZ2d3M/edit#gid=1145721990

laurenluz:
@CarlosQ96 is this already on production? because I think the spreadsheet only populates w/ data from prod, not staging...

is there a separate "master" sheet for staging? or actually, is it the same spreadsheet as this one? https://docs.google.com/spreadsheets/d/1Wk8_bG75uQGZnVywos2Ii2t-4QosHoC0pRkMlZ85Ga8/edit#gid=1770697429

maryjaf:
I did these steps but I couldn't see the expected result 
please check the scenario,did I do the right flow ?

-roundId : 77 

-I created a new sybil record with this data 

[...]

PR #1391, opened 2024-03-05
1303 hotfix fix sybil address add single item

mohammadranjbarz:
related to https://github.com/Giveth/impact-graph/issues/1303#issuecomment-1977398621
I added wallet address instead of userId in creating single sybil item page, so it's easier in UX, I also added some test cases for that.

Screenshot 1402-12-15 at 11 17 27 in the morning

### DIFF
--- a/migration/1709633770434-add_wallet_address_to_sybil_table.ts
+++ b/migration/1709633770434-add_wallet_address_to_sybil_table.ts
@@ -1,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddWalletAddressToSybilTable1709633770434
+  implements MigrationInterface
+{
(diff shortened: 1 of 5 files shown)


mohammadranjbarz:
@CarlosQ96 It's on staging right now
@maryjaf  Can you approve it to merge the hotfix on production?

maryjaf:
yeah I tested and it seems it's ok now 
but I have a question about error handling that I asked in chat but I think Lauren hasn't seen it 

"The messages of 
-User not found or
 -Sybil item for this user and qfRound is in the DB already
are displayed if the records are entered individually.
And if the CSV file is filled , the error will not be displayed but the expected result would be happened, for example if one of them is repetitive or user not found this record isn't added in the list
Is this your expected result and approved? "

CarlosQ96 (approved):
LGTM!

Issue #932, opened 2023-03-28
Add & automate update reminders to notification center & emails for verified project owners


WhyldWanderer:
Some project owners either don't check their email accounts or have unsubscribed from Giveth emails.
The verification team thinks that it is important to remind verified project owners that they need to add an update so that they dont lose their verified status.

Two notifications to be added to the notification center - these would be sent to verified projects 

45 days without an update - friendly reminder: "Don't forget to add regular updates to your project to keep your verified status! - Learn more"

90 days without an update - "You have 14 days to add an update to avoid badge expiration - Learn more"

Something like that.

MoeNick:
if @CarlosQ96 is busy this week you may like to take it @mohammadranjbarz , can you take a look?

NikolaCreatrix:
Hey hi guys, I would love to check in here to see if there is any chance to estimate when this can be implemented, please. @MoeNick @CarlosQ96 or @mohammadranjbarz

mohammadranjbarz:
> Hey hi guys, I would love to check in here to see if there is any chance to estimate when this can be implemented, please. @MoeNick @CarlosQ96 or @mohammadranjbarz

Hey @NikolaCreatrix @MoeNick  I haven't looked over this issue, but as we need to modify something in notification-center and impact-graph I think it would last about 2-3 days( assuming copies are ready and there is no blocker)
@CarlosQ96  WDYT?

MoeNick:
Let's plan for it next week.

divine-comedian:
@MoeNick here is the template
Can you add in the link for the template for this
https://www.figma.com/file/hNqwpebu4JCy1XIed2c862/Giveth-Public?type=design&node-id=23-48916&mode=design&t=7GnTh1lZ7WYORXWj-4

MoeNick:
I talked with @mohammadranjbarz and he will take care of web and email notification of this issue. FOR THIS WEEK

mohammadranjbarz:
> Some project owners either don't check their email accounts or have unsubscribed from Giveth emails. The verification team thinks that it is important to remind verified project owners that they need to add an update so that they dont lose their verified status.
> 
> Two notifications to be added to the notification center - these would be sent to verified projects
> 
> 45 days without an update - friendly reminder: "Don't forget to add regular updates to your project to keep your verified status! - Learn more"
> 
> 90 days without an update - "You have 14 days to add an update to avoid badge expiration - Learn more"
> 
> Something like that.


@WhyldWanderer 
[...]

PR #1396, opened 2024-03-10
change project verificationStatus to null when a new update is added

RamRamez:
https://github.com/Giveth/impact-graph/issues/932#issuecomment-1983556567


mohammadranjbarz (changes requested):
Thanks @RamRamez  Codes seems good but would appreciate if you add at least one test case

RamRamez:
> Thanks @RamRamez Codes seems good but would appreciate if you add at least one test case

Done!

mohammadranjbarz (approved):
Thanks @RamRamez  great job.

PR #1397, opened 2024-03-11
remove Ortto tag only if user has no other projects in the round

RamRamez:


mohammadranjbarz (approved):
Thanks @RamRamez LGTM

PR #1410, opened 2024-03-13
Fix actual matching view and add test cases for it

mohammadranjbarz:
related to https://github.com/Giveth/giveth-dapps-v2/issues/3840

Test cases added for these scenarios:
* Ensures the view is not null for projects with no donations
* Verifies accurate calculations for a project with a single donation
* Confirms donations from recipients of verified projects are excluded
* Validates correct aggregation of multiple donations to a project
* Ensures accurate calculation when a single user makes multiple donations
* Confirms donations under $1 are correctly ignored in calculations
* Verifies aggregated donations from a user exceeding $1 are included
* Asserts that donations to non-verified projects are properly included
* Checks that donations to unlisted projects are included in the analysis
* Ensures donations from identified Sybil users are excluded
* Validates that donations from users with insufficient passport scores are excluded
* Validates that donations in non-eligible networks are excluded from both pre-analysis and post-analysis
* Confirms that donations to flagged fraud projects are not considered
* Ensures pending and failed donation statuses are excluded from both pre-analysis and post-analysis totals

### DIFF
--- a/migration/1710322367912-project_actual_matching_v11_.ts
+++ b/migration/1710322367912-project_actual_matching_v11_.ts
@@ -1,0 +1,169 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ProjectActualMatchingV11_1710322367912
+  implements MigrationInterface
+{
+  async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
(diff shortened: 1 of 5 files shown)


mohammadranjbarz:
@CarlosQ96  Feel free to merge it if you approved that because I think @laurenluz  needs it soon

CarlosQ96 (approved):
Great test cases! looks good!

Issue #1176, opened 2023-10-30
Two projects with same slug and recipient address


WhyldWanderer:
These two projects have the same slug and recipient address. 
Shouldn't this be impossible?

https://giveth.io/project/women-economic-empowerment

Project IDs are 2881 and 2882

Maybe we can delete the one that is not verified?

jainkrati:
@mohammadranjbarz pls guide @RamRamez to take this up. Just trying to onboard him to solve these kind of product heavy issues

RamRamez:
@WhyldWanderer I checked and we have only one project with the slug "women-economic-empowerment" that is id = 2881.

WhyldWanderer:
Yes.. I believe the 2882 one got deleted in the end. However it doesn't solve the issue as Giveth shouldn't allow for this use case. 

Maybe it's still worth doing some testing to create some projects with the same title to see what happens on the backend?

RamRamez:
@WhyldWanderer of course, I did some testing, it's impossible to add duplicate projects with UI, but by sending multiple direct requests to backend at the same time, I could create different projects with same title and wallet address.
To solve this, I added a unique index for slug field in Project entity. Also, we had some projects with duplicate slug on staging and production that I solved them all.
I've sent the PR and it's waiting for review.

RamRamez:
The PR for this issue is ready. I'll merge it on staging after SuperFluid release.

RamRamez:
@maryjaf to test and verify

maryjaf:
It has been resolved and know there is no way to create one project with same name and address

PR #1418, opened 2024-03-19
Two projects with same slug and recipient address

RamRamez:
#1176


mohammadranjbarz:
@RamRamez  two test cases have been failed

RamRamez:
> @RamRamez two test cases have been failed

Fixed!

mohammadranjbarz (changes requested):
Thanks @RamRamez  LGTM.
But you need to add a migration file to add the unique index in the migration (because the annotations just will work on test environment, for local or production ENV you should write migration file

RamRamez:
> Thanks @RamRamez LGTM. But you need to add a migration file to add the unique index in the migration (because the annotations just will work on test environment, for local or production ENV you should write migration file

Done!

RamRamez:
> Thanks @RamRamez LGTM. But you need to add a migration file to add the unique index in the migration (because the annotations just will work on test environment, for local or production ENV you should write migration file

Done!

CarlosQ96 (approved):
LGTM but before prod merging we should check there arent duplicates and delete/modify them or else the deploy will fail in the migration

RamRamez:
> LGTM but before prod merging we should check there arent duplicates and delete/modify them or else the deploy will fail in the migration

Yes, I've already changed the slug of duplicate projects in both staging and production

mohammadranjbarz (approved):
Thanks @RamRamez nice job.

Issue #1407, opened 2024-03-12
P2 | Queries to fetch stats about new donors


divine-comedian:
According to our planned OKRs for Dapp WG we want to track some information about new donors for our marketing & comms efforts. 

We would like to be able to query two things: 

- [x] The number of new donors that have arrived to the platform between two dates (users who have made their first donations).
- [x] The volume of donations made in USD from new donors between two dates (the total USD value of the first donations made by users) 

This we will need to run at least once every month... If we think it's worth the work we can add the queries into the stats.giveth.io page so the devs don't need to be responsible for doing this all the time. 

@jainkrati - for your info - I would recommend @RamRamez for this one!

jainkrati:
@RamRamez pls work on this once p0 issues are solved. @mohammadranjbarz pls guide him only if he is blocked.

RamRamez:
@maryjaf to test and verify

RamRamez:
@maryjaf You can test it with[ this link](https://analytics-dashboard-git-add-new-donors-count-givethio.vercel.app/) on staging and after SuperFluid release, you can test it on production with the main URL (https://stats.giveth.io/)

divine-comedian:
In order to test this issue: 

1. make a new user on Giveth (or use an existing profile that has never made any donation before) 
2. make a donation on giveth staging with a token with a tracked price (ETH is always good)
3. check results on stats dashboard https://analytics-dashboard-git-add-new-donors-count-givethio.vercel.app/
4. results should be reflected correctly - new user count should increase by 1, and New Donors Donation Total USD
5. value of first donation should be reflect as well

If the test goes well make a 2nd new user and repeat steps 2-5
  
 ### What should not happen
- the second donation made by the new user should not be reflected in the count
- donations made by users who have already made donations before should not affect the count nor USD value calculations

maryjaf:
-when I donate with a new user without any donation these fields be increased as expected 
Donations Count:will increase
Donors Count:will increase

-when I donate again with this user Donors Count 
Donors Count: remains unchanged
Donations Count : will increase

-when I donate with a new user but anonymous 
Donors Count: remains unchanged
Donations Count : will increase

and for these new fields, by donating first time with new user these fileds be increased and for second time remains unchanged

New Donors Count
New Donors Donation Total USD

maryjaf:
Should recurring donation be considered in this calculation? or it needs a separate issue ? @RamRamez @divine-comedian

divine-comedian:
[...]

PR #1422, opened 2024-03-22
add Queries to fetch stats about new donors

RamRamez:
#1407


CarlosQ96 (approved):
Queries look good, thanks for making tests.

Issue #1316, opened 2024-02-13
Remove RegisterResolver and ConfirmUserResolver resolvers and their dependencies


aminlatifi:
As the impact-graph project was originally designed for another purpose and then refactored for Giveth by the initial dev team, some code leftovers are not aligned with the current application.
One important functionality (which I am not sure is working yet) is allowing users to register with a password, as is traditional in web2.
So, if we all agree that this feature is not used, we must remove it asap.

aminlatifi:
@jainkrati @mohammadranjbarz @CarlosQ96 @MoeNick

mohammadranjbarz:
> @jainkrati @mohammadranjbarz @CarlosQ96 @MoeNick

I totally agree, we have lots of unused codes that it's better to be deleted

CarlosQ96:
Yeah I agree, if those endpoints are not used by frontend, we can remove them. The impact graph grew a lot.

jainkrati:
@MoeNick pls confirm if we should remove the feature "allow user to register with password"

MoeNick:
Yeah its deprecated. We can go ahead to remove. Im not sure its P1, I'd put p2, unless if it has a tech priority.

aminlatifi:
Removing those also allows us to delete some dependencies that are so dated and have vulnerabilities. One is `nodemailer` package

RamRamez:
I wonder why "REGISTER_USERNAME_PASSWORD" is true in config file, because we are not using it in FE.
![image](https://github.com/Giveth/impact-graph/assets/73441857/a14999fc-8321-4d6f-a236-f00c932a9b5e)

RamRamez:
The PR for this issue is ready. I'll merge it on staging after SuperFluid release.

PR #1429, opened 2024-03-25
Remove registerResolver and confirmUserResolver resolvers

RamRamez:
#1316


mohammadranjbarz (approved):
Thanks @RamRamez, removing these zombie codes are really necessary.

Issue #1416, opened 2024-03-18
Send Emails for Superfluid Notifications 


divine-comedian:
Related to https://github.com/Giveth/notification-center/issues/62

We will add 4 new notifications to the notification center, we will need to trigger some emails when these events occur (depending on user notification settings)

- 1 month stream balance left warning 
- 1 week stream balance left warning 
- user runs out of stream balance
- project owner receives new recurring donation to their project (only if monthly USD value is greater than $20)
   - The project will only receive this email ONCE - when the recurring donation is created - we are able to know how much the user proposed to donate monthly, we can find the tokens and get the USD value. We should NOT send this every month, but only upon the first creation of the recurring donation.

## SuperFluid Balance Warning
We'll use one new activity to capture the first 3 events `act:cm:superfluid-balance-warning` with these variables: 
- `"bol:cm:isended"`: {( true/false - if stream has status of 'ended')}
- `"str:cm:criticaldate"` {1 month || 1 week || null}:
- `"str:cm:email"`: {emailAddress},
- `"str:cm:tokensymbol"`: {underlyingTokenSymbol}
- `"str:cm:userid"`: {userId}

for the balance warnings we will send the same email but replace the text in criticalDate. if the activity is sent with isEnded = true we will send a slightly different email.

## Recurring Donation Received
We should use the existing activity `act:cm:testing-donation-received` and append a new field to it that we send with each activity call

- `"bol:cm:isrecurringdonation"`: {true || false}

and modify existing fields we send: 

- `"str:cm:donationamount"`: {amountDonatedMonthly}
- `"str:cm:donationtoken"`: {underlyingTokenSymbol}

In our existing donation received Journey we will check the value of isRecurringDonation and send a slightly different email if the value = true

---

@RamRamez - @CarlosQ96 has prepared these notifications and you should sync with him on where they need to be triggered from.

RamRamez:
The related PR is not merged yet and I can't start this issue, @CarlosQ96 please ping me when it's ready

krmet:
@divine-comedian have these activities been added to Ortto?

divine-comedian:
> @divine-comedian have these activities been added to Ortto?

Yep! These activites exist already - you can find them from the activities page in Ortto..

RamRamez:
@maryjaf you can test "Recurring Donation Received" (for donations more than 20$ monthly).

[...]

PR #1443, opened 2024-03-31
add donation received email service for recurring donations

RamRamez:
#1416


mohammadranjbarz (approved):
Thanks @RamRamez  Good job.

Issue #1453, opened 2024-04-04
Bug in showing recurring donations in public view of user page 


maryjaf:
My user has 4 recurring donations 
3 active and 1 archive 
I see below pic in my recurring donation

<img width="1338" alt="image" src="https://github.com/Giveth/impact-graph/assets/111529185/c36e8bfc-961d-4173-807b-fbc54b5a127f">

but in **user page (public view)** only 2 recurring donations are shown 

https://staging.giveth.io/user/0xa1179f64638adb613ddaac32d918eb6beb824104?tab=recurring-donations
<img width="1357" alt="image" src="https://github.com/Giveth/impact-graph/assets/111529185/0f75ffc4-2d57-4a07-b4a4-c14d2d66ff3d">

I've checked this with cc: @MohammadPCh and he said this should be fixed by backend 
cc: @jainkrati @mohammadranjbarz 

expected result:
in **owner view** should be shown all of recurring donations (active, ended, archive, failed, pending)
in **public view ** should be shown all Active recurring donations 
@divine-comedian @MoeNick

MoeNick:
In Public View | profile Page, The backend should filter all recurring donations based on 
1- status: Active
2- Owner

@mohammadranjbarz can you tale a look?

mohammadranjbarz:
> My user has 4 recurring donations 3 active and 1 archive I see below pic in my recurring donation
> 
[...]

PR #1458, opened 2024-04-07
Show user's totalDonated and donationsCount correctly even with recurring donations

mohammadranjbarz:
#1453 , https://github.com/Giveth/giveth-dapps-v2/issues/3886


RamRamez (approved):
Great! Thanks @mohammadranjbarz

PR #1464, opened 2024-04-09
fix accountBalance issues

RamRamez:
#1416


mohammadranjbarz (approved):
Thanks @RamRamez  LGTM

PR #1465, opened 2024-04-09
fix super tokens symbol problem

RamRamez:
#1416

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -35,6 +35,93 @@ export const superTokensToToken = {
   GIVx: 'GIV',
 };
 
+export const superTokens = [
+  {
+    underlyingToken: {
+      decimals: 18,
+      id: '0x2f2c819210191750F2E11F7CfC5664a0eB4fd5e6',
+      name: 'Giveth',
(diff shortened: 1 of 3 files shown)


src/provider.ts
```diff
@@ -35,6 +35,93 @@ export const superTokensToToken = {
...
+      name: 'USD Coin',
+      symbol: 'USDC',
+    },
+    decimals: 18,
+    id: '0x8430f084b939208e2eded1584889c9a66b90562f',
+    name: 'Super USD Coin',
+    symbol: 'USDCx',
+  },
+];
+
```
mohammadranjbarz, line 124: @RamRamez Can you write a migration for adding them into DB?

mohammadranjbarz (changes requested):
All things seems good, just a missing part is writing migration

src/provider.ts
mohammadranjbarz: @RamRamez My mistake, it's good to go

mohammadranjbarz (approved):
LGTM

PR #1473, opened 2024-04-11
1459 document admin panel role permissions

Meriem-BM:


mohammadranjbarz (approved):
Thanks @Meriem-BM LGTM

Issue #1471, opened 2024-04-10
Create new admin role for QF stuff


mohammadranjbarz:
The role name can be **qfManager**.

This needs to have these permissions (based on what @laurenluz said in the Discord)
- [x] add project to QF round
- [x] remove project from QF round
- [x] update QF round histories
- [x] list project
- [x] create/edit a new/active QF round
- [x] update all QF round parameters (min USD, matching pool, start/end date)
- [x] add/remove sybils
- [x] add/remove disqualified projects(project fraud)
- [x] export QF round data to spreadsheet

mohammadranjbarz:
@maryjaf  It's on staging you can test it now, Please tag @Meriem-BM  and Me if there is any problem.
Permissions should work as it's written in the https://github.com/Giveth/impact-graph/blob/staging/docs/adminPermissions.md

maryjaf:
@mohammadranjbarz I've created this new user and set "qf manager" role https://impact-graph.serve.giveth.io/admin/resources/User/records/733/show 
but I couldn't login with this  setting email address and passwrod
maryam.jafarymehr@gmail.com
Should something else be set ?

mohammadranjbarz:
> @mohammadranjbarz I've created this new user and set "qf manager" role https://impact-graph.serve.giveth.io/admin/resources/User/records/733/show but I couldn't login with this setting email address and passwrod [maryam.jafarymehr@gmail.com](mailto:maryam.jafarymehr@gmail.com) Should something else be set ?

I just added a qfManager user few minutes ago and it works well, I guess we have more than one user with that email address, in this case it might not work, please make sure of that with an admin user

maryjaf:
- [x] 1- QF manager role han't access to create/ remove a  record on Sybil and Project Fraud page, it has only filter and view access on these 2 pages
![image](https://github.com/Giveth/impact-graph/assets/111529185/4c6809eb-15ed-40b0-9362-ae03dee26c47)
@mohammadranjbarz @Meriem-BM

maryjaf:
> * add project to QF round
> * [x]  remove project from QF round
> *

- [x] 2- There is no access to these 2 actions for QF manager role
 add project to QF round
 remove project from QF round

maryjaf:
- [x] 3-
Project Status reason and Project status history are also shown for qf manager role, and these 2 menu aren't exist in above list,Is there a special reason for showing them or should be removed based on the list? 

[...]

PR #1480, opened 2024-04-15
1471 qf stuff role bugs

Meriem-BM:
Related to #1471


mohammadranjbarz (approved):
Thanks @Meriem-BM LGTM

PR #1482, opened 2024-04-15
add streamBalanceWarning to user

RamRamez:
[#62](https://github.com/Giveth/notification-center/issues/62#issuecomment-2054164024)


mohammadranjbarz (changes requested):
@RamRamez  What happens if a donor have recurring donations for ETH and GIV, one of them is running out but the other token is ok?
I think you should have different values for each token, for instance you can change `streamBalanceWarning` instead of string to something like:
streamBalanceWarning:{
'Eth':...,
'GIV':...
}

WDYT?

RamRamez:
> @RamRamez What happens if a donor have recurring donations for ETH and GIV, one of them is running out but the other token is ok? I think you should have different values for each token, for instance you can change `streamBalanceWarning` instead of string to something like: streamBalanceWarning:{ 'Eth':..., 'GIV':... }
> 
> WDYT?

Done!

mohammadranjbarz (approved):
Thanks @RamRamez LGTM

PR #1485, opened 2024-04-16
Add qfround title and description

CarlosQ96:
Related to: https://github.com/Giveth/giveth-dapps-v2/issues/3675


CarlosQ96:
I think this should be enough. As the title and description is now fetch-able from any project in the allProjects Query.

My suggestions is  if its null, use the previous implementation, else the data in the fields @mohammadranjbarz  @RamRamez

jainkrati (approved):

jainkrati (approved):

PR #1493, opened 2024-04-18
Fix: The Project Status History Tab should not be displayed for users with the QF Manger role.

Meriem-BM:
Related to https://github.com/Giveth/impact-graph/issues/1471#issuecomment-2061624418


mohammadranjbarz (approved):
Thanks @Meriem-BM LGTM.

PR #1504, opened 2024-04-24
fix recurring donation count

RamRamez:
[#3093](https://github.com/Giveth/giveth-dapps-v2/issues/3093)


CarlosQ96 (approved):
Thanks LGTM!

Issue #1462, opened 2024-04-09
Webservice for archived qf rounds


mohammadranjbarz:
We require this information to be fetched from the backend. Could you please provide it?

We need to retrieve all ended (archived) QF rounds with the following details:

1. Slug
2. Name
3. Matching pool
4. Total donations amount
5. Unique donors
6. Start date
7. End date

Additionally, we need to sort them by the following criteria:

- Matching pool
- Number of unique donors
- Newest
- Oldest

Thank you.
<img width="1214" alt="Screenshot 2024-04-09 at 1 58 29 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/17097576/e10c50cd-7688-4192-96b6-640341d37d9d">
cc: @laurenluz @jainkrati

_Originally posted by @MohammadPCh in https://github.com/Giveth/giveth-dapps-v2/issues/3748#issuecomment-2044676561_

mohammadranjbarz:
@CarlosQ96 Can you pick this?

RamRamez:
> @CarlosQ96 Can you pick this?

I'm on it

maryjaf:
Could you please add more data on this issue how can I test it ? @RamRamez

RamRamez:
@maryjaf I think it's better to test it with the FE issue at the same time
https://github.com/Giveth/giveth-dapps-v2/issues/3748

PR #1509, opened 2024-04-27
Archived qf round summary page & link on the main qf page #3748 

RamRamez:
#1462 https://github.com/Giveth/giveth-dapps-v2/issues/3748

### DIFF
--- a/src/repositories/qfRoundRepository.ts
+++ b/src/repositories/qfRoundRepository.ts
@@ -1,5 +1,6 @@
 import { QfRound } from '../entities/qfRound';
 import { AppDataSource } from '../orm';
+import { QfArchivedRoundsOrderBy } from '../resolvers/qfRoundResolver';
 
 const qfRoundEstimatedMatchingParamsCacheDuration = Number(
   process.env.QF_ROUND_ESTIMATED_MATCHING_CACHE_DURATION || 60000,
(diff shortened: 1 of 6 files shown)


src/repositories/qfRoundRepository.ts
```diff
@@ -11,6 +13,84 @@ export const findAllQfRounds = async (): Promise<QfRound[]> => {
...
+  uniqueDonors = 'uniqueDonors',
+  beginDate = 'beginDate',
+}
+
+registerEnumType(QfArchivedRoundsSortType, {
+  name: 'QfArchivedRoundsSortType',
+  description: 'The attributes by which archived rounds can be sorted.',
+});
+
+@ObjectType()
```
CarlosQ96, line 28: I'd Add some indexes to things like Name, slug, to facilitate adminjs searchs.

The Id column must be of type   @PrimaryGeneratedColumn()
which also sets and index.

RamRamez: QFArchivedRounds is used for defining GraphQL types and not directly as a TypeORM entity, I think we don't need to use @PrimaryGeneratedColumn() or @Index(). These are only relevant if the class is being used as a database entity.

CarlosQ96 (approved):
Thanks Ramin!

PR #1528, opened 2024-05-03
Hotfix staging fix latency

CarlosQ96:
Changes:
1. Convert queries from .query method to .createQueryBuilder. So the slave database is used.
2. Add cache for 1 minute for avoiding extra queries when loading the project.

### DIFF
--- a/src/repositories/donationRepository.ts
+++ b/src/repositories/donationRepository.ts
@@ -3,9 +3,10 @@ import moment from 'moment';
 import { Project } from '../entities/project';
 import { Donation, DONATION_STATUS } from '../entities/donation';
 import { ResourcesTotalPerMonthAndYear } from '../resolvers/donationResolver';
-import { AppDataSource } from '../orm';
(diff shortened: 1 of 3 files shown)


src/repositories/donationRepository.ts
```diff
@@ -481,42 +482,44 @@ export async function sumDonationValueUsdForQfRound(params: {
...
+    `"projectEstimatedMatchingView"`,
+  )
+    .select(`projectEstimatedMatchingView."sumValueUsd"`)
+    .where(`projectEstimatedMatchingView."projectId" = :projectId`, {
+      projectId,
+    })
+    .andWhere(`projectEstimatedMatchingView."qfRoundId" = :qfRoundId`, {
+      qfRoundId,
+    })
+    .cache(`sumDonationValueUsdForQfRound-${projectId}-${qfRoundId}`, 60000)
```
jainkrati: should we keep cache eviction configurable from server?

jainkrati (approved):

Issue #1535, opened 2024-05-06
Too many calls for refresh project_donation_summary_view materialized view


aminlatifi:
There are plenty of such logs generated on the backend. Refreshing the materialized view is a heavy process that must not be done frequently. 
We must revise the policy of updating this materialized view, like debouncing!
```
query failed: |  
-- | --
  |   | error: error: canceling statement due to statement timeout |  
  |   |   |  
  |   | REFRESH MATERIALIZED VIEW project_donation_summary_view |  
  |   | query failed: |  
  |   | error: error: canceling statement due to statement timeout |  
  |   |   |  
  |   | REFRESH MATERIALIZED VIEW project_donation_summary_view |  
  |   | query failed:

```

geleeroyale:
After @Rolazo set a time-limit for queries we can notice which queries are affected. Above query is constantly timing out. We will collect all the timed out queries in issues like this, but this seems to be the **most unoptimized one**

maryjaf:
Does this issue need to be tested on QA side?  @CarlosQ96 @aminlatifi

maryjaf:
Does this issue need to be tested on QA side?  @CarlosQ96 @aminlatifi

Rolazo:
The merge didnt work, after disabling time our, `refresh project_donation_summary_view materialized view` was getting 6 minutes to be executed.

jainkrati:
@CarlosQ96 pls guide @RamRamez for this

RamRamez:
project_donation_summary_view is completely removed.

@maryjaf please test and verify

- [x] sumDonationValueUsd and countUniqueDonors should be shown correctly for each project.
- [ ] after each successful donation, sumDonationValueUsd and countUniqueDonors should be updated correctly

maryjaf:
> * after each successful donation, sumDonationValueUsd and countUniqueDonors should be updated correctly

after new donation the total donation has been updated correctly, but the number of donor isn't changed 

-I've donated to[ this project](https://staging.giveth.io/project/how-many-photos-is-too-many-photos?tab=donations) with [new donor ](https://staging.giveth.io/user/0x0b2d7ccd6bef317749641431d54ec9d76dff7efb?tab=donations)
-I haven't donated with this user before
-I expect that the number of donor be changed from 41 to 42 but it isn't changed

RamRamez:
> > * after each successful donation, sumDonationValueUsd and countUniqueDonors should be updated correctly
> 
[...]

PR #1539, opened 2024-05-07
remove refresh_project_summary_from totals and add user cache

CarlosQ96:
No calculated function from any entity should be without cache.
Removed refreshProjectDonationSummaryView from the updateTotal donations of project. That is a really common function. Let the cronjob take do the update every 10 minutes.

Related to https://github.com/Giveth/impact-graph/issues/1535

### DIFF
--- a/src/repositories/powerBoostingRepository.ts
+++ b/src/repositories/powerBoostingRepository.ts
@@ -165,7 +165,11 @@ export const findPowerBoostingsCountByUserId = async (
     .leftJoin('powerBoosting.user', 'user')
     .addSelect(publicSelectionFields)
     .where(`percentage > 0`)
-    .andWhere(`"userId" =${userId}`);
+    .andWhere(`"userId" =${userId}`)
+    .cache(
+      `findPowerBoostingsCountByUserId-recurring-${userId}`,
+      Number(process.env.USER_STATS_CACHE_TIME || 60000),
+    );
   return query.getCount();
 };
 
--- a/src/services/donationService.ts
+++ b/src/services/donationService.ts
@@ -198,9 +198,6 @@ export const updateTotalDonationsOfProject = async (
     `,
       [projectId],
     );
-
-    // we want to update the project donation summary view after updating the total donations
-    refreshProjectDonationSummaryView();
   } catch (e) {
     logger.error('updateTotalDonationsOfAProject error', e);
   }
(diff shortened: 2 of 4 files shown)


src/services/donationService.ts
```diff
@@ -198,9 +198,6 @@ export const updateTotalDonationsOfProject = async (
     `,
       [projectId],
     );
```
CarlosQ96, line 200: this is a pretty common function, called even inside the parent functions like syncing donations with blockchain. So it was calling refreshProjectDonationSummaryView twice at least. 
And on every donation. It's best to remove from here and let the cronjob and other functions update the data

jainkrati (approved):

src/repositories/powerBoostingRepository.ts
```diff
@@ -165,7 +165,11 @@ export const findPowerBoostingsCountByUserId = async (
     .leftJoin('powerBoosting.user', 'user')
     .addSelect(publicSelectionFields)
     .where(`percentage > 0`)
-    .andWhere(`"userId" =${userId}`);
+    .andWhere(`"userId" =${userId}`)
+    .cache(
```
aminlatifi, line 169: Are you sure it doesn't impact user experience? I wonder whether UI relies on this value for rendering modals or components, @MohammadPCh can you confirm or introduce a FE dev who is involved with this?
If it's called a lot and calculated by each user fetch, we may have to find another way around like removing it as a field and making a separate endpoint/resolver to fetch it.

aminlatifi (changes requested):

src/repositories/powerBoostingRepository.ts
CarlosQ96: I agree we must either precalculate and save the data.
Or add cache.

We can't have queries executed everytime on the user on any minimal call.
Because basically 1 query to the user does 4 more under the hood @aminlatifi. So the complexity keeps growing.

CarlosQ96: It was the exact same issue with the project queries that I added cache.

CarlosQ96: @MohammadPCh @RamRamez can double check if those user values are required to be updated every second. I added a 1 minute cache.

Rolazo:
Even with the cache, the issue still persist

Issue #1556, opened 2024-05-11
Investigate AND verify stuck & missing donations


divine-comedian:
On thursday/friday a user reported their donations were not showing up on ETH mainnet to this project 

https://giveth.io/project/emergency-relief-fund-for-brazil-floods

I was able to confirm this by making a donation myself in ETH to their project on mainnet, I cannot find the donation in 'my donations' or on the project table, nor in the back-up service. 

Here are the three donations we know of that we're not caught anywhere: 
https://etherscan.io/tx/0x331ff224c7e4dcc3ce9bb6aa6b3a67e1d7b1c7f1c27e6fdead150fd09a6499b6
https://etherscan.io/tx/0xf1a4fb4c446c25529fed988c83381db41fd6cbbe8aae69a58f2a93a60a8b9f73
https://etherscan.io/tx/0x0373f7efaeb12c205e2c26f2b14a49194b9e619c6728c8a5a4f8cf36f635ef6f

Alternatively, when I was testing donations to other projects on ETH mainnet my donations show up in my donations but they are stuck in a pending state, I also cannot find them in the back-up service. I made some donations on Optimism mainnet and encountered no issues for comarison.

![image](https://github.com/Giveth/impact-graph/assets/67759413/def66460-8aaf-4ff5-82f5-eb78669994b8)

Here are the txs for my stuck pending donations:
https://etherscan.io/tx/0xc81021a481b1a21947b1aed7199cd7febdd662cd9a4ee839eb734849b59b24c2
https://etherscan.io/tx/0x5c4cab8d517393178a6f59af693185848c57b21975710cb71f706e8984b117db

--- 

I tested today, late friday making a donation on eth mainnet this donation was successful and show on my donations and the projects donations table. however the previous donations on eth mainnet do not show. 

How did this happen? We should 1st add the missing donations into our DB. 2nd we need to investigate how this happens that donations are stuck as pending and how it can occuer that other donations are just not caught by any service at all, but only for eth mainnet

divine-comedian:
Also, there is 61 failed donations in the back-up service that needed to be checked they have been added back to our UI! This is from last week when the donation verification service went down unexpectedly.

![image](https://github.com/Giveth/impact-graph/assets/67759413/7fb2dc35-27ff-4e2c-a404-aa913ae62f02)

mohammadranjbarz:
@divine-comedian 
[...]

PR #1563, opened 2024-05-13
Surround all refresh materialized views in try..catch.. scope to make sure the failure of that doesnt have any side effect on our application

mohammadranjbarz:
related to #1556


CarlosQ96 (approved):
LGTM!

PR #1573, opened 2024-05-16
Ignore small differences of amounts when matching draft donation for …

mohammadranjbarz:
@jainkrati @divine-comedian  @CarlosQ96 
The reason the the last tx in https://github.com/Giveth/impact-graph/issues/1556#issuecomment-2108706229 didn't match is because we rounded the number in our db to `40.5555` but the real one is `40.555499` https://gnosisscan.io/tx/0xfa65ef0a52e2f3b96c5802dcee4783858511989b7235035e8cab4d527fa15a1a
So comparing tx data for erc20 token transfer fails, we need to save `amoun` as bignumber in draftDonationResovler to prevent this error but as it needs both frontend and backend changes and also would not solve previous draft donations I just added something to ignore small differences ( less than `0.001`) but we should change it to bigNumber later

I'm not available today anymore so if no one can merge/monitor it on production I can do it on SUN

### DIFF
--- a/src/resolvers/draftDonationResolver.ts
+++ b/src/resolvers/draftDonationResolver.ts
@@ -41,6 +41,7 @@ export class DraftDonationResolver {
 
   @Mutation(_returns => Number)
   async createDraftDonation(
+    // TODO we should change it to bigInt in both backend and frontend to not round numbers
     @Arg('amount') amount: number,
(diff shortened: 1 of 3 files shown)


CarlosQ96 (approved):
LGTM thanks man!

PR #1593, opened 2024-05-28
Search functionality hot-fix

RamRamez:


jainkrati (approved):

jainkrati:
i am curious, why was the previous change done in the first place and now how are we benefitting by this change

RamRamez:
> i am curious, why was the previous change done in the first place and now how are we benefitting by this change



> i am curious, why was the previous change done in the first place and now how are we benefitting by this change

This change is suggested by Amin to add similarity search with pg_trgm extension.

PR #1613, opened 2024-06-07
fix: filter unique donors based on passport score and knownAsSybilAdd…

Meriem-BM:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/4245

Added a condition to count only the unique donors that has 
- passportScore >= minimumPassportScore
- knownAsSybilAddress is false

cc: @CarlosQ96 

## Summary by CodeRabbit

- **New Features**
  - Added a new field `isDataAnalysisDone` to the QF rounds, providing more detailed tracking of data analysis completion.

- **Enhancements**
  - Improved the calculation for `uniqueDonors` by including additional conditions based on user attributes.

- **Style**
  - Updated styling properties for components in the QF Round Sponsors Images section for better visual consistency.

### DIFF
--- a/src/repositories/qfRoundRepository.ts
+++ b/src/repositories/qfRoundRepository.ts
@@ -108,7 +108,6 @@ export const findArchivedQfRounds = async (
   const fullRounds = await QfRound.createQueryBuilder('qfRound')
     .where('"isActive" = false')
     .leftJoin('qfRound.donations', 'donation')
-    .leftJoin('donation.user', 'user')
     .select('qfRound.id', 'id')
     .addSelect('qfRound.name', 'name')
     .addSelect('qfRound.slug', 'slug')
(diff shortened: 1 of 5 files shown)


src/repositories/qfRoundRepository.ts
```diff
@@ -105,14 +108,19 @@ export const findArchivedQfRounds = async (
...
     .addSelect('qfRound.isActive', 'isActive')
     .addSelect('qfRound.endDate', 'endDate')
     .addSelect('qfRound.eligibleNetworks', 'eligibleNetworks')
+    .addSelect('qfRound.isDataAnalysisDone', 'isDataAnalysisDone')
     .addSelect('SUM(donation.amount)', 'totalDonations')
-    .addSelect('COUNT(DISTINCT donation.fromWalletAddress)', 'uniqueDonors')
+    .addSelect(
+      'COUNT(DISTINCT CASE WHEN user.passportScore >= qfRound.minimumPassportScore AND user.knownAsSybilAddress = FALSE THEN donation.fromWalletAddress ELSE NULL END)',
+      'uniqueDonors',
+    )
```
CarlosQ96, line 129: This can be ineficient, because during each count it will execute this instruction instead of for example a WHERE filter that has a smaller subset of data. 
A faster approach would be to do a subquery that calculates this. For example:
```
.addSelect(
    qb => qb
      .select('COUNT(DISTINCT donation.fromWalletAddress)')
      .from('donation', 'subDonation')
      .leftJoin('subDonation.user', 'subUser')
      .where('subDonation.qfRoundId = qfRound.id')
      .andWhere('subUser.passportScore >= qfRound.minimumPassportScore')
      .andWhere('subUser.knownAsSybilAddress = FALSE'),
    'validUniqueDonors'
  )
```

CarlosQ96 (changes requested):
Just this change for performance and then can merge.

src/repositories/qfRoundRepository.ts
CarlosQ96: Try a similar approach and let me know how it goes

Meriem-BM: Thanks @CarlosQ96 for pointing that out.

PR #1624, opened 2024-06-12
fix: retrieve correct USD value of donations for an archived QF round

Meriem-BM:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/4245#issuecomment-2162362533

## Summary by CodeRabbit

- **Refactor**
  - Improved queries for archived rounds to include more detailed information on donations and users.


CarlosQ96 (approved):
LGTM!

CarlosQ96:
This query however can be heavy if the round is HUGE, I would recommend we pre-save the total at the end of the round in its own column and then just display it here.

Meriem-BM:
> This query however can be heavy if the round is HUGE, I would recommend we pre-save the total at the end of the round in its own column and then just display it here.

I agree, for now I'll merge it, but I'll open another PR to improve it with what you have suggested.

PR #1682, opened 2024-07-15
Fix usdc decimals for recurring donations

mohammadranjbarz:
related to https://github.com/Giveth/giveth-dapps-v2/issues/4196

### DIFF
--- a/src/services/recurringDonationService.ts
+++ b/src/services/recurringDonationService.ts
@@ -179,7 +179,8 @@ export const createRelatedDonationsToStream = async (
       const donation = Donation.create({
         amount: normalizeNegativeAmount(
           streamPeriod.amount,
-          tokenInDb!.decimals,
+          // All Super fluid tokens have 18 decimals even USDSx https://optimistic.etherscan.io/token/0x8430f084b939208e2eded1584889c9a66b90562f
+          18,
         ),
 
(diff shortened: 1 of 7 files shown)


src/services/recurringDonationService.ts
```diff
@@ -179,7 +179,8 @@ export const createRelatedDonationsToStream = async (
       const donation = Donation.create({
         amount: normalizeNegativeAmount(
           streamPeriod.amount,
-          tokenInDb!.decimals,
+          // All Super fluid tokens have 18 decimals even USDSx https://optimistic.etherscan.io/token/0x8430f084b939208e2eded1584889c9a66b90562f
```
jainkrati, line 182: just curious: was the previous code tokenInDb!.decimals not giving 18 as the value ?

jainkrati (approved):

src/services/recurringDonationService.ts
mohammadranjbarz: @jainkrati The decimals of `USDC` is `6` https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

But for the superfluid tokens all decimals are `18` so we should not use the decimals that we have in our DB for the tokens

PR #1683, opened 2024-07-15
fix USD value not showing

RamRamez:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new action to fill prices for donations without a price in the admin panel.
	- Added functionality to update donation prices and related database entities directly.

- **Refactor**
	- Simplified and streamlined the permissions and actions related to donation price management.

- **Chores**
	- Removed outdated logic and functions related to filling prices for donations in the latest round.

### DIFF
--- a/src/server/adminJs/tabs/donationTab.ts
+++ b/src/server/adminJs/tabs/donationTab.ts
@@ -272,6 +272,7 @@ export const buildDonationsQuery = (
 export const FillPricesForDonationsWithoutPrice = async () => {
   const donationsWithoutPrice = await Donation.createQueryBuilder('donation')
     .where('donation.valueUsd IS NULL')
+    .andWhere('donation.status = :status', { status: DONATION_STATUS.VERIFIED })
     .orderBy({ 'donation.createdAt': 'DESC' })
     .getMany();
   const coingeckoAdapter = new CoingeckoPriceAdapter();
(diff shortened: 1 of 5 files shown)


src/server/adminJs/tabs/donationTab.ts
```diff
@@ -265,6 +269,55 @@ export const buildDonationsQuery = (
   return query;
 };
 
+export const FillPricesForDonationsWithoutPrice = async () => {
```
mohammadranjbarz, line 272: @RamRamez  Can you add a test case for this function?
Create a donation and ,ake the valueUsd of donation as null
Then check if the price has filled or not

mohammadranjbarz (approved):
Thanks @RamRamez LGTM
But If you add a test case for that would be great

src/server/adminJs/tabs/donationTab.ts
```diff
@@ -265,6 +269,55 @@ export const buildDonationsQuery = (
...
+          message: `${donationsWithoutPrice.length} donations without price found. ${filledPrices.length} prices filled`,
+          type: 'success',
+        },
+      };
+    }
+    donation.valueUsd = donation.amount * price;
+    donation.priceUsd = price;
+    filledPrices.push({ donationId: donation.id, price });
+    await donation.save();
+  }
```
CarlosQ96, line 310: I think we are missing to update the projects totals, but that might be a little heavy. What do you think? @RamRamez

CarlosQ96 (approved):
LGTM Ramin! As mohammad said, would be good to add tests to the function. And also update the project totals, but that might be heavy on the admin js.

src/server/adminJs/tabs/donationTab.ts
RamRamez: You're right. I added a test case.
Thanks @mohammadranjbarz

src/server/adminJs/tabs/donationTab.ts
RamRamez: Yeah, great point, thanks @CarlosQ96.
I added to update project totals, user total received and user total donated.

PR #1698, opened 2024-07-18
Optimise relevant donation flow

ae2079:
Add migration to add and fill donation percentage field in donation table.
Use this field to calculate donation metrics faster

## Summary by CodeRabbit

- **New Features**
  - Introduced a new field to track the percentage of donations.
  
- **Refactor**
  - Renamed function to improve clarity in retrieving donations linked to a specific project ID.

- **Tests**
  - Updated test cases to incorporate the new donation percentage feature and reflect function renaming.

- **Chores**
  - Simplified the test setup by using a hardcoded project ID for improved efficiency.


ae2079:
related to: https://github.com/Giveth/analytics-dashboard/issues/15

src/repositories/donationRepository.ts
```diff
@@ -537,36 +537,17 @@ export async function isVerifiedDonationExistsInQfRound(params: {
...
-      } else {
-        throw new Error(
-          `the relevant donation to this donation does not exist: donation id = ${donation.id}`,
-        );
-      }
-    }
-  });
-
-  return { donationsToGiveth, pairedDonations };
+  return donations.filter(donation => donation.projectId === givethProjectId);
```
RamRamez: @ae2079 I think you can add this condition `donation.projectId === givethProjectId` to your query.
Then it will be much faster.

src/repositories/donationRepository.ts
```diff
@@ -537,36 +537,17 @@ export async function isVerifiedDonationExistsInQfRound(params: {
   }
 }
 
-export async function findRelevantDonations(
+export async function findDonationsToGiveth(
```
RamRamez: I think you can rename this to `findDonationsByProjectId`
because you getting `givethProjectId` as input and actually it could be any project id not only Giveth project

src/services/donationService.ts
```diff
@@ -529,7 +529,7 @@ export async function getDonationToGivethWithDonationBoxMetrics(
     throw new Error('giveth project not found!');
   }
 
-  const { donationsToGiveth, pairedDonations } = await findRelevantDonations(
+  const donationsToGiveth = await findDonationsToGiveth(
     startDate,
     endDate,
     givethProject.id,
```
RamRamez: `const givethProject = await findProjectIdBySlug('giveth');`
This line of code only fetches Giveth project in staging. In production the slug is different, also the slug maybe changed in future so it's better to fetch projects from DB by ID (not slug).
Giveth project ID is `1` in both staging and prod, and you don't need to fetch that from DB. You can consider that as a constant.

RamRamez (changes requested):
Great work, thanks @ae2079 
Just some minor changes needed.

src/repositories/donationRepository.ts
ae2079: I named it this because we have a condition that donation should have `useDonationBox = true`, so I changed it to `findDonationsByProjectIdWitchUseDonationBox` which is a better name

src/repositories/donationRepository.ts
ae2079: Done

src/services/donationService.ts
ae2079: If we know the given project ID is always 1, we can do this.

ae2079: Done

src/repositories/donationRepository.ts
RamRamez: Thanks @ae2079. There's a typo in `findDonationsByProjectIdWitchUseDonationBox`, please fix (Witch -> which)

src/repositories/donationRepository.ts
```diff
@@ -537,36 +537,16 @@ export async function isVerifiedDonationExistsInQfRound(params: {
   }
 }
 
-export async function findRelevantDonations(
+export async function findDonationsByProjectIdWitchUseDonationBox(
   startDate: Date,
   endDate: Date,
   givethProjectId: number,
```
RamRamez: You can rename `givethProjectId` to `projectId`

ae2079: Done

RamRamez (approved):
Great! Thanks @ae2079

src/repositories/donationRepository.ts
ae2079: Done

PR #1729, opened 2024-07-25
improve query execution time

ae2079:
Use raw select query instead of find function from typeorm

## Summary by CodeRabbit

- **Performance Improvements**
	- Enhanced data retrieval for donations using a raw SQL query, potentially improving performance.
  
- **Refactor**
	- Updated the method for finding donations linked to specific projects, shifting from ORM-based to raw SQL execution.


RamRamez (approved):
Thanks @ae2079

PR #1730, opened 2024-07-25
Release query execution optimization 

ae2079:
Use raw select query instead of find function from typeorm


RamRamez (approved):
Thanks @ae2079

PR #1733, opened 2024-07-29
Fix/dual scoring model issues

Meriem-BM:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/4247#issuecomment-2241573083


RamRamez (approved):
Thanks @Meriem-BM

PR #1754, opened 2024-08-07
feat: ajusts quary to make it the same as allProjects query

Reshzera:
## Summary by CodeRabbit

- **New Features**
	- Enhanced project filtering to include only active and listed projects, improving result accuracy.
- **Bug Fixes**
	- Improved security of project queries to prevent SQL injection vulnerabilities.

### DIFF
--- a/src/repositories/projectRepository.ts
+++ b/src/repositories/projectRepository.ts
@@ -437,7 +437,10 @@ export const totalProjectsPerDate = async (
   }
 
   if (onlyListed) {
-    query.andWhere(`project."reviewStatus" = 'Listed'`);
+    query.andWhere(
+      `project.statusId = ${ProjStatus.active} AND project.reviewStatus = :reviewStatus`,
+      { reviewStatus: ReviewStatus.Listed },
+    );
   }
 
   if (networkId) {
--- a/src/repositories/projectRepository.ts
+++ b/src/repositories/projectRepository.ts
@@ -424,6 +424,8 @@ export const totalProjectsPerDate = async (
 ): Promise<number> => {
   const query = Project.createQueryBuilder('project');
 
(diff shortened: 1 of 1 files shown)


Reshzera:
https://github.com/Giveth/analytics-dashboard/issues/34

kkatusic (approved):
LGTM ;), thc @Reshzera

src/repositories/projectRepository.ts
```diff
@@ -437,7 +437,10 @@ export const totalProjectsPerDate = async (
   }
 
   if (onlyListed) {
-    query.andWhere(`project."reviewStatus" = 'Listed'`);
+    query.andWhere(
+      `project.statusId = ${ProjStatus.active} AND project.reviewStatus = :reviewStatus`,
```
RamRamez: I think it's not a good idea to add `statusId` condition inside `onlyListed` condition, because those are separate conditions

RamRamez (changes requested):

src/repositories/projectRepository.ts
kkatusic: @RamRamez he can move it outside, but what if this _statusId_ is only related to listed projects? I just asking, want to know :)

RamRamez: @kkatusic I checked and we have some listed projects with status other than `active`, so we should check them separately. I asked Mitch in the issue for clarification.

kkatusic: thx

Reshzera: @RamRamez  I just fixed the error could you please take a look again

RamRamez (approved):
Thanks @Reshzera

Issue #1760, opened 2024-08-12
update base address of all endaoment projects


divine-comedian:
The Endaoment team has update the receiver contracts used for all of their contracts on BASE network - they should now all be equal to the same address used for the same project on MAINNET. We will need to write a migration to update those addresses on PRODUCTION. 

![Image](https://github.com/user-attachments/assets/a2798d7c-3866-4156-a648-bb2ea5e35102)

So in this example above you can see the base address is the only one that has a different address. Now all of the recipient addresses on all networks should be the same.

divine-comedian:
@mohammadranjbarz says it is easy, should only take a few mins

mohammadranjbarz:
@divine-comedian It's done, you can test it now on Production ENV

divine-comedian:
@LatifatAbdullahi 

to test this one go onto PRODUCTION and review some Endaoment projects listed.

We should check that recipient addresses are the same across all networks for a single project.

LatifatAbdullahi:
@divine-comedian @mohammadranjbarz 

**Test update: Pass**

The recipient address of the Base Network is same as all other networks for a single project.


Jafra Foundation for Relief and Youth Development

![SR](https://github.com/user-attachments/assets/50c3a403-fc79-4da8-841e-7527f6872a7d)



UNRWA USA

![Y](https://github.com/user-attachments/assets/cccc500e-10d0-4be7-8169-428069ac32ed)


Nova Ukraine

![M](https://github.com/user-attachments/assets/0cb6d83f-a94b-4ec6-a308-ef9fe19d8d95)


Maitri India
![S](https://github.com/user-attachments/assets/772f71a5-47c2-4d19-9f37-97265d2254f7)


Reach Out NGO
![R](https://github.com/user-attachments/assets/63371036-ce96-44a2-84c4-2a27bf0d7e99)

PR #1768, opened 2024-08-19
Update base address of endaoment projects

mohammadranjbarz:
Related to #1760


Meriem-BM (approved):
LGTM, thanks @mohammadranjbarz

PR #1806, opened 2024-09-09
fix: add time condition to filter out old transactions

Meriem-BM:


mohammadranjbarz (approved):
Thanks @Meriem-BM LGTM

PR #1871, opened 2024-11-12
update recurring donation token info

divine-comedian:


CarlosQ96:
From the testing that @kkatusic showed me the new token works good. Issue is the old one. The PR looks good to me, if this is the final version.

CarlosQ96 (approved):
LGTM!

PR #2083, opened 2025-07-31
add conditions for proyects to evaluation

CarlosQ96:
issue https://github.com/orgs/Giveth/projects/26/views/2?sliceBy%5Bvalue%5D=CarlosQ96&pane=issue&itemId=115743434&issue=Giveth%7Cimpact-graph%7C2009

## Summary by CodeRabbit

* **New Features**
  * Projects are now filtered based on their token balances on the Polygon network; only projects with a positive token balance are included.
* **Bug Fixes**
  * Improved reliability in project filtering by ensuring only verified projects with sufficient token balances are shown.
* **Tests**
  * Enhanced test reliability by standardizing token balance checks during testing.

### DIFF
--- a/src/services/cronJobs/projectEvaluationService.ts
+++ b/src/services/cronJobs/projectEvaluationService.ts
@@ -1,8 +1,13 @@
 import { schedule } from 'node-cron';
 import axios from 'axios';
+import { ethers } from 'ethers';
 import config from '../../config';
 import { logger } from '../../utils/logger';
 import { Cause } from '../../entities/project';
+import { erc20ABI } from '../../assets/erc20ABI';
+
+const tpolTokenAddress = '0xc7B1807822160a8C5b6c9EaF5C584aAD0972deeC'; // GIV token address for production
+const givTokenAddress = '0xc20CAf8deE81059ec0c8E5971b2AF7347eC131f4'; // TPOL token address for staging
 
 // Cron expression for how often to run the evaluation
 const cronJobTime =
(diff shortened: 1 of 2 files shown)


src/services/cronJobs/projectEvaluationService.ts
```diff
@@ -1,8 +1,13 @@
 import { schedule } from 'node-cron';
 import axios from 'axios';
+import { ethers } from 'ethers';
 import config from '../../config';
 import { logger } from '../../utils/logger';
 import { Cause } from '../../entities/project';
+import { erc20ABI } from '../../assets/erc20ABI';
+
+const tpolTokenAddress = '0xc7B1807822160a8C5b6c9EaF5C584aAD0972deeC'; // GIV token address for production
+const givTokenAddress = '0xc20CAf8deE81059ec0c8E5971b2AF7347eC131f4'; // TPOL token address for staging
```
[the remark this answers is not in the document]
ae2079: @CarlosQ96 please double check the addresses, because one of the comments or the values are swapped

src/services/cronJobs/projectEvaluationService.ts
```diff
@@ -13,6 +18,45 @@ const cronJobTime =
...
+
+// Check balance with caching
+export const checkBalance = async (
+  address: string,
+  provider: ethers.providers.Provider,
+): Promise<string> => {
+  // Check cache first
+  if (balanceCache.has(address)) {
+    return balanceCache.get(address)!;
+  }
```
ae2079, line 39: As we are running distribution more frequently, I don't think cache will be a good solution and we need the real-time balance

src/services/cronJobs/projectEvaluationService.ts
```diff
@@ -67,14 +113,19 @@ export const getActiveCausesWithProjects = async () => {
...
         AND cp."isIncluded" = true
         AND p."statusId" = 5
+        AND p.verified = true
         AND (cat.id IS NULL OR cat."isActive" = true)
       ORDER BY c.id, cp."projectId", cat.id
     `);
 
-    // Group results by cause
+    // Get provider for balance checking
+    const { getProvider } = await import('../../provider');
```
ae2079: Please move the imports to the start of the file

src/services/cronJobs/projectEvaluationService.ts
```diff
@@ -91,9 +142,24 @@ export const getActiveCausesWithProjects = async () => {
...
+      // Add project ID if not already added and has balance > 0
       if (!causeData.projectIds.includes(row.projectId)) {
-        causeData.projectIds.push(row.projectId);
+        // Check balance for the project address
+        try {
+          const balance = await checkBalance(row.project_address, provider);
+          const balanceNumber = parseFloat(balance);
+
+          // Only add project if balance > 0
+          if (balanceNumber > 0) {
```
ae2079: It is better to use a minimum threshold like 0.1 tokens, because based on distribution logics, some times it remains very small value of tokens in the pool addresses

ae2079 (commented):
Thanks @CarlosQ96

src/services/cronJobs/projectEvaluationService.ts
CarlosQ96: thanks

src/services/cronJobs/projectEvaluationService.ts
CarlosQ96: ok!

src/services/cronJobs/projectEvaluationService.ts
CarlosQ96: I think this is mostly for this version, we will do a follow up version. We keep the cache in memory during execution only. In case project 1 is in two causes, it only gets the balance once.
Mitch said as long as the balance is within one hour its fine.

Issue #2071, opened 2025-07-28
add field to project data for ownerTotalEarned


divine-comedian:
we should add two new fields on project data related to Causes - these are 

`ownerTotalEarned` 
`ownerTotalEarnedUsdValue`

these two values represent the total amount and usd value that the cause owner has received from the distributions made by distro service. 

this data will be retrieved on FE for cause owner on "my causes" page

these new fields should be retrievable on at least these two queries: `projectById` and `projectBySlug`

## AC 

- [x] querying `projectById` for existing user on impact-graph there is a field  `ownerTotalEarned` 
- [x] querying `projectById` for existing user on impact-graph there is a field  `ownerTotalEarnedUsdValue`
- [x] querying `projectBySlug` for existing user on impact-graph there is a field  `ownerTotalEarned` 
- [x] querying `projectBySlug` for existing user on impact-graph there is a field  `ownerTotalEarnedUsdValue`

divine-comedian:
cc: @kkatusic

kkatusic:
@divine-comedian I will add update of this values only when some projects gets amount from distribution service, don't wont to run some query that will fetch all projects, their amounts and then update those two fields values. When distribution service provide some amount for the project inside some cause it will update existing values inside project table. If you @ae2079 agree with this. I will sure update existing ones.

divine-comedian:
> [@divine-comedian](https://github.com/divine-comedian) I will add update of this values only when some projects gets amount from distribution service, don't wont to run some query that will fetch all projects, their amounts and then update those two fields values. When distribution service provide some amount for the project inside some cause it will update existing values inside project table. If you [@ae2079](https://github.com/ae2079) agree with this. I will sure update existing ones.

I think the distribution service needs to handle updating those amounts - not the impact-graph directly - same as how Ali implemented updating `totalDistributed` number - the task for actually updating the fields is contained here https://github.com/Giveth/agent-distribution-service/issues/13

maryjaf:
> ## AC
> * [ ]  querying `projectById` for existing user on impact-graph there is a field  `ownerTotalEarned`[ ]  querying `projectById` for existing user on impact-graph there is a field  `ownerTotalEarnedUsdValue`[ ]  querying `projectBySlug` for existing user on impact-graph there is a field  `ownerTotalEarned`[ ]  querying `projectBySlug` for existing user on impact-graph there is a field  `ownerTotalEarnedUsdValue`

these fields are added but I see the value is set 0 


[...]

PR #2084, opened 2025-08-01
Adding ownerTotalEarned and ownerTotalEarnedUsdValue to cause

kkatusic:
- https://github.com/Giveth/impact-graph/issues/2071

## Summary by CodeRabbit

* **New Features**
  * Added new fields to projects showing total earnings received by the project owner, in native and USD values.
  * Updated project detail queries to include owner earnings information.
  * Improved distribution updates to accurately reflect changes in owner earnings.

* **Tests**
  * Added tests to verify owner earnings fields are correctly initialized and saved for projects.


kkatusic:
@ae2079, cc @divine-comedian, when you see this PR please check cause/project table updates inside distribution function, is that ok with you you can merge it, thx

CarlosQ96 (approved):
LGTM

migration/1754042145230-updateOwnerEarningsForCauses.ts
```diff
@@ -0,0 +1,35 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ae2079: We don't need this one, because it is not the correct value for it

src/repositories/causeProjectRepository.ts
```diff
@@ -132,6 +132,13 @@ export const updateCauseProjectDistribution = async (
 
   await causeProject.save();
 
+  // Update project ownerTotalEarned and ownerTotalEarnedUsdValue inside project table
+  await Project.update(projectId, {
+    ownerTotalEarned: (cause.ownerTotalEarned || 0) + amountReceived,
+    ownerTotalEarnedUsdValue:
+      (cause.ownerTotalEarnedUsdValue || 0) + amountReceivedUsdValue,
+  });
```
ae2079: this is not correct as well, we should get the amounts from the update cause distribution data endpoint

ae2079 (changes requested):
Thanks @kkatusic, we need to change the way that we fill the data

Issue #2077, opened 2025-07-29
add causesTotalEarned field to user data


divine-comedian:
we should add two fields called `causesTotalEarned` and `causesTotalEarnedUsdValue` to user data fields, this allows us to track the total amount the user has earned from all causes they own. 

we should be able to get these two new fields off of these queries: 

`userByAddress`

## AC 

- [ ] querying `userByAddress` for existing user on impact-graph there is a field  `causesTotalEarned` 
- [ ] querying `userByAddress` for existing user on impact-graph there is a field  `causesTotalEarnedUsdValue`

divine-comedian:
cc: @kkatusic

maryjaf:
-     "causesTotalEarned": null,
-     "causesTotalEarnedUsdValue": null,

@ae2079 

cURL:

```
curl --location 'https://impact-graph.serve.giveth.io/graphql?=null' \
--header 'accept: */*' \
--header 'accept-language: en,en-US;q=0.9,fa;q=0.8' \
--header 'authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwdWJsaWNBZGRyZXNzIjoiMHhBMTE3OWY2NDYzOGFkYjYxM0REQUFjMzJEOTE4RUI2QkVCODI0MTA0IiwiZXhwaXJhdGlvbkRhdGUiOiIyMDI1LTA4LTMwVDEwOjIyOjQ1LjY3OFoiLCJqdGkiOiIxNzUzOTU3MzY1Njc3LTI0NzVkMzc5ZTQiLCJpYXQiOjE3NTM5NTczNjUsImV4cCI6MTc1NjU0OTM2NX0.a3QKV5GQtCHeVKQtNKqnzpear_p4UpAfxzfIzxZv0L4' \
--header 'authversion: 2' \
--header 'content-type: application/json' \
--header 'origin: https://staging.giveth.io' \
--header 'priority: u=1, i' \
--header 'referer: https://staging.giveth.io/' \
--header 'sec-ch-ua: "Not)A;Brand";v="8", "Chromium";v="138", "Google Chrome";v="138"' \
--header 'sec-ch-ua-mobile: ?0' \
--header 'sec-ch-ua-platform: "macOS"' \
--header 'sec-fetch-dest: empty' \
--header 'sec-fetch-mode: cors' \
--header 'sec-fetch-site: same-site' \
[...]

PR #2085, opened 2025-08-01
Add causes total earned user

kkatusic:
-https://github.com/Giveth/impact-graph/issues/2077

## Summary by CodeRabbit

* **New Features**
  * Added new fields to user profiles to display total earnings from causes and their USD values.
  * Extended user and project queries to return total earned amounts and their USD equivalents.

* **Bug Fixes**
  * Ensured new users have default values set for the new earnings fields.

* **Enhancements**
  * Improved earnings updates to ensure user totals are updated atomically during cause distributions.


kkatusic:
@ae2079 please first this one:

https://github.com/Giveth/impact-graph/pull/2084

same as cause/project update, but for the user, if you need to fix distribution data update be free to do it, thx

migration/1754042145230-updateOwnerEarningsForCauses.ts
```diff
@@ -0,0 +1,35 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ae2079: we don't need this one too, because the owner share is different

migration/1754050166738-populateCauseEarningsToUser.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ae2079: we don't need this one, because the owner share is different

src/repositories/causeProjectRepository.ts
```diff
@@ -132,6 +133,26 @@ export const updateCauseProjectDistribution = async (
...
+  // Update user's causesTotalEarned and causesTotalEarnedUsdValue
+  const user = await User.findOne({
+    where: { id: cause.adminUserId },
+  });
+
+  if (user) {
+    await User.update(cause.adminUserId, {
+      causesTotalEarned: (user.causesTotalEarned || 0) + amountReceived,
+      causesTotalEarnedUsdValue:
+        (user.causesTotalEarnedUsdValue || 0) + amountReceivedUsdValue,
```
ae2079: we should get these data from update cause distribution data endpoint

ae2079 (changes requested):
thanks @kkatusic , we need to update the way that we get the owner shares

kkatusic:
@ae2079 please first this one:

https://github.com/Giveth/impact-graph/pull/2084

same as cause/project update, but for the user, if you need to fix distribution data update be free to do it, thx

PR #2103, opened 2025-08-07
Adding function for title edit validation adn fixing fromTokenAmount …

kkatusic:
## Summary by CodeRabbit

* **New Features**
  * Added validation for editing cause titles, allowing users to check if a new title is valid when updating an existing cause.
* **Bug Fixes**
  * Improved handling of donation creation so that missing token amounts default to zero instead of being left undefined, ensuring more consistent data for both donations and draft donations.


CarlosQ96 (approved):
approved!

PR #2108, opened 2025-08-08
Fetching user profile causes by user id improving fetch

kkatusic:
This solve loading user admin profile cause tab faster

## Summary by CodeRabbit

* **New Features**
  * Added a new query to retrieve a paginated list of causes administered by a specific user, including detailed related data and total count.

### DIFF
--- a/src/resolvers/causeResolver.ts
+++ b/src/resolvers/causeResolver.ts
@@ -741,6 +741,8 @@ export class CauseResolver {
           cp.id as "causeProjectId",
           cp."amountReceived",
           cp."amountReceivedUsdValue",
+          cp."isIncluded",
+          cp."userRemoved",
           p.id AS "projectId",
           p.title,
           p.verified,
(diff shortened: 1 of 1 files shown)


src/resolvers/causeResolver.ts
```diff
@@ -640,4 +666,141 @@ export class CauseResolver {
...
+          isIncluded: row.isIncluded,
+          userRemoved: row.userRemoved,
+          project: {
+            id: row.projectId,
+            title: row.title,
+            verified: row.verified,
+            isGivbackEligible: row.isGivbackEligible,
+            status: row.status,
+            addresses: row.addresses,
+          },
```
ae2079, line 786: I think we are missing some projects here

ae2079 (commented):
Thanks @kkatusic

src/resolvers/causeResolver.ts
kkatusic: thx

ae2079 (approved):
LGTM

Issue #2140, opened 2025-08-22
Update createDonation Mutation to qualify a given donation as part of a particular active QF round


laurenluz:
## Description
Update the donation creation flow so the mutation accepts a **QF round ID (integer)** and stores it with the donation.

## Requirements
- `createDonation` input includes **`roundId: int | null`**.
- When provided, save `roundId` on the donation.
- No server-side auto-selection or inference.

## Acceptance Criteria
make sure the project donation is valid and the project is in qf round 
send tokens from wallet to project recipient address outside giveth UI then add donation via graphql mutation
- [x] `createDonation` accepts `roundId` (int).
- [x] Donations created with `roundId` persist that value.
- [x] Existing flows work when `roundId` is omitted.
- [x] Donation reads return `roundId` if present.

divine-comedian:
need some validation 
- does qf round exist 
- is qf round active 
- is project in qf round

maryjaf:
> * `createDonation` accepts `roundId` (int).[x]  Donations created with `roundId` persist that value.[x]  Donation reads return `roundId` if present.

**PASS**

- this donation is recorded correctly in admin js with roundId

<img width="3014" height="1568" alt="Image" src="https://github.com/user-attachments/assets/2b8305e0-141e-4ab9-8ac4-bff53353c216" />

____________






> * Existing flows work when `roundId` is omitted. 

**BLOCKED**

if there is no active round:

- [x] when I try to inactive the only active round I got this error 
    "redirectUrl": "/admin/resources/QfRound",
    "notice": {
        "message": "duplicate key value violates unique constraint \"qf_round_slug_unique\"",
        "type": "danger"
    }
}

https://github.com/user-attachments/assets/88d02e9d-b655-4e3f-98c6-51556fdd34c6

@ae2079

divine-comedian:
@CarlosQ96 - this might be related to pushes you did earlier today ☝️

maryjaf:
> > * Existing flows work when `roundId` is omitted.
> 
> **BLOCKED**
> 
> if there is no active round:

This problem has been fixed and this scenario is also passed

PR #2155, opened 2025-09-02
Update create donation to accept qf round id

ae2079:
related to #2140 

## Summary by CodeRabbit

- **New Features**
  - Optionally provide a roundId when creating a donation; donations now include qfRoundId and round details in queries.
  - Donations may record a non-blocking round error message when round selection is invalid.

- **Validation**
  - Clear errors for missing, inactive, out-of-range rounds, or when a project isn’t in the specified round.
  - Explicit roundId takes priority over auto-selection; auto-selection remains when omitted.

- **Tests**
  - Added comprehensive tests for explicit, auto-selected, and error scenarios (note: test block duplicated). 

- **Chores**
  - Migration added to persist the round error message field.

### DIFF
--- a/src/resolvers/donationResolver.ts
+++ b/src/resolvers/donationResolver.ts
@@ -21,7 +21,7 @@ import { Donation, DONATION_STATUS, SortField } from '../entities/donation';
 import { ApolloContext } from '../types/ApolloContext';
 import { Project, ProjStatus } from '../entities/project';
 import { Token } from '../entities/token';
-import { publicSelectionFields, User } from '../entities/user';
+import { publicSelectionFields } from '../entities/user';
 import SentryLogger from '../sentryLogger';
 import { i18n, translationErrorMessagesKeys } from '../utils/errorMessages';
 import { NETWORK_IDS } from '../provider';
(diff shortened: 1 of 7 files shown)


src/resolvers/donationResolver.ts
```diff
@@ -935,6 +939,29 @@ export class DonationResolver {
...
+        const now = new Date();
+        if (now < qfRound.beginDate || now > qfRound.endDate) {
+          throw new Error('QF round is not currently active');
+        }
+        // Check if project is in the QF round
+        const projectInQfRound = project.qfRounds?.some(
+          qr => qr.id === roundId,
+        );
+        if (!projectInQfRound) {
+          throw new Error('Project is not part of the specified QF round');
```
CarlosQ96: We should not throw errors because the donation was done already. Then we would need to insert it later. Instead just keep it out of the qfRound by not assigining the qf Round Id, maybe we can add a field for explaining the reason? But ideally dont throw errors here.

CarlosQ96 (changes requested):

CarlosQ96 (approved):
Awesome!

PR #2163, opened 2025-09-10
update middle table statistics of qfround

CarlosQ96:
## Summary by CodeRabbit

* **New Features**
  * Per–QF round project stats: projects now track donation totals and unique donors for a specified round while preserving overall totals.

* **Refactor**
  * Admin Projects tab no longer shows per-round donation/donor counters.
  * Associating projects with a QF round no longer resets their per-round counters.

### DIFF
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -1,9 +1,10 @@
 import { Project } from '../entities/project';
+import { ProjectQfRound } from '../entities/projectQfRound';
 import {
   countUniqueDonorsAndSumDonationValueUsd,
   getProjectQfRoundStats,
 } from '../repositories/donationRepository';
-import { findActiveQfRound } from '../repositories/qfRoundRepository';
+import { findQfRoundById } from '../repositories/qfRoundRepository';
 
 export const getAppropriateSlug = async (
   slugBase: string,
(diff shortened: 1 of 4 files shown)


src/services/projectService.ts
```diff
@@ -43,13 +51,38 @@ export const updateProjectStatistics = async (projectId: number) => {
   const { totalDonations, uniqueDonors } =
     await countUniqueDonorsAndSumDonationValueUsd(projectId);
 
+  // Update Project entity with overall stats
   await Project.update(projectId, {
     totalDonations,
     countUniqueDonors: uniqueDonors,
+    totalRaised: totalDonations,
     sumDonationValueUsdForActiveQfRound,
```
ae2079, line 59: If we send the QF round ID in the parameters, the code will use this one as the active round, and I think this value will be updated incorrectly. Maybe we should not update this value here, wdyt @CarlosQ96 ?

ae2079 (commented):
thanks @CarlosQ96

src/services/projectService.ts
CarlosQ96: No because this is both a general method and related to qf. 
So the totalRaised is actually the general projects donations received. and in the sumDonationValueUsd in the middle table is the one related to QF.

kkatusic (approved):

Issue #2138, opened 2025-08-22
Admin.js + Backend: “Add Project to Round” (multi-active support)


laurenluz:
## Goal
Extend the existing **Add Project to Round** flow to work when **multiple QF rounds are active**—with the simplest possible UI and idempotent linking.

## What to build (MVP)
- **Projects list → “…” action**
  - Where we currently have **“Add Project to Round”**, on click, open a standard **Admin.js action parameter form**
  - Form shows a **dropdown of active rounds only**
  - Form shows a dropdown of active rounds only (display title in the list; submit the roundId)
  - list sorted by startDate (ideally)
  - On submit, link the project to the selected round and show a success toast.
  - If no active rounds exist, **disable** the action with tooltip: “No active QF rounds”.

- **Linking logic (reuse/extend existing)**
  - Implement/ensure a single call (service/action) that links `projectId` ↔ `roundId`.
  - **Idempotent**: if already linked, return success with “Already added” toast; do not error.
  - Validate basic inputs (valid project/round; round is active).

## Acceptance Criteria

TBD - needs review of current implementation

<s>
- [ ] The Projects list shows **“Add Project to Round”** for each project.
- [ ] Clicking the action shows a **dropdown of active rounds** (no closed/archived, shows names, submits round ID)
- [ ] Selecting a round and submitting **adds the link** (idempotent if already present).
- [ ] Success toast appears and the form closes; “Already added” shows when applicable.
- [ ] Action is **disabled** when there are no active rounds (tooltip present).</s>

## Stretch (optional, only if quick)
- **Round edit page: CSV/text batch add** of project IDs (comma/newline)... we have this as an option currently but it is broken
  - On save: validate, link idempotently, and show a summary (added / already present / invalid).

divine-comedian:
depends on #2137

maryjaf:
> ## What to build (MVP)
> * **Projects list → “…” action**
>   
>   * Where we currently have **“Add Project to Round”**, on click, open a standard **Admin.js action parameter form**
>   * Form shows a **dropdown of active rounds only**
>   * Form shows a dropdown of active rounds only (display title in the list; submit the roundId)
>   * list sorted by startDate (ideally)
>   * On submit, link the project to the selected round and show a success toast.
>   * If no active rounds exist, **disable** the action with tooltip: “No active QF rounds”.
> *

should these actions exist for removal scenarios as well? @laurenluz

CarlosQ96:
[...]

PR #2254, opened 2025-11-04
Fix issue in remove project from qf round

ae2079:
related to: #2138 

## Summary by CodeRabbit

* **Refactor**
  * Strengthened project–QF round relationship handling with unique constraints and indexing to prevent duplicate relations and improve query performance.

* **Chores**
  * Switched persistence to idempotent upsert operations to ensure consistent donation/stat tracking and safer test/setup behavior.
  
* **Migration**
  * Added a migration to enforce the new unique constraint on project–QF round relations.

### DIFF
--- a/src/entities/projectQfRound.ts
+++ b/src/entities/projectQfRound.ts
@@ -1,6 +1,6 @@
 import { Field, ID, ObjectType, Float, Int } from 'type-graphql';
 import {
-  PrimaryColumn,
+  PrimaryGeneratedColumn,
   Column,
   Entity,
   ManyToOne,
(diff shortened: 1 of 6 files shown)


kkatusic (commented):
I checked on my localhost and it is finally working :D

Great job @ae2079 

We will wait @RamRamez to check also ;)

src/entities/projectQfRound.ts
```diff
@@ -1,31 +1,34 @@
...
   UpdateDateColumn,
   Index,
+  Unique,
 } from 'typeorm';
 import { Project } from './project';
 import { QfRound } from './qfRound';
 
 @Entity('project_qf_rounds_qf_round')
 @ObjectType()
+@Unique(['projectId', 'qfRoundId'])
```
RamRamez, line 18: @ae2079 if we don't have that migration already, we need to add it

RamRamez (changes requested):
Thanks @ae2079

src/entities/projectQfRound.ts
ae2079: Done!

RamRamez (approved):
Thanks @ae2079

PR #2279, opened 2025-12-27
register users from v6 fe to impact-graph then send to v6 be

RamRamez:
* https://github.com/Giveth/giveth-v6-core/issues/7


ae2079 (approved):
LGTM, thanks @RamRamez

kkatusic (approved):

PR #2281, opened 2026-01-06
Disable access and visibility for QF Round actions in admin panel

RamRamez:
* https://github.com/Giveth/giveth-v6-core/issues/7


kkatusic (approved):
Thx @RamRamez

PR #2300, opened 2026-02-10
Changing calculation for new round number based on month

kkatusic:
- https://github.com/Giveth/giveth-dapps-v2/issues/5496


RamRamez (approved):
Thanks @kkatusic

PR #2312, opened 2026-04-17
stabilize giveconomy reverse sync and related tests

ae2079:
### Summary
This PR stabilizes the mirrored GIVeconomy -> impact-graph sync flow and folds in the follow-up CI fixes that were needed to keep the branch mergeable. It preserves the earlier precision handling while also normalizing mirrored totals that only overflow 100 because of 2-decimal rounding drift in production data.

### Changes
- Round mirrored giveconomy sync percentages using the configured precision and filter values that collapse to zero before saving on impact-graph.
- Normalize rounded mirrored boosting totals back to an exact 100 when the overflow is only rounding drift, which fixes the production stall on later sync events like event `42`.
- Add regression coverage for both the precision-rounding case and the production-shaped `100.01 after rounding` case in `giveconomyPowerSyncService` tests.
- Include the branch’s CI follow-ups: fix cause recipient network id handling in tests and make resolver assertions independent from seeded staging snapshot data.

### How to Test
1. Run `NODE_ENV=test npx mocha ./src/services/giveconomyPowerSyncService.test.ts` and confirm the mirrored sync service tests pass, including the new normalization case.
2. Run `npx eslint src/services/giveconomyPowerSyncService.ts src/services/giveconomyPowerSyncService.test.ts` and confirm the updated sync files are lint-clean.
3. In a production-like environment, configure `ENABLE_GIVECONOMY_POWER_SYNC=true` with `GIVPOWER_BOOSTING_PERCENTAGE_PRECISION=2`, replay a mirrored event whose rounded percentages sum to `100.01`, and confirm `pullGiveconomyPowerSync()` applies it instead of failing.
4. After deploy, verify the `power_sync_cursor` row for `sourceSystem='giveconomy'` advances beyond event `41` and newer mirrored events begin applying on `impact-graph`.
5. Spot-check one of the stalled users, such as `6791` or the production-shaped overflow case, and confirm the v5 `power_boosting` rows match the latest v6 mirrored payload after the cron runs.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boosting percentage normalization to correctly handle rounding adjustments and ensure total allocation remains within acceptable bounds.

### DIFF
--- a/src/services/giveconomyPowerSyncService.test.ts
+++ b/src/services/giveconomyPowerSyncService.test.ts
@@ -117,4 +117,137 @@ describe('pullGiveconomyPowerSync', () => {
     assert.isFalse(params.emitOutboxEvent);
     assert.isFunction(params.beforeSave);
   });
+
+  it('normalizes rounded mirrored percentages when they overflow 100 by rounding drift', async () => {
+    sinon.stub(powerSyncCursorRepository, 'getPowerSyncCursor').resolves({
+      sourceSystem: 'giveconomy',
+      lastEventId: 41,
+    } as any);
+    sinon
+      .stub(powerSyncCursorRepository, 'savePowerSyncCursor')
+      .resolves({} as any);
+    sinon
+      .stub(powerSyncOutboxRepository, 'getLatestPowerSyncOutboxEventForUser')
+      .resolves(null);
+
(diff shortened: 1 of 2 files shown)


RamRamez (approved):
Thanks @ae2079

PR #2340, opened 2026-07-30
fix: prevent Stellar QR draft donation race conditions and add bounded reconciliation

kkatusic:
## Issue

https://github.com/Giveth/giveth-v6-core/issues/444

## Summary

Fixes race conditions in the Stellar QR donation flow that could cause:

- A successfully matched draft to be incorrectly marked as failed.
- Duplicate donations when multiple matcher executions process the same payment.
- Late Horizon indexing to leave valid donations permanently failed.
- Overlapping cron executions to process stale draft state.

## Changes

- Added atomic, guarded draft-status transitions.
- Added PostgreSQL advisory locks for:
  - Individual draft donations.
  - Stellar transactions.
  - The complete QR cron execution.
- Added bounded lock waits and Horizon request timeouts.
- Re-checks draft and donation state while holding the appropriate locks.
- Scans expired drafts one final time before marking them as failed.
- Only accepts payments made within the draft’s validity window.
- Reconciles recently failed drafts when their Stellar payment appears later.
- Prevents matched drafts or drafts linked to live donations from being failed.
- Allows drafts linked to failed donations to be renewed and retried.
- Moves slow blockchain synchronization outside advisory-lock scopes.
- Adds an idempotent SQL script to repair historical failed drafts linked to successful donations.

## Reconciliation behavior

Recently failed drafts are rescanned for 30 minutes to account for delayed Horizon indexing. Older affected records can be repaired using:

`./scripts/reconcile-failed-qr-drafts.sql`

The script includes an inspection query and only updates failed QR drafts whose linked donation is not failed. It is safe to run repeatedly.

## Tests

Added coverage for:

- Concurrent matcher and cron executions.
- Two drafts attempting to claim the same transaction.
- Stale expiry updates racing with successful matches.
- Existing-transaction reconciliation.
- Late Horizon transaction discovery.
- Payment validity-window enforcement.
- Incorrect memo and amount handling.
- Draft renewal and failure guards.
- Cron overlap prevention.
- Repository-level atomic status transitions.
- Historical reconciliation script behavior and idempotency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR donation matching to avoid duplicate processing during concurrent activity.
  * Automatically reconciles incorrectly failed QR drafts with their existing donations.
  * Prevents valid matched donations from being incorrectly marked as failed.
  * Strengthened QR draft renewal/expiration and transaction validation (including grace-period handling).
  * Added safer duplicate-transaction handling and improved consistency for failure cases.
  * Added smart QF round selection with fallback when creating recurring donations.
* **Tests**
  * Expanded coverage for QR reconciliation, idempotency, locking, expiration/renewal rules, and concurrency scenarios.

### DIFF
--- a/scripts/reconcile-failed-qr-drafts.sql
+++ b/scripts/reconcile-failed-qr-drafts.sql
@@ -1,0 +1,46 @@
+-- One-off data repair for Stellar QR draft donations that were overwritten to
+-- 'failed' by the cron race fixed in fix/stellar-draft-race-conditions
+-- (see https://github.com/Giveth/giveth-v6-core/issues/444, staging draft 2271).
+--
+-- Runtime reconciliation (reconcileDraftWithMatchedDonation) treats every
+-- non-failed donation as live — 'verified', 'pending' and 'swapPending' alike
+-- — so this script uses the same rule: a failed draft referencing any
+-- non-failed donation is repaired to 'matched'. Drafts referencing a
+-- genuinely failed donation are never touched.
+--
+-- The script is idempotent: repaired rows have status = 'matched' and no
+-- longer match the WHERE clause on a second run. Historical failed drafts
+-- outside the cron's reconciliation window only self-heal when the donor
+-- revisits the donation page, so run this once on staging/production
+-- alongside the deploy.
+
+-- 1) Inspect the affected rows first:
+SELECT
+  d.id AS draft_id,
+  d.status AS draft_status,
+  d."matchedDonationId",
+  d."fromWalletAddress" AS draft_from,
+  dn.status AS donation_status,
+  dn."transactionId",
+  dn."createdAt" AS donation_created_at
+FROM draft_donation d
+JOIN donation dn ON dn.id = d."matchedDonationId"
+WHERE d."isQRDonation" = true
+  AND d.status = 'failed'
+  AND dn.status != 'failed';
+
+-- 2) Repair: a failed QR draft that references a non-failed donation is a
+--    successful donation. Also backfill fromWalletAddress from the donation,
+--    but only when the draft's value is null or empty.
+UPDATE draft_donation d
+SET
+  status = 'matched',
+  "fromWalletAddress" = CASE
+    WHEN COALESCE(d."fromWalletAddress", '') = '' THEN dn."fromWalletAddress"
+    ELSE d."fromWalletAddress"
+  END
+FROM donation dn
+WHERE dn.id = d."matchedDonationId"
+  AND d."isQRDonation" = true
+  AND d.status = 'failed'
+  AND dn.status != 'failed';
--- a/scripts/reconcile-failed-qr-drafts.sql
+++ b/scripts/reconcile-failed-qr-drafts.sql
@@ -8,13 +8,21 @@
 -- non-failed donation is repaired to 'matched'. Drafts referencing a
 -- genuinely failed donation are never touched.
 --
+-- The whole script runs in a single transaction and the inspection SELECT
+-- locks the affected draft and donation rows (FOR UPDATE OF d, dn), so a
+-- concurrent status transition (e.g. the blockchain sync job failing a
+-- donation) cannot slip in between inspection and repair and leave a matched
+-- draft referencing a failed donation.
+--
 -- The script is idempotent: repaired rows have status = 'matched' and no
 -- longer match the WHERE clause on a second run. Historical failed drafts
 -- outside the cron's reconciliation window only self-heal when the donor
 -- revisits the donation page, so run this once on staging/production
 -- alongside the deploy.
 
--- 1) Inspect the affected rows first:
+BEGIN;
+
+-- 1) Inspect the affected rows, locking them for the repair below:
 SELECT
   d.id AS draft_id,
   d.status AS draft_status,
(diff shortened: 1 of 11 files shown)


RamRamez:
## Code review — critical issues

Overall this is a solid, well-reasoned hardening of the QR flow: replacing read-modify-write with guarded conditional `UPDATE`s is the right fix shape, it's applied consistently across all five call sites, and the concurrency tests are real rather than happy-path. It also silently fixes a latent bug — the old `getDraftDonationById` expiry check compared `.getTime` as a function reference and never fired.

Two things I think should be resolved before merge.

---

### 1. 🔴 Unconditional `return` after the lock block defeats both the expiry-fail and the rest of the scan

`src/services/cronJobs/checkQRTransactionJob.ts:559`

`matchUnderLocks` has several early-return paths that create nothing: existing donation already claimed by another draft, the `toWalletAddress`/`amount` guard failing, `freshDraft` already matched or linked to a live donation, token/project missing, `createDonation` returning falsy. Either advisory lock failing to acquire within 30s also skips it entirely. In all of those cases control reaches the unconditional `return;` at line 559, so:

**(a) the expired draft is never failed.** Lines 563–574 — the block whose own comment promises *"An expired draft still gets this (last) scan… If no such payment is found it is failed below"* — are unreachable whenever **any** matching record was seen. The draft stays `pending` and gets one Horizon GET every cron minute until the 48h `delecteExpiredDraftDonations` sweep (~2,880 wasted requests per stuck draft). This is precisely the state the new test asserts at `src/resolvers/draftDonationResolver.test.ts:1227` (`untouched.status === PENDING`) — correct for that instant, but it never recovers.

**(b) a legitimate payment can be permanently lost.** Records come back `order=desc`, and a draft matches on *any* payment inside its window — including another draft's. Concretely:

- Draft A created `t0-5m`, expires `t0+10m`; donor A pays at `t0+5m` → matched, `matchedDonationId` set.
- Draft B created `t0`, expires `t0+15m`; donor B pays at `t0+2m`.

For draft B the loop hits A's payment first (it's newer), finds `claimingDraft === A`, skips, and `return`s — B's own unclaimed payment is never reached. Same outcome on every subsequent tick. Donor B's XLM sits at the project address with no donation row.

The precondition is the one the new test itself sets up at `src/resolvers/draftDonationResolver.test.ts:1195` — *"same project address, same project-level memo, same amount"*. Part of this predates the PR (`if (existingDonation) return;`), but this PR restructures exactly this code and already switched the memo-mismatch case from `return` to `continue`, so it seems in scope.

Suggested shape:

```ts
// only stop scanning when this payment actually became / attached to a donation
if (createdDonationId) {
  await syncDonationStatusWithBlockchainNetwork({ donationId: createdDonationId });
  notifyClients({
    type: 'new-donation',
    data: { donationId: createdDonationId, draftDonationId: donation.id },
  });
  return;
}
if (reconciled) return;   // set inside matchUnderLocks when it links an existing donation
continue;                 // otherwise keep scanning, and let the expiry block run
```

---

### 2. 🔴 Session-scoped advisory locks on pooled connections (please verify against the prod pooler)

`src/services/cronJobs/checkQRTransactionJob.ts:79-108`

`withQrAdvisoryLock` uses `pg_try_advisory_lock` / `pg_advisory_unlock` — **session**-scoped — on a TypeORM queryRunner drawn from the pool.

`src/orm.ts:8-20` documents that these DataSources "run behind a Postgres connection pooler (DigitalOcean managed Postgres / PgBouncer)", citing a real production login-churn incident. DO connection pools default to **transaction** mode, where consecutive statements on one client connection are not guaranteed to reach the same backend — which is exactly the guarantee session-level advisory locks rely on. Mutual exclusion then silently stops holding, and an unlock issued against a backend that doesn't hold the lock returns false and leaks it.

For `QR_CRON_RUN_LOCK_ID = 0` a single leak means **QR matching never runs again until the process restarts**, observable only as one `logger.info` per minute (line 597).

The same permanent-starvation outcome is reachable even without a pooler: if the unlock query throws, `finally { queryRunner.release() }` returns the connection to the pool with the lock still held.

The existing precedent in this repo is transaction-scoped and pooler-safe — `pg_advisory_xact_lock` at `src/repositories/powerBoostingRepository.ts:251` and `:447`. Suggested split:

- **per-tx / per-draft locks** (short critical sections) → `pg_try_advisory_xact_lock` inside an explicit transaction on the queryRunner. That would also want the Coingecko `getTokenPrice` and `calculateGivbackFactor` calls hoisted out, same reasoning already applied to `syncDonationStatusWithBlockchainNetwork`.
- **whole-run lock** (held across many HTTP calls, so it can't be a transaction) → a Redis `SET NX PX <ttl>` lease (Redis is already a dependency) or a row lease with an expiry timestamp, so it self-heals rather than wedging.

At minimum: confirm the prod connection string targets a session-pooled or direct port, and alert on the *"another execution is still running"* line instead of leaving it at `info`.

---

Non-blocking notes (unbounded scan set + Horizon request volume, the repair SQL living outside the migration system, silent lock-timeout skips, `logger.debug` on the "real payment but no donation created" path, null-`expiresAt` drafts becoming unmatchable, and the `isQR` token mutation leaking across suites) I'll leave separately so they don't crowd the two above.

kkatusic:
@RamRamez when you find time can you review again? thx

RamRamez (approved):
Thanks @kkatusic

Issue #2347, opened 2026-09-02
Stellar QR donation matching misses payments when an exchange deducts a fee


laurenluz:
**What happened (prod):** A donor withdrew 20 XLM from Binance to donate via the Stellar QR flow. Binance deducted its withdrawal fee, so 19.996 XLM arrived on chain. The draft donation was never matched (the matcher requires an exact amount match), and the team matched it manually. With the upcoming Stellar QF round we expect more donations sent from exchanges, and fee sizes vary by platform, so exact-amount matching can't cover them.

**Where (according to Claude):** `src/services/cronJobs/checkQRTransactionJob.ts`

- The tx is matched by exact amount equality first; the memo is only checked afterwards.
- Only transactions under 2 minutes old are considered, so a slow exchange withdrawal can be missed even when the amount is exact.
- The created donation is saved with the draft amount, not the actual on-chain amount.

**Ask:** Investigate and propose a more robust matching approach. There are two distinct cases:

1. Recipient address has no memo of its own → the tx memo is the draft donation ID (unique). Matching on memo + destination should be safe regardless of amount.
2. Recipient address requires its own memo → all donations to that project carry the same memo, so the memo can't disambiguate. Fixing case 1 alone may be enough for now — case 2 can stay on exact-amount matching.

Record the actual on-chain amount on the created donation (matters for QF matching and GIVbacks).

**Done means:** a QR donation that arrives with the correct memo but a slightly reduced amount (exchange fee) is matched automatically, and the donation records the amount actually received on chain.

kkatusic:
@maryjaf how to easy test this, add for example 1XML to donate then on the wallet change donation amount to lower: 0,98 and it should be matched.

maryjaf:
> [@maryjaf](https://github.com/maryjaf) how to easy test this, add for example 1XML to donate then on the wallet change donation amount to lower: 0,98 and it should be matched.

I tested both scenarios:

* For a project that doesn't require its own memo, when the received amount is slightly lower than the donation amount, the donation is matched successfully. 
* For a project that requires its own memo, when the received amount is lower, the donation fails to match.

Is this the expected behavior for the second case? @kkatusic

kkatusic:
@maryjaf when you find time test again, thx.

maryjaf:
> > [@maryjaf](https://github.com/maryjaf) how to easy test this, add for example 1XML to donate then on the wallet change donation amount to lower: 0,98 and it should be matched.
> 
[...]

PR #2349, opened 2026-09-04
Fix Stellar QR donation matching for exchange withdrawal fees

kkatusic:
## Isssue

https://github.com/Giveth/impact-graph/issues/2347

## Description

Fixes Stellar QR donation matching when the amount received on-chain differs from the amount requested in the draft donation due to exchange withdrawal fees.

For example, a donor may request a **20 XLM** withdrawal from Binance, but Binance deducts its withdrawal fee before broadcasting the transaction, so only **19.996 XLM** arrives at the project's Stellar address.

Previously, the QR matcher required an exact amount match before checking the memo, causing these donations to remain unmatched.

### Stellar QR matching

For recipients that **do not require their own memo**:

- Match donations using the **destination address + draft donation ID in the Stellar memo**.
- Do not require the received amount to equal the draft amount.
- Store the **actual amount received on-chain** in the final Donation.
- Calculate USD value using the actual received amount.
- Allow delayed exchange withdrawals to be reconciled within the existing failed-draft reconciliation window.

For recipients that **require their own memo**:

- Keep the existing amount-based matching behavior because the memo cannot uniquely identify the draft.
- Amount comparisons use the database's float4 precision to avoid false mismatches caused by numeric storage precision.

### Additional robustness

- Correctly handle Stellar transactions containing multiple payment operations.
- Prefer successful payments over failed attempts.
- When multiple memo-matching payments exist, prefer the largest successful payment so a dust payment does not outrank the real donation when both are available.
- Deduplicate candidate operations belonging to the same transaction.
- Prevent duplicate donations through the existing transaction/draft locking mechanism.
- Use shared Stellar operation helpers for payment/create-account handling.
- Fix Stellar transaction timestamps to use epoch seconds during transaction verification.
- Notify the QR client immediately after the donation is created, before the full donation synchronization completes.

## Example

Before:

`20 XLM requested → 19.996 XLM received → donation not matched`

After:

`20 XLM requested → 19.996 XLM received → donation matched → 19.996 XLM recorded`

## Tests

Added/updated coverage for:

- Exact-amount Stellar QR donations
- Exchange fee-reduced amounts
- Actual on-chain amount and USD value
- Wrong destination
- Wrong draft memo
- Recipient-required memo behavior
- Delayed exchange withdrawals
- Duplicate transaction protection
- Multiple payments to the same address
- Multiple operations in a single Stellar transaction
- Dust payment vs real payment
- Failed payment followed by successful retry
- `create_account` operations
- Stellar transaction verification for batched payments

## Summary by CodeRabbit

* **Improvements**
  * Improved Stellar payment matching for QR donations, including memo-based matching, amount precision tolerance, duplicate prevention, and multi-operation transactions.
  * Added support for Stellar account-creation funding and failed-payment retries.
  * Payments can now be accepted during a late-payment window, with donations recording the actual amount received.
  * Improved transaction verification by selecting the most relevant payment and rejecting stale or incorrectly addressed transactions.

* **Tests**
  * Expanded coverage for Stellar QR matching and transaction verification scenarios.

## Behavior changes from the `timestamp` fix (files this PR doesn't touch)

`NetworkTransactionInfo.timestamp` for Stellar was Horizon's ISO string in a `number`-typed field; it is now epoch seconds. Two checks that were silently dead for Stellar (`NaN` comparisons) now enforce:

1. **Staleness check** (`src/services/chains/index.ts:251`): `TRANSACTION_CANT_BE_OLDER_THAN_DONATION` now fires for Stellar. QR donations are exempt (`importedFromDraftOrBackupService`), so the reachable case is the AdminJS *add donation* flow, where reconciling a transaction mined more than an hour before the hand-entered timestamp now fails.
2. **QF round-period guard** (`src/server/adminJs/tabs/donationTab.ts:428`): previously passed unconditionally for Stellar (`Invalid Date` compares false both ways); it now rejects admin attempts to attach an out-of-window Stellar transaction to a round.

Also addressed in review follow-up (2368be87): AdminJS `createdAt` now uses `timestamp * 1000` for every chain; CASE 1 matching gained a one-sided minimum-amount floor (`QR_CASE1_MIN_AMOUNT_FACTOR = 0.9`) that accepts fee-reduced payments and rejects dust; the memo-less exact-amount `create_account` accept path is restored inside the tight validity window.

### DIFF
--- a/src/services/chains/stellar/stellarOperations.ts
+++ b/src/services/chains/stellar/stellarOperations.ts
@@ -8,10 +8,8 @@
 // Addresses are compared case-insensitively deliberately: Stellar strkeys
 // are canonically uppercase, but stored wallet addresses are user-supplied
 // and must mean the same account in every layer regardless of casing.
-export const stellarOperationPaysAddress = (
-  record: any,
-  address: string,
-): boolean => {
+// Module-private: callers go through isNativeStellarDeposit below.
+const stellarOperationPaysAddress = (record: any, address: string): boolean => {
   const recipient =
     record?.type === 'payment'
       ? record.to
--- a/src/services/chains/stellar/stellarOperations.ts
+++ b/src/services/chains/stellar/stellarOperations.ts
@@ -1,0 +1,53 @@
+// Shared knowledge about Stellar Horizon operation records, used by both the
+// QR draft matcher (checkQRTransactionJob) and donation verification
+// (transactionService) so the two layers can never disagree about which
+// operation pays an address or how much it transferred. Add new operation
+// shapes (e.g. path_payment_strict_send) here, never inline at a call site.
+
+// Whether this Horizon operation record delivers funds to `address`.
+// Addresses are compared case-insensitively deliberately: Stellar strkeys
+// are canonically uppercase, but stored wallet addresses are user-supplied
+// and must mean the same account in every layer regardless of casing.
+export const stellarOperationPaysAddress = (
+  record: any,
+  address: string,
+): boolean => {
+  const recipient =
+    record?.type === 'payment'
+      ? record.to
+      : record?.type === 'create_account'
+        ? record.account
+        : undefined;
+  return (
+    typeof recipient === 'string' &&
+    recipient.toLowerCase() === address.toLowerCase()
+  );
+};
+
+// What this operation actually transferred on-chain: payments carry
+// `amount`, create_account carries `starting_balance`. An unknown operation
+// type (or a malformed record) yields NaN, which callers must treat as
+// "not a usable operation".
+export const stellarOperationAmount = (record: any): number =>
+  record?.type === 'create_account'
+    ? Number(record.starting_balance)
+    : Number(record?.amount);
+
+// The full "is this operation a usable XLM deposit to `address`" predicate:
+// the operation-type whitelist (native payments and create_account), the
+// recipient check, and a parseable amount, in one place — so the QR matcher
+// and donation verification can never disagree about which operations count.
+export const isNativeStellarDeposit = (record: any, address: string): boolean =>
+  (record?.type === 'payment'
+    ? record.asset_type === 'native'
+    : record?.type === 'create_account') &&
+  stellarOperationPaysAddress(record, address) &&
+  Number.isFinite(stellarOperationAmount(record));
+
+// The shared tie-breaking rule for which operation of one transaction a
+// donation refers to: the largest transferred amount. Callers layer their
+// own higher-priority keys on top (the QR matcher ranks successful
+// transactions first; verification prefers the amount closest to the
+// donation being verified).
+export const byLargestStellarOperation = (a: any, b: any): number =>
+  stellarOperationAmount(b) - stellarOperationAmount(a);
--- a/src/services/chains/stellar/transactionService.test.ts
+++ b/src/services/chains/stellar/transactionService.test.ts
@@ -1,0 +1,204 @@
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { getStellarTransactionInfoFromNetwork } from './transactionService';
+import { ChainType } from '../../../types/network';
+import { NETWORK_IDS } from '../../../provider';
+import {
+  assertThrowsAsync,
+  generateRandomStellarAddress,
+  generateRandomStellarTxHash,
+  horizonPaymentRecord,
+  stubStellarHorizonPayments,
+} from '../../../../test/testUtils';
+
+describe(
+  'getStellarTransactionInfoFromNetwork test cases',
+  getStellarTransactionInfoFromNetworkTestCases,
+);
+
+function getStellarTransactionInfoFromNetworkTestCases() {
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  function baseInput(overrides: Partial<any> = {}) {
+    return {
+      networkId: NETWORK_IDS.STELLAR_MAINNET,
+      chainType: ChainType.STELLAR,
+      symbol: 'XLM',
+      txHash: generateRandomStellarTxHash(),
+      fromAddress: generateRandomStellarAddress(),
+      toAddress: generateRandomStellarAddress(),
+      amount: 19.996,
+      timestamp: Math.floor(Date.now() / 1000),
+      ...overrides,
+    };
+  }
+
+  it('should validate a single-operation payment transaction (existing behavior)', async () => {
+    const input = baseInput();
+    stubStellarHorizonPayments([
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '19.9960000',
+      }),
+    ]);
+
+    const info = await getStellarTransactionInfoFromNetwork(input as any);
+    expect(info.to).to.equal(input.toAddress);
+    expect(info.amount).to.equal(19.996);
+  });
+
+  it('should pick the payment operation targeting the expected recipient in a multi-operation (exchange batch) transaction', async () => {
+    const input = baseInput();
+    stubStellarHorizonPayments([
+      // Another withdrawal in the same exchange batch, listed first
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: generateRandomStellarAddress(),
+        amount: '55.0000000',
+      }),
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '19.9960000',
+      }),
+    ]);
+
+    const info = await getStellarTransactionInfoFromNetwork(input as any);
+    expect(info.to).to.equal(input.toAddress);
+    expect(info.amount).to.equal(19.996);
+  });
+
+  it('should select the largest of several operations paying the same recipient, mirroring the QR matcher', async () => {
+    const input = baseInput();
+    stubStellarHorizonPayments([
+      // A smaller operation to the same recipient, listed first: the QR
+      // matcher credits the largest, so verification must resolve it too
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '0.1000000',
+      }),
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '19.9960000',
+      }),
+    ]);
+
+    const info = await getStellarTransactionInfoFromNetwork(input as any);
+    expect(info.to).to.equal(input.toAddress);
+    expect(info.amount).to.equal(19.996);
+  });
+
+  it('should prefer the operation matching the expected amount when a batch pays the recipient for two different donations', async () => {
+    // Two donors' withdrawals to the same project address in one exchange
+    // batch: this donation is the smaller op, so amount must outrank size
+    const input = baseInput({ amount: 5 });
+    stubStellarHorizonPayments([
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        to: input.toAddress,
+        amount: '100.0000000',
+      }),
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '5.0000000',
+      }),
+    ]);
+
+    const info = await getStellarTransactionInfoFromNetwork(input as any);
+    expect(info.amount).to.equal(5);
+    expect(info.from).to.equal(input.fromAddress);
+  });
+
+  it('should break an equal-amount tie between operations by the expected sender', async () => {
+    // Two ops of the same amount to the same address, different senders;
+    // only the sender distinguishes which one this donation refers to
+    const input = baseInput({ amount: 5 });
+    stubStellarHorizonPayments([
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        to: input.toAddress,
+        amount: '5.0000000',
+      }),
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '5.0000000',
+      }),
+    ]);
+
+    const info = await getStellarTransactionInfoFromNetwork(input as any);
+    expect(info.from).to.equal(input.fromAddress);
+    expect(info.amount).to.equal(5);
+  });
+
+  it('should reject a transaction much older than the donation (staleness check)', async () => {
+    const input = baseInput();
+    stubStellarHorizonPayments([
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '19.9960000',
+        createdAt: new Date(Date.now() - 2 * 60 * 60 * 1000),
+      }),
+    ]);
+
+    await assertThrowsAsync(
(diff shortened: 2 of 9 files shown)


src/services/chains/stellar/transactionService.ts
```diff
@@ -22,37 +30,71 @@ const getStellarTransactionInfo = async (
...
+        Number(operationSender(b) === expectedFrom) -
+          Number(operationSender(a) === expectedFrom) ||
+        byLargestStellarOperation(a, b),
+    )[0] ?? records[0];
 
   if (!transaction) return null;
 
+  // Horizon's created_at is an ISO string; NetworkTransactionInfo carries
+  // epoch seconds (the staleness validation subtracts timestamps).
+  const timestampSecs = new Date(transaction.created_at).getTime() / 1000;
```
RamRamez, line 64: **Critical** — this is the right value (the interface has always declared `timestamp: number`; the old code returned Horizon's ISO string in violation of it), but there's a consumer written *around* that violation.

`src/server/adminJs/tabs/donationTab.ts:543`:

```ts
createdAt:
  chainType === ChainType.STELLAR
    ? new Date(transactionInfo?.timestamp)        // assumed an ISO string
    : new Date(transactionInfo?.timestamp * 1000),
```

`new Date(1780000000)` reads the number as **milliseconds**, giving `1970-01-21T14:26:40Z`. So every Stellar donation created through the AdminJS *add donation* flow gets a `createdAt` roughly 56 years in the past — and that date feeds project statistics, user totals and QF round attribution.

TypeScript can't flag it: the declared type was already `number`, so the branch that exists to handle a string looks correct to the compiler, and the Horizon records are `any` so nothing caught the old lie either.

Worth noting this is the same manual-reconciliation flow the CASE 1 limitation notes fall back on for dust and no-memo cases, so it will get exercised.

**Fix:** delete the Stellar branch in `donationTab.ts` — `new Date(timestamp * 1000)` is now correct for every chain.

src/services/cronJobs/checkQRTransactionJob.ts
```diff
@@ -414,6 +437,45 @@ export async function checkTransactions(
...
+    //   payment broadcast after expiry stays acceptable for
+    //   QR_LATE_PAYMENT_ACCEPTANCE_WINDOW_MS.
+    // - CASE 2 (recipient requires its own memo): several drafts to the same
+    //   project can share destination and memo, so the draft amount remains
+    //   the only disambiguator (compared at float4 precision, the column
+    //   type) and the validity window stays tight: without a unique
+    //   identifier a late payment must not be guessed at. create_account has
+    //   no memo to distinguish it with, so it keeps amount-only matching.
+    // Known limitation, accepted with issue #2347: CASE 1 deliberately has
+    // no amount condition — an exact, tolerance or minimum-amount rule would
```
RamRamez: **High** — a *minimum* rule wouldn't, and that asymmetry is the whole point: exchange fees only ever reduce the amount, so a one-sided floor accepts every fee-reduced payment while rejecting dust. Binance's XLM withdrawal fee is on the order of 0.002 XLM, and `19.996 / 20 = 0.9998`; a floor at `amount * 0.9` clears the example in the PR description with four orders of magnitude of headroom and rejects a `0.0000001` XLM dust payment by a factor of roughly 2×10^8.

The exposure that's left is cheap and reliable rather than theoretical:

- the memo is the draft's sequential `SERIAL` id, printed in the QR;
- `verifyQRDonationTransaction(id)` (`draftDonationResolver.ts:616`) is unauthenticated and returns the whole `DraftDonation`, so `toWalletAddress` and `amount` are enumerable by id;
- dust carrying `memo = <draft id>` claims the draft — `MATCHED`, dust donation created — and the donor's real payment then hits the already-claimed-hash path and needs manual reconciliation.

The candidate ordering below only defends the case where both payments are on the same Horizon page at scan time. With a per-minute cron and a draft id known at QR-generation time, the attacker picks the tick. The round's `minimumValidUsdValue` limits the matching-fund damage, but not the donor-paid-and-got-nothing outcome.

I don't think closing this needs the product change described here (an unguessable memo token) — a floor *is* a matching-rule change and closes it now. Suggest `opAmount >= amount * FLOOR` with `FLOOR` ≈ 0.9, keeping the deliberate absence of an upper bound. And correct this comment either way: as written it would talk the next reader out of the fix.

src/services/cronJobs/checkQRTransactionJob.ts
```diff
@@ -423,267 +485,314 @@ export async function checkTransactions(
...
-          }
-          creationContext = preparedContext;
-          // Missing token/project configuration: nothing can be created this
-          // tick, and the draft must not be failed on that evidence alone.
-          if (!creationContext) {
-            abortReason =
-              'QR donation configuration unavailable (missing token or project)';
-            break;
-          }
+      } else if (memo !== id.toString()) {
```
RamRamez: **Medium** — flagging this as a decision to make deliberately rather than absorb as a side effect. The reasoning is right *given* no amount floor: memo-less plus no amount condition would accept any deposit at all. But the two constraints are coupled — with the floor suggested above, "exact amount, no memo, `create_account`" becomes safe to keep as an *additional* accept path, and the narrowing stops being necessary.

Until then a donor funding a previously-unfunded project address without a memo moves from auto-matched to manual reconciliation. If the floor doesn't land, this belongs in the release notes so support expects those tickets.

src/services/cronJobs/checkQRTransactionJob.ts
```diff
@@ -414,6 +437,45 @@ export async function checkTransactions(
...
+    // unguessable memo token instead of the draft id, or crediting every
+    // memo-matching payment — not a matching-rule tweak.
+    // Known limitation (pre-existing): donations are keyed by transaction
+    // hash, so one exchange batch transaction carrying payments for TWO
+    // CASE 2 drafts (same address and memo, different amounts) can only ever
+    // credit one of them — the second draft skips the already-claimed hash
+    // until it fails at expiry. Crediting both needs per-(hash, operation)
+    // donation keying, a schema/product decision.
+    const requiresRecipientMemo = Boolean(toWalletMemo);
+    const paymentWindowEndMs = requiresRecipientMemo
```
RamRamez: **Low** — `expiresAtDate` already carries `DRAFT_DONATION_EXPIRY_GRACE_MS`, while `getScannableDraftDonations` cuts off at `expiresAt > now - QR_FAILED_DRAFT_RECONCILIATION_WINDOW_MS`. So the final minute of the acceptance window is unreachable from the cron and exists only on the `graphql-verify` path — the two aren't quite "the same span" the constant's comment describes: same constant, different base.

Either base this on the raw `expiresAt`, subtract the grace, or adjust the comment. The drift between comment and code is the part worth removing.

src/services/chains/stellar/transactionService.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      horizonPaymentRecord({
+        txHash: input.txHash,
+        from: input.fromAddress,
+        to: input.toAddress,
+        amount: '19.9960000',
+        createdAt: new Date(Date.now() - 2 * 60 * 60 * 1000),
+      }),
+    ]);
+
+    await assertThrowsAsync(
```
RamRamez, line 160: **Low** — `assertThrowsAsync(fn, undefined)` takes the else branch of `testUtils.ts:65` and calls `assert.throw(f)` with no expected message, so this passes on *any* throw. It would go green on an unrelated `TRANSACTION_NOT_FOUND` — which is exactly the regression it exists to catch, given that the staleness check it covers was dead code (`NaN > ONE_HOUR`) until this PR. Same for the wrong-recipient case at :176.

Worth passing the expected i18n key to both.

src/utils/utils.ts
```diff
@@ -482,3 +482,16 @@ export const generateRandomNumericCode = (digits: number = 6): number => {
   const max = Math.pow(10, digits) - 1;
   return Math.floor(min + Math.random() * (max - min + 1));
 };
+
+export function generateHexNumber(len: number): string {
```
RamRamez: **Low** — breaking the `src → test/testUtils` import in `chainvineMockAdapter.ts` was the right call, and that adapter is now the only production file that needed it. The side effect is that a `Math.random()` address generator ships in production utils for one mock adapter's benefit. Defining both helpers inside `chainvineMockAdapter.ts` keeps the blast radius at the mock. Non-blocking.

src/services/chains/stellar/stellarOperations.ts
```diff
@@ -0,0 +1,53 @@
...
+// QR draft matcher (checkQRTransactionJob) and donation verification
+// (transactionService) so the two layers can never disagree about which
+// operation pays an address or how much it transferred. Add new operation
+// shapes (e.g. path_payment_strict_send) here, never inline at a call site.
+
+// Whether this Horizon operation record delivers funds to `address`.
+// Addresses are compared case-insensitively deliberately: Stellar strkeys
+// are canonically uppercase, but stored wallet addresses are user-supplied
+// and must mean the same account in every layer regardless of casing.
+export const stellarOperationPaysAddress = (
```
RamRamez: **Low** — `stellarOperationPaysAddress` has no consumer outside this module; `isNativeStellarDeposit` is its only caller. Fine to keep exported if you expect the path-payment shapes the header comment mentions to need it directly, otherwise it can drop to module-private. Non-blocking.

RamRamez (commented):
Reviewed at 14a170d7e (three-dot diff vs `staging`). Careful, well-documented work — the CASE 1 / CASE 2 split is the right model for this problem, the restructuring into an identification pass, a ranking and a lock cycle is a real improvement over the old single-pass loop, and the test coverage is genuinely thorough. `tsc --noEmit` and `eslint` are clean on the branch and all 13 CI checks pass. I couldn't run the DB-backed suites locally (test Postgres on :5443 was down), so I'm relying on CI for those.

Two things I'd want changed before merge, both left inline:

- **Critical** — the `timestamp` change breaks `donationTab.ts:543`, putting `createdAt` in January 1970 for every admin-created Stellar donation.
- **High** — CASE 1 has no amount floor, and the rationale given for that doesn't hold. A one-sided minimum accepts fee-reduced payments *and* rejects dust.

Plus one medium and four low notes inline.

### Two more consequences of the `timestamp` change, in files this PR doesn't touch

Behaviour changes rather than defects, but they're invisible in the diff and worth a line in the PR description:

**1. `src/services/chains/index.ts:251`** — `input.timestamp - transaction.timestamp > ONE_HOUR` was `number − string` → `NaN` for Stellar, and `NaN > 3600` is `false`, so the staleness check never fired on this chain. It fires now. QR donations are exempt (`importedFromDraftOrBackupService` is true once a matched draft exists) and ordinary donations are created before their transaction is mined, so the reachable case is `donationTab.ts:235`, where `timestamp` is typed in by hand: reconciling a transaction mined more than an hour before the entered timestamp now fails with `TRANSACTION_CANT_BE_OLDER_THAN_DONATION`. That's the same manual path the CASE 1 limitation notes fall back on — consider deriving the admin timestamp from the on-chain time, or passing `importedFromDraftOrBackupService: true` there.

**2. `src/server/adminJs/tabs/donationTab.ts:428`** — `new Date(transactionInfo?.timestamp * 1000)` was `new Date(NaN)`, and both `<` and `>` against an Invalid Date are `false`, so the QF round-period guard passed unconditionally for Stellar. It now enforces. Correct, but it will start rejecting admin attempts to attach an out-of-window Stellar transaction to a round that previously went through.

### Verified correct

- `Math.fround` is sound here — `real` columns round-trip through node-pg as the widened float4, so `fround(read) === fround(onChain)` holds across Stellar's 7-decimal amounts.
- `FAILED → MATCHED` is permitted by `updateDraftDonationStatus` (only `status != MATCHED` guards the matched path), so the late-reconciliation claim genuinely works for drafts the frontend already timed out.
- Recording the actual on-chain amount keeps the downstream `closeTo(transaction.amount, input.amount)` check consistent, and the matcher's ranking is aligned with the new verification ranking, so both resolve the same operation of a multi-op transaction.
- Moving `notifyClients` ahead of `syncDonationStatusWithBlockchainNetwork` fixes a real bug in passing: a throwing sync previously meant the donor's QR screen never received its event.
- CASE 2's validity window and amount semantics are unchanged from `staging`.
- The new `stellar/transactionService.test.ts` is picked up by CI's `src/services/**/*.test.ts` matrix glob, so no `test:*` script is needed despite the repo convention.

<sub>Two further findings were set aside as low-probability: the verification ranking at `transactionService.ts:51` orders on amount-closeness before sender, so it can select an operation that then fails the `from` check while a fully valid sibling exists in the same transaction; and the per-hash dedupe credits only the highest-ranked operation, so one transaction paying the address twice under the draft's memo records one amount and drops the rest.</sub>

src/services/chains/stellar/transactionService.ts
kkatusic: Fixed in 2368be87 — dropped the Stellar branch in donationTab.ts: `createdAt` is now `new Date(transactionInfo?.timestamp * 1000)` for every chain, with a comment noting the timestamp is epoch seconds everywhere (Stellar included).

src/services/cronJobs/checkQRTransactionJob.ts
kkatusic: Applied in 2368be87 — added `QR_CASE1_MIN_AMOUNT_FACTOR = 0.9`: a CASE 1 memo-identified operation must now carry at least 90% of the draft amount (one-sided, no upper bound). The limitation comment is rewritten around the floor instead of arguing against it, and a new test covers dust and a 17-of-20 underpayment with the correct memo both leaving the draft pending.

src/services/cronJobs/checkQRTransactionJob.ts
kkatusic: Done in 2368be87 — with the floor in place, the memo-less float4-exact create_account accept path is restored inside the tight validity window (the extended post-expiry window still requires the memo). Tests updated accordingly: the exact-amount memo-less funding matches again, plus a boundary test asserting the memo-less path does not get the extended window.

src/services/cronJobs/checkQRTransactionJob.ts
kkatusic: Aligned in 2368be87 — the CASE 1 acceptance window is now based on raw `expiresAt` (no grace term), so it shares both base and span with getScannableDraftDonations and ends exactly when the cron stops scanning the draft; the constant's comment now says exactly that.

src/services/chains/stellar/transactionService.test.ts
kkatusic: Fixed in 2368be87 — both tests now assert the exact i18n messages via `errorMessages`: `TRANSACTION_CANT_BE_OLDER_THAN_DONATION` for the staleness case and `TRANSACTION_TO_ADDRESS_IS_DIFFERENT_FROM_SENT_TO_ADDRESS` for the wrong-recipient case.

src/utils/utils.ts
kkatusic: Done in 2368be87 — the two generators are now private to chainvineMockAdapter; `src/utils/utils.ts` and test/testUtils.ts are back to their previous shape, so nothing test-flavored ships in production utils and production still doesn't import testUtils.

src/services/chains/stellar/stellarOperations.ts
kkatusic: Done in 2368be87 — `stellarOperationPaysAddress` is module-private now; `isNativeStellarDeposit` is the module's public surface for shape checks.

RamRamez (approved):
Nice job, thanks @kkatusic

PR #2352, opened 2026-09-09
Staging to master update (without GIVeconomy power-sync removal)

kkatusic:
Production release of the Stellar QR donation matching work (#2347 / #2349 / #2351). Deliberately excludes #2345 (GIVeconomy power-sync consumer removal, 88dfe9cb) — not wanted on production yet. Verified identical to staging minus that one change.


RamRamez (approved):
Thanks @kkatusic

PR #2354, opened 2026-09-15
Fix email.csv unauthenticated downloads

mmahdigh:


src/server/bootstrap.ts
```diff
@@ -209,21 +210,27 @@
...
+        const filePath = path.join(
+          exportsDir,
+          path.basename(req.params.filename),
+        );
+        if (path.dirname(filePath) !== exportsDir) {
+          res.status(400).send('Invalid filename');
+          return;
+        }
+        res.download(filePath);
+      },
```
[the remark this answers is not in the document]
mmahdigh: Not a big issue as it's already admin-only. However, added a rate-limiter anyway.

src/middleware/adminSessionAuthentication.ts
```diff
@@ -0,0 +1,29 @@
...
+ * PII) can never be fetched by anonymous or non-admin callers.
+ */
+export const adminSessionAuthentication = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  // getCurrentAdminJsSession returns the db user behind the session cookie,
+  // or false when there is no valid session
+  const admin = await getCurrentAdminJsSession(req);
```
kkatusic, line 19: **Blocker:** a request with no `Cookie` header at all (plain `curl`, which is exactly the caller this PR is meant to stop) never gets a response.

`getCurrentAdminJsSession` does `cookie.parse(request.headers.cookie)` outside its try block, and `cookie@0.4.2` throws `TypeError: argument str must be a string` on `undefined`. That rejection escapes this `await`, Express 4 does not catch async middleware errors, so the socket just hangs until the proxy times out and we get an `unhandledRejection` in Sentry on every probe. The same happens for any other rejection here, e.g. a transient DB error in `findUserById`.

Suggest wrapping this in `try/catch` (treat errors as no session and return 401, or `next(err)`), and hardening the helper with `cookie.parse(cookieHeader || '')`.

src/middleware/adminSessionAuthentication.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // Re-check the role in case it was downgraded after login (same criterion
+  // as the login lookup: role != restricted)
+  if (!admin || admin.role !== UserRole.ADMIN) {
```
kkatusic, line 23: **Blocker:** this locks out QF managers. `projectPermissions[UserRole.QF_MANAGER].exportEmails` is `true` in `adminJsPermissions.ts`, and the `exportEmails` bulk action is gated by that table, so a `qfManager` can run the export, get the "refresh to download" notice, and then hit a 401 here.

The comment above says "same criterion as the login lookup: role != restricted", which matches neither this code nor the permission table. I would reuse the existing helper so both sides stay in sync:

```ts
if (!admin || !canAccessProjectAction({ currentAdmin: admin }, ResourceActions.EXPORT_EMAILS)) {
```

src/server/bootstrap.ts
```diff
@@ -209,21 +211,36 @@ export async function bootstrap() {
...
-      res.download(filePath);
-    });
+    // To download email addresses of projects in AdminJS projects tab.
+    // Authenticated with the AdminJS session: the exported CSV contains user
+    // PII and must never be downloadable by anonymous or non-admin callers.
+    app.get(
+      '/admin/download/:filename',
+      // Limit attempts before session lookup and file access, independently of
+      // the global limiter's /admin exemption. Share counts across instances.
+      createAdminDownloadRateLimiter(
```
kkatusic, line 221: We never call `app.set('trust proxy', ...)` anywhere, but `ipWhitelist.ts` reads `X-Forwarded-For`, so in prod `req.ip` is the proxy address for everyone. That turns this into a single 10/min bucket shared by all admins, and since the limiter runs before auth, an anonymous caller can burn it and block every real download for the rest of the minute. express-rate-limit 7 will also log `ERR_ERL_UNEXPECTED_X_FORWARDED_FOR` on the first proxied request.

Either set `trust proxy` properly, or put the limiter after `adminSessionAuthentication` and key on the admin user id / session cookie instead of IP. The unit test injects `req.ip` directly so it cannot catch this.

src/middleware/adminSessionAuthentication.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  } finally {
+    getCurrentAdminJsSession.restore();
+  }
+  return { status, send, next };
+};
+
+const adminWithRole = (role: UserRole) => ({ role }) as User;
+
+describe('adminSessionAuthentication', () => {
+  it('rejects anonymous callers (no valid session) with 401', async () => {
```
kkatusic, line 28: This test stubs `getCurrentAdminJsSession` to resolve `false` and passes `{}` as the request, but the real helper cannot return `false` for a cookie-less request, it throws (see the comment on the middleware). So the test named "rejects anonymous callers" never exercises the anonymous path, and there is no `QF_MANAGER` case either.

Could you add a case that calls the middleware with a real request object lacking a `Cookie` header, one with a garbage signed `adminjs` cookie, and one for `QF_MANAGER`? A single supertest request against the assembled route would be even better.

Also, in the rate limiter test `sinon.useFakeTimers()` with no options fakes all timers process-wide, and the full mocha run shares the process with live ioredis/TypeORM connections. Narrowing it with `toFake: ['Date', 'setTimeout', 'setInterval', 'clearTimeout', 'clearInterval']` would avoid odd cross-suite flakes.

kkatusic (commented):
Thanks for this, the direction is right and the new unit tests pass locally for me. A few things I think we need to handle before merging, the first two are blockers:

1. **Cookie-less requests hang instead of getting a 401** (inline comment on `adminSessionAuthentication.ts`).
2. **QF managers get locked out of a download they are allowed to trigger** (inline comment on `adminSessionAuthentication.ts`).
3. **Rate limiter is keyed on IP but we never set `trust proxy`** (inline comment on `bootstrap.ts`).
4. **Expired sessions log a TypeError on every request.** In `getCurrentAdminJsSession` (`src/server/adminJs/adminJs.ts:117`), connect-redis calls the callback with no arguments on a cache miss, so `sessionObject.adminUser` throws and we log `getCurrentAdminJsSession error` for every stale or forged `adminjs` cookie. This helper had no callers before this PR, so the latent bug becomes live noise now. `sessionObject?.adminUser` fixes it. While there, the `new RedisStore(...)` inside that function could be hoisted to module scope and shared with the router config.
5. **The tests stub out the part that can fail** (inline comment on the test file).

Not blocking, but worth a follow-up issue: authenticating the route does not fix the underlying `emails.csv` design in `projectsTab.ts`. It is one fixed file per container, never cleaned up, so an admin can download another admin's previous export, and with several replicas the redirect can land on an instance that never wrote the file. Returning the CSV from the action directly, or writing a random-token file with a short redis TTL that the route validates and unlinks, would remove the shared file and most of the need for the IP limiter.

Small nits: the `RedisStore` + `@ts-expect-error sendCommand` block is now duplicated from the global limiter, a shared factory would be nicer. And the comment about being "independent of the global limiter's /admin exemption" is a bit misleading, the route is registered before `app.use(limiter)` so the global limiter never applied here anyway.

kkatusic (approved):

mmahdigh:
> Thanks for this, the direction is right and the new unit tests pass locally for me. A few things I think we need to handle before merging, the first two are blockers:
> 
> 1. **Cookie-less requests hang instead of getting a 401** (inline comment on `adminSessionAuthentication.ts`).
> 2. **QF managers get locked out of a download they are allowed to trigger** (inline comment on `adminSessionAuthentication.ts`).
> 3. **Rate limiter is keyed on IP but we never set `trust proxy`** (inline comment on `bootstrap.ts`).
> 4. **Expired sessions log a TypeError on every request.** In `getCurrentAdminJsSession` (`src/server/adminJs/adminJs.ts:117`), connect-redis calls the callback with no arguments on a cache miss, so `sessionObject.adminUser` throws and we log `getCurrentAdminJsSession error` for every stale or forged `adminjs` cookie. This helper had no callers before this PR, so the latent bug becomes live noise now. `sessionObject?.adminUser` fixes it. While there, the `new RedisStore(...)` inside that function could be hoisted to module scope and shared with the router config.
> 5. **The tests stub out the part that can fail** (inline comment on the test file).
> 
> Not blocking, but worth a follow-up issue: authenticating the route does not fix the underlying `emails.csv` design in `projectsTab.ts`. It is one fixed file per container, never cleaned up, so an admin can download another admin's previous export, and with several replicas the redirect can land on an instance that never wrote the file. Returning the CSV from the action directly, or writing a random-token file with a short redis TTL that the route validates and unlinks, would remove the shared file and most of the need for the IP limiter.
> 
> Small nits: the `RedisStore` + `@ts-expect-error sendCommand` block is now duplicated from the global limiter, a shared factory would be nicer. And the comment about being "independent of the global limiter's /admin exemption" is a bit misleading, the route is registered before `app.use(limiter)` so the global limiter never applied here anyway.

1- This is okay and not related to the current changes
2- We have only admins right now, although there's code for qf managers and various other roles
3- The rate limit itself was not even needed, this is already behind an admin gate anyway
4- Again, predating this PR
5- I'll check this later; good point, but it's just the tess

